PR #1, opened 2021-05-13
Swipe mixin

KemoPaw:


components/mixins/Swipe/Swipe.ts
```diff
@@ -0,0 +1,18 @@
+export const mobileSwipe = {
+    created: () => {
+      // can add functions here on start-up if needed
+    },
+    methods: {
+      settingsSwipeHandler (currThis: any) {
```
RetroPronghorn, line 6: let's refactor this to swipeSidebar instead so that it can be used in multiple components instead of just building this for settings. After doing so please also implement it onto the sidebar for chatting (url /chat/direct)

layouts/settings/Settings.html
```diff
@@ -1,4 +1,4 @@
-<div id="app" :class="`with-sidebar ${sidebar ? 'is-open' : 'is-collapsed'}`">
+<div id="app" :class="`with-sidebar ${sidebar ? 'is-open' : 'is-collapsed'}`" v-touch:swipe="settingsSwipeHandler(this)" v-touch-options="{ swipeTolerance: 75 }">
```
RetroPronghorn, line 1: break this up with newlines, it's getting a bit lengthy, just means putting soft returns so like 
<div
  id="app"
  :class="..."
  etc..
  />

KemoPaw: Okay! Thank you!

PR #2, opened 2021-05-14
chore(settings):

WanderingHogan:
Updating audio settings and adding permissions request to get microphone sources/inputs. Adding detectrtc library because it handles some of the problems firefox and safari have with not implementing the full permissions api.

### DIFF
--- a/pages/settings/audio/index.vue
+++ b/pages/settings/audio/index.vue
@@ -3,17 +3,22 @@
 <script lang="ts">
 import Vue from 'vue'
 import { mapState } from 'vuex'
+import { UserPermissions } from '../../../components/mixins/UserPermissions'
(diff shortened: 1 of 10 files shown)


pages/settings/audio/index.vue
```diff
@@ -44,29 +57,98 @@ export default Vue.extend({
...
-          default:
-            break
+        })
+        .catch((err) => {
+          // Error is returned if user selects Block/Deny
+          this.$data.userDeniedAudioAccess = true
+        })
+    },
+  },
+  mounted() {
```
RetroPronghorn, line 124: One small change, it looks like we're doing the same thing in mounted as the enable audio method, if that's the case we can just do mounted() { this.enableAudio() }

PR #7, opened 2021-05-26
Emoji picker added (emoji-mart-vue lib)

jpanay:


components/tailored/core/chatbar/Chatbar.vue
```diff
@@ -23,6 +35,33 @@ export default Vue.extend({
...
   },
+  methods: {
+    toggleEmojiPicker(e: any) {
+      this.$data.showEmojiPicker = !this.$data.showEmojiPicker
+      if (!this.$data.showEmojiPicker) {
+        let div = this.$refs.emojiPicker
+        // @ts-ignore
+        div.$el.style.opacity = 0
+      } else {
+        let divChecker = setInterval(() => {
```
RetroPronghorn, line 46: This doesn't seem ideal, what is the need for this interval? We should just be able to conditionally render the components opacity based on a click handler that checks local data.

components/tailored/core/chatbar/Chatbar.html
```diff
@@ -8,8 +8,12 @@
     <div
       :data-tooltip="$t('global.glyphs')"
       class="control-tooltip-wrap has-tooltip has-tooltip-top has-tooltip-primary">
-      <font-awesome-icon :icon="['far', 'grin-tongue-wink']" class="control-icon" />
+      <font-awesome-icon :icon="['far', 'grin-tongue-wink']" class="control-icon" @click="toggleEmojiPicker"/>
     </div>  
+    <picker :data="emojiIndex" class="emoji-picker" @select="addEmoji" v-if="showEmojiPicker"
```
RetroPronghorn, line 13: this can be something like `:class="`emoji-picker ${showEmojiPicker ? 'visible' : 'hidden'}`"`

RetroPronghorn:
@jpanay looks good, I'll probably manually update the position in css later but this works for now. Can you add documentation to your methods then we're good to merge.

PR #20, opened 2021-07-02
Sa 12

WanderingHogan:
@RetroPronghorn any formatting issues can we screenshare to get my environment doing the same formatting as yours?

I am using prettier and eslint.


WanderingHogan:
This rips out the vue-scroll-lock directive we were using for one that lets us watch for events for when a user scrolls to the top or bottom.

It also sets the meta html tag to have a min and max zoom. The user is able to manually pinch to zoom, but overrides the default safari behavior of zooming in when clicking a text input.

This shouldn't conflict with the other scroll bar PR. This library does have its own overrides to hide the scroll bars in this component.

WanderingHogan:
closing this until i figure out mousewheel issue, will re-open shortly

PR #25, opened 2021-07-06
Moved new folder modal + new modal system

jpanay:


RetroPronghorn:
Looks good but make sure you're using eslint there are a few formatting errors I'm seeing that it should have complained about.

PR #34, opened 2021-07-09
SA34 fixed friends reloading every refresh it resolves SA-33 jazzicon undefined issue automatically

nickjjang:


store/friends/actions.ts
```diff
@@ -3,13 +3,16 @@ import { Friends } from '~/mock/friends'
...
 }
 export default {
   handler: () => {},
-  async fetchFriends({ commit }: FetchFriendsArguments) {
-    commit('loading', 'friends')
-    await new Promise((resolve) => setTimeout(resolve, 3000))
-    commit('fetchFriends', Friends)
-    commit('load', 'friends')
+  async fetchFriends({ commit, state }: FetchFriendsArguments) {
+    // commit('loading', 'friends')
```
RetroPronghorn, line 11: let's just remove this if it's not needed.

RetroPronghorn:
This should be a fine fix after the commented code is removed, Just a note for the future, we should consider an "updating" state so we can display a spinner in the UI. Will open a jira task for this.

PR #36, opened 2021-07-12
Fixed issue and made CSS improvements

jpanay:


components/tailored/core/media/user/User.less
```diff
@@ -13,6 +13,7 @@
     position: absolute;
     top: calc(50% - 38.5px);
     left: calc(50% - 32.5px);
+    z-index: 10;
```
RetroPronghorn, line 16: For this, let's use the base index and add one. This will prevent us arbitrarily setting z index in the app. for example for tag use `z-index: calc(@base-z-index  + 1);` and for the other just use the `z-index: @base-z-index;`

components/tailored/core/media/Media.less
```diff
@@ -13,12 +13,16 @@
     align-items: center;
     flex: 1;
     width: 100%;
-    overflow: hidden;
+    overflow: scroll;
```
RetroPronghorn, line 16: What does this accomplish, just want to be sure we actually want it to scroll.

RetroPronghorn (changes requested):

components/tailored/core/media/Media.less
jpanay: Makes it so that when it's not fullscreen and there's a lot of users you can scroll horizontally to see everyone instead of just having it cut off by the window

components/tailored/core/media/user/User.less
jpanay: fixed

components/tailored/core/media/Media.less
RetroPronghorn: let's remove this as it's addressed in a different ticket in Jira with a different solution since scrolling is not optimal

jpanay: done

PR #38, opened 2021-07-13
feat(servers): introducing new modal for server creation

IvanM3x:
Introducing new modal for server creation with new component to crop images.


components/tailored/core/imageCropper/ImageCropper.less
```diff
@@ -0,0 +1,57 @@
...
+  padding: 35px;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
+  margin: auto;
+  width: 380px;
+  height: 380px;
+  box-shadow: 0 3px 6px rgba(0, 0, 0, 0.16), 0 3px 6px rgba(0, 0, 0, 0.23);
+  animation: zoom-in-zoom-out 0.3s ease-out;
```
RetroPronghorn, line 44: Please use the animation speed global here. It should be in the globals.less

components/tailored/core/servers/create/Create.html
```diff
@@ -0,0 +1,49 @@
+<div class="modal is-active">
+    <div class="modal-background"></div>
+    <div class="modal-card">
+        <section class="modal-card-body">
+            <div class="modal-heading">
+                <span class="close" v-on:click="close">×</span>
```
RetroPronghorn, line 6: Let's create a global component under components/ui for the modal, you should be able to then just do something like <UiModal>...yourcodehere</UiModal>

components/tailored/core/servers/list/List.vue
```diff
@@ -17,6 +17,11 @@ export default Vue.extend({
         return []
       },
     },
+    openModal: {
+      type: Function,
+      default: undefined,
```
RetroPronghorn, line 22: set default to `() => {}`

RetroPronghorn (changes requested):
A few notes, otherwise looks good.

PR #45, opened 2021-07-20
feat(sounds): add SoundManager class

molimauro:
Jira ref: SA-43


plugins/local/classLoader.ts
```diff
@@ -1,22 +1,26 @@
...
-import WebRTC from '~/utilities/WebRTC/WebRTC'
 import SolanaManager from '~/utilities/SolanaManager/SolanaManager'
+import SoundManager from '~/utilities/SoundManager/SoundManager'
+import WebRTC from '~/utilities/WebRTC/WebRTC'
 
 declare module 'vue/types/vue' {
   interface Vue {
     $WebRTC: typeof WebRTC
     $SolanaManager: typeof SolanaManager
+    $sounds: SoundManager
```
RetroPronghorn, line 12: Let's keep classes title case so simply change to `$Sounds`

plugins/local/classLoader.ts
```diff
@@ -1,22 +1,26 @@
...
   }
 }
 
 declare module '@nuxt/types' {
   interface Context {
     $WebRTC: typeof WebRTC
     $SolanaManager: typeof SolanaManager
+    $sounds: SoundManager
```
RetroPronghorn, line 20: Same thing here, `$Sounds`

utilities/SoundManager/SoundManager.ts
```diff
@@ -0,0 +1,85 @@
+import { Howl } from 'howler'
+import { Config } from '~/config'
+
+// Keep this type in sync with Config.sounds
+export enum Sounds {
+  NewMessage = 'newMessage',
```
RetroPronghorn, line 6: Let's make enums all uppercase, I don't know if it's a standard but it's what I've been doing for the other enums so it's clear they are enums without digging in. so `NEW_MESSAGE` `CALL` `HANGUP`.. etc.

utilities/SoundManager/SoundManager.ts
```diff
@@ -0,0 +1,85 @@
...
+   * @constructs SoundManager
+   */
+  constructor() {
+    this.sounds = {} as Record<Sounds, Howl>
+    for (const [key, value] of Object.entries(Config.sounds) as [
+      Sounds,
+      string
+    ][]) {
+      this.sounds[key] = new Howl({
+        src: [`${Config.ipfs.browser}${value}`],
```
RetroPronghorn, line 33: let's change this in the config and here to `Config.ips.gateway` this was a terminology mistake I made early on.

utilities/SoundManager/SoundManager.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /** @function
+   * Plays a specified sound
+   * @name play
+   * @argument sound Name of the sound file to play.
+   * @returns null
+   */
+  playSound(sound: Sounds) {
+    if (!this.sounds[sound]) {
```
RetroPronghorn, line 48: We use this logic a few times, is there a way we can make a private method to check if the sound exists and if not throw the error and stop execution?

components/tailored/core/media/actions/Actions.vue
```diff
@@ -3,11 +3,21 @@
...
 
 export default Vue.extend({
   computed: {
     ...mapState(['audio']),
   },
+  methods: {
+    toggleMute() {
+      const muted = this.$store.state.audio.muted
+      if (!muted) this.$sounds.playSound(Sounds.Mute)
+      if (muted) this.$sounds.playSound(Sounds.Unmute)
```
RetroPronghorn, line 16: let's use else logic here

RetroPronghorn (changes requested):

molimauro:
PR updated! Let me know if this is ok ;)

RetroPronghorn (approved):

PR #48, opened 2021-07-21
feat(UI): introducing new scroll area component

IvanM3x:
New component for wrapping scroll areas


RetroPronghorn:
One small change here, can you implement this component to replace the scrolling section of the chat? Additionally if possible please insure the v-body-scroll behavior works on this. I.e. we need to go on an iOS device and scroll til it bounces then try to scroll again. This breaks with normal scrolling so we add the v-body-scroll plugin to fix it. See other scrolling elements for an example.

IvanM3x:
> One small change here, can you implement this component to replace the scrolling section of the chat? Additionally if possible please insure the v-body-scroll behavior works on this. I.e. we need to go on an iOS device and scroll til it bounces then try to scroll again. This breaks with normal scrolling so we add the v-body-scroll plugin to fix it. See other scrolling elements for an example.


What is the correct behavior for you? I saw from the other components that the "v-scroll-lock" prop prevents bouncing, the component I created allows you to scroll the contents of a div without ever bouncing (tested on IOS)

RetroPronghorn:
Okay as long as it doesn't glitch out on iOS we're fine, could we implement it on one of the scroll areas so WA can test I'll create a new issue to implement it everywhere.

PR #51, opened 2021-07-22
fix(media): added button to trigger the modal of the call

IvanM3x:
Added temporary button in the server list to simulate an incoming call


RetroPronghorn (changes requested):

IvanM3x:
To test this component you must first clear the application cache, to allow the status of the incoming call to be effectively blacklisted

PR #54, opened 2021-07-23
SA-49: Add "encrypted message" message to the top of direct/server chat

KemoPaw:
Top of message history it should have the notice that messages are encrypted. Please see the old app for an example of this.


pages/chat/direct/Direct.html
```diff
@@ -1,3 +1,8 @@
 <div id="direct">
+  <div id="encrypted-yellow">
+    <font-awesome-icon :icon="['fas', 'key-skeleton']" />
```
RetroPronghorn, line 3: Small change, let's not add the solid icons yet, remove them from the package.JSON and use the outline version for now

RetroPronghorn (changes requested):

pages/chat/direct/Direct.html
KemoPaw: Okay, working on the fixes now!

KemoPaw: @RetroPronghorn just fixed it!

PR #59, opened 2021-07-28
feature(enhancers): introducing global custom directive for handle co…

IvanM3x:
introducing global custom directive for handle component outside click and  enhancers fullscreen layout for mobile

### DIFF
--- a/utilities/directives.ts
+++ b/utilities/directives.ts
@@ -1,0 +1,27 @@
+import Vue from 'vue'
+
+// Initialize the annoying-background directive.
+export const closeOutside = {
+  bind: function (el: any, binding: any) {
+    // Define Handler and cache it on the element
+    const bubble = binding.modifiers.bubble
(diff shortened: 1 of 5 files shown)


utilities/directives.ts
```diff
@@ -0,0 +1,27 @@
+import Vue from 'vue'
```
RetroPronghorn, line 1: What is the need for this vs the click outside handler we already have in plugins? Also this should be a plugin if we still need it. https://github.com/Satellite-im/Satellite-Absolute/blob/main/plugins/thirdparty/clickoutside.ts see this link.

RetroPronghorn (changes requested):

RetroPronghorn:
https://github.com/Satellite-im/Satellite-Absolute/commit/c8510e986d3b4ea6afecb7d64aca80816f02327b

PR #61, opened 2021-07-29
SA-45 Wire Volume Slider done, SA-46 Smart Profile positioning fixed

nickjjang:


RetroPronghorn:
@nickjjang merge conflict, then should be good. Let's also delete instead of commenting out un-needed code

PR #63, opened 2021-07-29
feat(sidebar): server and user lists made scrollable with visible too…

IvanM3x:
feat(sidebar): server and user lists made scrollable with visible tooltips


IvanM3x:
I was able to implement the feature without upsetting the structure of the components, but this implies a fixed height for the list of users and servers

PR #86, opened 2021-08-12
Eric/SA-94

genie-magic:
Created pull request about AutoScroll component & its example on components.


components/ui/AutoScroll/AutoScroll.html
```diff
@@ -0,0 +1,3 @@
+<div class="auto-scroll" ref="autoScroll" :style="`height: ${height};`" @click="mouseClick" @mouseover="mouseOver">
```
RetroPronghorn, line 1: please extend the existing scroll component so that we keep the same styling and scroll behavior from this.

RetroPronghorn:
We should also be able to send test messages to test the component. Additionally there should be mutations to the mock data so that we can append messages to test, bind this to the chatbar component.

RetroPronghorn (changes requested):

PR #95, opened 2021-08-17
SA 103 : Replace hard coded text

KemoPaw:
replace all hardcoded text to be located in the 'en-US.json' file

### DIFF
--- a/locales/en-US.js
+++ b/locales/en-US.js
@@ -228,7 +228,6 @@ export default {
       video_quality: 'Video Quality',
       no_options: 'No options',
       stream_options: 'Stream Options',
-
     },
   },
 }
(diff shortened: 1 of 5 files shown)


locales/en-US.js
```diff
@@ -209,4 +209,29 @@ export default {
...
+      bitrate: 'Bitrate',
+      sample_size: 'Sample size',
+      video_quality: 'Video Quality',
+      no_options: 'No options',
+      stream_options: 'Stream Options',
+
+    },
+  },
+  sidebar: {
+    friends_label: '14'
```
RetroPronghorn, line 235: we can keep this hard coded since it will be computed in the future.

RetroPronghorn (changes requested):

locales/en-US.js
KemoPaw: @RetroPronghorn Okay!

PR #115, opened 2021-08-30
feat(ProfilePageRedux & Search & Chatbar & UserSearch): added new profile mockup, updated appearances and functions and added user search component

nickjjang:
After added mockup for profile, there are two problems.

- As you can see, there is padding of Settings screen for wrapping children component, so Profile image couldn't match the parent's width like the mockup design.
- I used multiselect component for selecting locations and languages, but it needs to be updated in order to implement your design requirements. May I update it?

I also added components/interactables/Chip for global usage.


nickjjang:
I moved Buttongroup component in marketplace to components and updated it.
You can confirm its usage in marketplace. I have been thinking about it, but it's simple yet.

assets/img/icon_badge1.svg
```diff
@@ -0,0 +1,7 @@
+<svg width="24" height="23" viewBox="0 0 24 23" fill="none" xmlns="http://www.w3.org/2000/svg">
```
RetroPronghorn, line 1: Let's remove these SVGs for now. just use generic font awesome icons. There are a few example badges. I don't want to forget about these placeholder ones when we move to prod.

components/interactables/Chip/Chip.vue
```diff
@@ -30,7 +30,8 @@ export default Vue.extend({
 <style scoped lang="less">
 .chip-item {
   display: inline-block;
-  margin-right: 5px;
+  margin-right: 8px;
+  margin-bottom: 8px;
   padding: 4px 8px;
   border-radius: 24px;
```
RetroPronghorn, line 36: I know it's not part of this ticket, but can we move the border-radius: 24px to a global in the less files? I think we use it in quite a few places where we want rounded buttons.

components/tailored/marketplace/searchbox/Searchbox.html
```diff
@@ -2,7 +2,7 @@
   <InteractablesInput
```
RetroPronghorn, line 2: Also not part of this ticket, but can we open a new ticket in Jira to make the search input it's own standalone component so we may reuse it in other places where we may have different search filters but want the same input

pages/settings/profile/Profile.less
```diff
@@ -5,6 +5,29 @@
...
+.profile-left {
+  flex-shrink: 0;
+  width: 350px;
+  padding: @normal-spacing;
+  margin-top: -200px;
+}
+
+.profile-right {
+  flex-grow: 1;
+  margin-top: 32px;
```
RetroPronghorn, line 22: Let's use either @normal-spacing or @light-spacing here, if you need a bit more multiply the value. This way we can change spacing globally.

types/profile/profile.ts
```diff
@@ -0,0 +1,22 @@
+export enum ProfileBadgeType {
```
RetroPronghorn, line 1: We have existing badges in the app for the badges next to users names, let's reuse those typings. It should be in the config file somewhere if you need a reference.

RetroPronghorn (changes requested):

components/tailored/marketplace/searchbox/Searchbox.html
nickjjang: please open new ticket for this.

assets/img/icon_badge1.svg
nickjjang: I added some font awesome icons seems like badges

components/interactables/Chip/Chip.vue
nickjjang: I added spacing and corner rounding styles such as large, xlarge, xlight and xxlarge

types/profile/profile.ts
nickjjang: I used badges without proper meanings, so in order to include them in config file, should I make them meaningful?

nickjjang:
I added tailored/user/search for User Search component.

RetroPronghorn:
@nickjjang please port this code to the page under the settings component. This should be straightforward.

PR #140, opened 2021-09-13
fix(media): fix media full screen in servers chat and some aside bar …

IvanM3x:
Jira ref: SA-159 + SA-177

### DIFF
--- a/components/tailored/core/group/aside/Aside.less
+++ b/components/tailored/core/group/aside/Aside.less
@@ -62,7 +62,7 @@
       background: @semitransparent-lightest-gradient !important;
       cursor: pointer;
       font-size: @tiny-icon-size !important;
-      
+
       &:hover {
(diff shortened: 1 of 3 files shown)


components/tailored/core/group/aside/Aside.less
```diff
@@ -88,7 +88,7 @@
     position: fixed;
     right: 0px;
     top: 0px;
-    z-index: 2;
+    z-index: 999;
```
RetroPronghorn, line 91: Let's avoid using this, instead please determine what the z-index we need is, and use @base-z-index + required index

RetroPronghorn (changes requested):

components/tailored/core/group/aside/Aside.less
IvanM3x: ok fixed

PR #166, opened 2021-10-01
fix: make the app compatible with custom faucet implementation

iltumio:
Added support for custom faucet implementation for Solana devnet


WanderingHogan:
Faucet reverse proxy/ssl cert have been set up. https://faucet.satellite.one

PR #187, opened 2021-10-19
feat(ChatbarTag&EditLastMsg&ButtonGroups): added charbar's tag, edition of last msg by arrow up and replace nav buttons with button groups

nickjjang:


nickjjang:
still exists spacing issue in chatbar and in order to solve it, I'll model chatbar's content as structured data and implement with remarkable component.

PR #223, opened 2021-11-02
Fix(iOS:Bitrate/Samplesize): Fix Bitrate/Sample Size dropdowns stretch across entire screen(SA-315)

nickjjang:


components/tailored/core/media/actions/settings/Settings.vue
```diff
@@ -113,19 +113,24 @@ export default Vue.extend({
...
             clickedItem.classList.remove('open')
             clickedItemBody.style.height = '0'
+            clickedItemBody.style.overflow = 'hidden'
           } else {
             clickedItem.classList.add('open')
             clickedItemBody.style.height =
               (clickedItemBody.children[0] as HTMLElement).offsetHeight +
               2 +
               'px'
+            setTimeout(() => {
```
RetroPronghorn, line 123: we should find a solution that doesn't use setTimeout.

RetroPronghorn (changes requested):

PR #231, opened 2021-11-05
fix(Chat bar): Prevent empty message from being sent when clicking the send button

genie-magic:
Prevent empty message from being sent when clicking the send button


components/tailored/core/chatbar/Chatbar.vue
```diff
@@ -355,6 +355,7 @@ export default Vue.extend({
      * @example v-on:click="sendMessage"
      */
     sendMessage() {
+      if(!this.value) return
```
RetroPronghorn, line 358: Sorry small change but prettier will complain, add a space after `if()` so it's `if ()`

RetroPronghorn (changes requested):

PR #235, opened 2021-11-07
Feature messaging

iltumio:
# ** Breaking Change **
Some changes in friends has been done since we are not sharing the mailbox id, but a textile provided public key for sending messages between users. For that reason the *mailboxId* property in User and Friend objects has been renamed as *textilePublicKey*.

To make it work properly it is needed to clear the storage and to create a new fresh user with new fresh friends.

**Features:**
- Fetching direct conversations
- Sending text messages
- Sending reactions
- Sending replies
- Subscribe to new incoming messages (the subscription lasts 1 minute)

**Known issues:**
- Textile mailbox needs some time to be initialized. The UI does not provide any feedback on that and allows to start using the chat before it is ready. This causes an exception in the console (Textile not yet initialized)
- The subscription will be closed after 1 minute of inactivity


RetroPronghorn:
@iltumio can we add initializing textile to the loading tasks on the loading screen? I'm not sure if we're bypassing that screen right now but we should be moving from the unlock screen to the loading screen so we can initialize all these things.

PR #237, opened 2021-11-08
Fix spacebar doesn't work in chatbar

nickjjang:


RetroPronghorn:
@nickjjang we're reworking the chatbar, things have become very messy with it so I think it's best we revisit it's functionality. We can keep track of this PR for the future but for now going to close until further changes.

PR #245, opened 2021-11-10
fix(Enhancer): Fix 'After opening the Glyphs menu, user should be able to press it again to close it'

genie-magic:
Fix 'After opening the Glyphs menu, user should be able to press it again to close it'

### DIFF
--- a/components/tailored/core/chatbar/Chatbar.html
+++ b/components/tailored/core/chatbar/Chatbar.html
@@ -23,7 +23,6 @@
       @drop="handleDrop"></div>
     <div
       :data-tooltip="$t('wallet.send_money')"
-      v-on:click="toggleEnhancers"
       class="
         control-tooltip-wrap
         has-tooltip has-tooltip-top has-tooltip-primary
@@ -34,6 +33,7 @@
       />
     </div>
     <div
+      id='chat-enhancer-toggle'
       :data-tooltip="$t('global.glyphs')"
       v-on:click="toggleEnhancers"
       class="
--- a/components/tailored/messaging/enhancers/Enhancers.vue
+++ b/components/tailored/messaging/enhancers/Enhancers.vue
@@ -71,10 +71,7 @@ export default Vue.extend({
         })
         this.toggleEnhancers()
       } else {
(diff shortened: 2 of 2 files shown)


components/tailored/messaging/enhancers/Enhancers.vue
```diff
@@ -91,12 +88,16 @@ export default Vue.extend({
      * @description Toggles enhancers by commiting the opposite of it's current value (this.ui.enhancers.show) to toggleEnhancers in state
      * @example v-on:click="toggleEnhancers"
      */
-    toggleEnhancers() {
+    toggleEnhancers(event: Event) {
       this.clickEvent()
+      /* Prevent toggle twice when toggle enhancer button is clicked (for mobile case ) */
+      const chatToggleEnhancer = document.getElementById('chat-enhancer-toggle')
```
RetroPronghorn, line 94: We should use a ref here instead of ID

components/tailored/core/chatbar/Chatbar.html
```diff
@@ -34,6 +33,7 @@
       />
     </div>
     <div
+      id='chat-enhancer-toggle'
```
RetroPronghorn, line 36: switch to ref="chat-enhancer-toggle"

components/tailored/messaging/enhancers/Enhancers.vue
RetroPronghorn: We can use something like 
`const chatToggleEnhancer = this.$refs.chatToggleEnhancer as HTMLElement` if we need to access dom funcitonality

RetroPronghorn (changes requested):

components/tailored/core/chatbar/Chatbar.html
genie-magic: But we can't access to other component's ref.. isn't it?

RetroPronghorn: Hmm yeah, we should instead use the mutation to toggle then.

genie-magic:
Will close this pull request and create with dev

PR #246, opened 2021-11-10
fix issues: Spacebarmarkdown, bitrate/samplesize, searchbar, reactinginchatbar

nickjjang:


libraries/ui/Chatbar.ts
```diff
@@ -38,6 +38,55 @@ export function htmlToMarkdown(htmlString: string) {
...
+ * @param {string} htmlString - The html expression
+ * @param {string} key - The variable for the key event
+ * @returns Makrdown expression
+ */
+export function htmlToMarkdownInchatbar(htmlString: string, key: string) {
+  var temp = htmlString.replaceAll("<span>&nbsp;</span>", ' ')
+  htmlString = temp.replaceAll("&nbsp;", ' ')
+  let markDown = '';
+
+  if (key == 'Backspace') {
```
Jekrimo, line 53: Saw this Backspace string a couple times, should we make a shared enum constant containing position strings or pointless?

Jekrimo (approved):

libraries/ui/Chatbar.ts
nickjjang: Can we leave it for now and then do it in new jira card? these kinds of issues exists in many places.

nickjjang: I'll make a enum and resolve the issues in a new pull request.

PR #249, opened 2021-11-11
Sa 323 - Error Modal Component

Jekrimo:
Move over Error component from Legacy app. Update translate files. Add timeout prop to error


WanderingHogan (approved):
Good, we will customize it in another ticket. Thank you for bringing this over.

RetroPronghorn (approved):

PR #253, opened 2021-11-12
refactor(Keybinds): restructured everything around actions

jpanay:


RetroPronghorn:
Looks good other than the small changes

jpanay:
> Looks good other than the small changes

What's the small changes?

WanderingHogan:
I don't think this is working exactly how we need. There appear to be two issues.

The keybinds to mac are a little wonky, i am guessing its great for a windows keyboards.
Also - the keybinds only trigger if you manually click the buttons that toggle the same sing as the keybinds, and then if you click out of that window, the keyboards no longer get listened for until the component with the button is clicked in again

RetroPronghorn (approved):

PR #258, opened 2021-11-15
fix(Glyph): Fix 'After opening the Glyphs menu, user should be able to press it again to close it'

genie-magic:
Fix 'After opening the Glyphs menu, user should be able to press it again to close it'


genie-magic:
@RetroPronghorn  as you know that, we can't use refs of other component.. that's the reason why I use id here but If you have better solution then please let me know.

Jekrimo (approved):

WanderingHogan (approved):

PR #277, opened 2021-11-17
fix(Chat): Add friends component to new chat modal

genie-magic:
**What this PR does** 📖
Add friends component to new chat modal
SA-370
Fixes #
- Get rid of standard inputs & Add friends components
- @RetroPronghorn  I think we don't need copy logic from Create server modal.
  I already copied logic but decided to remove all because we can share TailoredSettingsPagesUserSearch component on Create Server Modal & New Chat Modal.  If we need some fixes or tweak, we can update TailoredSettingsPagesUserSearch  component.
- For now, group messages are not implemented (Manuel said we have to find a way to share a secret key between all the participants)... so in this case, We need to wait until they solve that..  We can just create group chat by modifying mock up data but that has no meaning.

### DIFF
--- a/components/tailored/friends/quick/Quick.html
+++ b/components/tailored/friends/quick/Quick.html
@@ -1,14 +1,22 @@
 <div id="quick-chat" v-if="$store.state.ui.modals.quickchat">
   <TypographySubtitle :size="6" text="New Chat" />
-  <TypographyText text="Select one or more of your friends to chat in groups, or one-on-one." />
-  <InteractablesInput 
-    type="primary"
+  <TypographyText
+    text="Select one or more of your friends to chat in groups, or one-on-one."
+  />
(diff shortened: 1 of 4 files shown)


Jekrimo (approved):

WanderingHogan (approved):

RetroPronghorn:
To be clear @genie-magic this adds in the ability to type into the search box and select multiple friends?

genie-magic:
@RetroPronghorn 

> To be clear @genie-magic this adds in the ability to type into the search box and select multiple friends?

Yes.  But group chat logic is not implemented yet on Textile side...  So It just add friends and when click 'new chat' no action happen.  I can create group chat after they finished that.

RetroPronghorn (approved):

PR #279, opened 2021-11-17
refactor(Typing Indicator): AP-37 - Typing indicator component now accepts an array of user names displaying up to 3 names, when 4+ names: displays many typing, and 0: nothing

Jekrimo:
**What this PR does** 📖
Typing indicator component now accepts an array of user names displaying up to 3 names, when 4+ names: displays many typing, and 0: nothing. This is a preparation for our user data models, will need more revision once permanent data structure is in place
AP-37
Fixes #
Multiple users typing

**Additional comments** 🎤

### DIFF
--- a/components/ui/TypingIndicator/TypingIndicator.vue
+++ b/components/ui/TypingIndicator/TypingIndicator.vue
@@ -23,11 +23,11 @@ export default Vue.extend({
     checkUsersTyping() {
       if(this.users.length <= 3 && this.users.length != 0) {
         return 1
-      } else if (this.users.length > 3){
-        return 0
-      } else {
+      }
+      if (this.users.length > 3){
         return 2
       }
+      return 0
     }
   },
   mounted() {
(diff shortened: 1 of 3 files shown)


components/ui/TypingIndicator/TypingIndicator.vue
```diff
@@ -2,15 +2,37 @@
...
   },
+    /**
+     * @methods checkUsersTyping
+     * @description Checks the length of the users names array being fed to component
+     * returns a numeric value that the template checks in order to display the correct ui to the user
+     * @example <div id="text-container" v-if="checkUsersTyping() === 1">
+     *          </div>
+     */
+  methods: {
+    checkUsersTyping() {
```
WanderingHogan, line 23: Can you re-order these, so 0 is 0 instead of 3, and do something like this instead of the else: 

```  
checkUsersTyping() {
      if (this.users.length > 3){
        return 2
      }
      if(this.users.length > 0) {
        return 1
      }
      return 0
    }```

WanderingHogan:
can you also updated `components/tailored/core/chatbar/footer/Footer.vue` to pass the fake user into UiTypingIndicator using the users prop, in an array

WanderingHogan (changes requested):

WanderingHogan (approved):

genie-magic (approved):

PR #286, opened 2021-11-19
fix(enhancers): Made enhancers position dependent on where you click enhancers from, if mobile, setting floating to true. hide scrollbars in chrome

WanderingHogan:
**What this PR does** 📖
These changes make the enhancers menu show up close to where the user clicks the enhancer from on desktop. On mobile it defaults to a floating enhancer. This also removes the chrome scrollbar for the enhancer.

Fixes # SA-187

Tested on desktop and through mobile view in chrome devtools. Reload in between aspect ratio/form factor changes. This is mostly taken from the @KemoPaw's recent PR

**Additional comments** 🎤


KemoPaw (approved):
Looks snazzy, code works!

components/tailored/messaging/enhancers/Enhancers.vue
```diff
@@ -100,18 +110,16 @@ export default Vue.extend({
      * @example v-on:click="toggleEnhancers"
      */
     toggleEnhancers(event: Event) {
-      this.clickEvent()
+      // this.clickEvent()
       /* Prevent toggle twice when toggle enhancer button is clicked (for mobile case ) */
       const chatToggleEnhancer = document.getElementById('chat-enhancer-toggle')
-      // @ts-ignore
+      // // @ts-ignore
```
Jekrimo, line 116: Can we just delete this and the this.click above?

Jekrimo (approved):

components/tailored/messaging/enhancers/Enhancers.vue
RetroPronghorn: If possible on this, we should use refs, using document selectors is an antipattern in vue.

store/ui/state.ts
```diff
@@ -26,6 +26,10 @@ const InitalUIState = (): UIState => ({
   enhancers: {
     show: false,
     floating: false,
+    position: [0, 0],
+    defaultWidth: '24rem',
+    defaultHeight: '30rem',
```
RetroPronghorn, line 31: I don't love storing these values in state, can we use global less vars here? Do these need to be dynamic? IF so can we use less selectors for it?

RetroPronghorn (approved):

store/ui/state.ts
WanderingHogan: We moved these to state because we need them in the style and for a calculation to detect if the enhancers component would be off screen

PR #287, opened 2021-11-19
fix(File): Fix style issue when open Anime image on File

genie-magic:
**What this PR does** 📖
Fix 'Looks like some text is showing up behind the Satellite logo while viewing the Anime File'

Jira ref:  AP-73

### DIFF
--- a/components/tailored/files/view/View.less
+++ b/components/tailored/files/view/View.less
@@ -5,7 +5,7 @@
   bottom: 0;
   right: 0;
   background-color: rgba(0, 0, 0, 0.75);
-  z-index: @base-z-index;
+  z-index: @base-z-index + 100;
   align-content: center;
 
   .image-container {


molimauro (approved):

molimauro:
Solution works, although I think we should open a new ticket for revisiting the z-index of the app since now they’re pretty messed up. There are a lot of @base-z-index + 1, +30, +2, +3, etc.. and I find them quite confusing, especially il the long run. The problem with z-index bugs is that they often result in a new z-index bug, like in this case: someone put the sidebar z-index higher in order to resolve a bug, but this led to this new one.
To keep things consistent we should divide the app in layers and use variables.

For example, with three layers we can define some of these vars:
```
@base-z-index: 1
@sidebar: @base-z-index + 1
@image-preview: @sidebar + 1
```

And in `.less` files using like this:
```
#view-file {
	z-index: @image-preview;
}
```
Variables are more meaningful than numbers to me. Now if someone changes the @sidebar z-index, no problems arise in the #view-file one.
Do y'all think this solution can work?

Jekrimo (approved):

WanderingHogan:
@molimauro great idea, i will add a ticket to the backlog of the AP board

stavares843 (approved):
fixed ✅

PR #299, opened 2021-11-23
feat(user): user state component

molimauro:
**What this PR does** 📖
Introduce the userState component and spread it across the app so that the user state is consistent in all components that consume it
AP-30

**Additional comments** 🎤

### DIFF
--- a/components/tailored/friends/add/Add.vue
+++ b/components/tailored/friends/add/Add.vue
@@ -55,14 +55,15 @@ export default Vue.extend({
           this.error = this.$t('friends.not_found') as string
           return
         }
+        // TODO : fix when all information are available
         this.friend = {
           ...friend,
           state: 'offline',
           // @ts-ignore
-          // TODO: fix
           account: {
             accountId: this.accountID,
           },
+          address: this.accountID,
         }
       } catch (error) {
         this.error = this.$t('friends.invalid_id') as string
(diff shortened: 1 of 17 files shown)


Jekrimo (approved):

components/tailored/friends/add/Add.vue
```diff
@@ -55,14 +55,15 @@ export default Vue.extend({
...
           return
         }
+        // TODO : fix when all information are available
         this.friend = {
           ...friend,
           state: 'offline',
           // @ts-ignore
-          // TODO: fix
           account: {
             accountId: this.accountID,
```
RetroPronghorn, line 64: This is unrelated to your PR but we should make a point to just use "id" here instead of accountId. It's in the account object so the fact that it's the accountId can be inferred.

components/tailored/settings/pages/user/Search.vue
```diff
@@ -53,7 +52,7 @@ export default Vue.extend({
   data() {
     return {
       search: '',
-      result: Users,
```
RetroPronghorn, line 56: Can we bind friends to mapState

components/tailored/settings/pages/user/Search.vue
```diff
@@ -113,7 +112,7 @@ export default Vue.extend({
      * @example
      */
     searchResult() {
-      this.result = Users.filter((user) =>
```
RetroPronghorn, line 116: mapState(['friends'])

RetroPronghorn (changes requested):

components/tailored/friends/add/Add.vue
molimauro: yeah, I saw you reported this on channel. Closing 👍

RetroPronghorn (approved):

PR #302, opened 2021-11-23
Fix(Login): Unlock should be persisted when store pin is enabled

nickjjang:
https://satellite-im.atlassian.net/browse/AP-22

### DIFF
--- a/pages/auth/unlock/index.vue
+++ b/pages/auth/unlock/index.vue
@@ -32,13 +32,18 @@ export default Vue.extend({
     ...mapGetters('accounts', ['getPinHash', 'getPhrase']),
     ...mapState(['accounts']),
     storePin: {
-      set(state) {
-        this.$store.commit('accounts/setStorePin', state)
+      async set(state) {
+        await this.$store.commit('accounts/setStorePin', state)
+        console.log(JSON.stringify(this.$store.state.prerequisites))
       },
       get() {
         return !this.accounts ? false : this.accounts.storePin
       },
     },
+  },
+  mounted() {
+    this.$store.commit('accounts/lock')
+    this.$store.commit('prerequisites/resetState')
   },
   methods: {
     /**
--- a/pages/auth/unlock/index.vue
+++ b/pages/auth/unlock/index.vue
@@ -32,9 +32,8 @@ export default Vue.extend({
     ...mapGetters('accounts', ['getPinHash', 'getPhrase']),
     ...mapState(['accounts']),
     storePin: {
-      async set(state) {
-        await this.$store.commit('accounts/setStorePin', state)
-        console.log(JSON.stringify(this.$store.state.prerequisites))
+      set(state) {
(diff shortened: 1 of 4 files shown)


stavares843:
quick question, how long is the PIN going to be stored? 30 days?

nickjjang:
Hey @stavares843
It uses localStorage so the variable is stored forever for now.

stavares843:
thanks!

stavares843 (changes requested):
the PIN is persisting 🎉 

2 things
- When you go directly to /#/auth/unlock still ask the PIN, the same doesn't happen on other URL's
- I'm seeing some issues after entering the PIN, getting stuck on linking, no relevant errors on the console, could you take a look if could be related to these code changes? I don't have this issue on the dev branch, I also rebooted the mac and deployed it again in case it was some local issue.

https://user-images.githubusercontent.com/29093946/143098924-6fdd45e5-9be4-4316-8fd6-fa98221b8135.mov

nickjjang:
Hi @stavares843 I fixed the issues you mentioned.

pages/auth/unlock/index.vue
```diff
@@ -32,14 +32,19 @@ export default Vue.extend({
     ...mapGetters('accounts', ['getPinHash', 'getPhrase']),
     ...mapState(['accounts']),
     storePin: {
-      set(state) {
-        this.$store.commit('accounts/setStorePin', state)
+      async set(state) {
+        await this.$store.commit('accounts/setStorePin', state)
+        console.log(JSON.stringify(this.$store.state.prerequisites))
```
RetroPronghorn, line 37: remove

RetroPronghorn (changes requested):

pages/auth/unlock/index.vue
RetroPronghorn: after resolved looks good code wise.

plugins/thirdparty/persist.ts
```diff
@@ -28,13 +28,19 @@ const propertiesBlacklist = [
   'prerequisites',
 ]
 
+const propertiesBlacklistWhenStorePin = ['prerequisites']
```
molimauro, line 31: Can you add also 'friends.all' to `propertiesBlacklistWhenStorePin ` so that we do not store friends in the state? Thanks!

nickjjang:
Hi @RetroPronghorn, @molimauro 
I've fixed the issues you requested.

plugins/thirdparty/persist.ts
nickjjang: Fixed.

pages/auth/unlock/index.vue
nickjjang: Sorry for inconvemient.
I've removed colsole.log

RetroPronghorn (approved):

molimauro (approved):

PR #308, opened 2021-11-24
fix - Duplicate friends are being added when page is re-rendered - AP-64

davidchan1219:
**What this PR does** 📖
PR request david/fixing to dev branch
store/friends/actions.js file
Fixes #
I fixed duplicated friends issue.
I thought duplicated issue is because addFriend same friend object when visit friends page.
But it was not that problem.
It is updateFriend method issue.

**Additional comments** 🎤

### DIFF
--- a/store/friends/actions.ts
+++ b/store/friends/actions.ts
@@ -140,11 +140,8 @@ export default {
     }
 
     const friendExists = state.all.find((fr) => fr.address === friend.address)
-
-    if (friendExists) {
-      commit('updateFriend', friend)
-    } else {
-      commit('addFriend', friend)
+    if (!friendExists) {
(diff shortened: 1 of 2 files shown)


store/friends/actions.ts
```diff
@@ -155,6 +152,8 @@ export default {
         'removeOutgoingRequest',
         friendAccountToOutgoingRequest(friendAccount).requestId
       )
```
RetroPronghorn, line 154: can we add a return here, then remove the else and just use `commit('updateFriend', friend)`

also remove any usage of `;` from your code.

RetroPronghorn (changes requested):

RetroPronghorn (approved):

PR #309, opened 2021-11-24
fix(Drop File): Fix 'Chat -  If you copy paste a invalid image, gets stuck on scanning image'

genie-magic:
**What this PR does** 📖
Fix 'Chat -  If you copy paste a invalid image, gets stuck on scanning image'

Fixes # AP-90

### DIFF
--- a/components/tailored/core/fileupload/FileUpload.html
+++ b/components/tailored/core/fileupload/FileUpload.html
@@ -35,7 +35,7 @@
             />
             <TypographySubtitle
               :size="6"
-              text="Upload Blocked"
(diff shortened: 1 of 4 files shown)


stavares843 (commented):
Captura de ecrã 2021-11-24, às 17 59 00


neat! 🎉 

maybe we could remove the scanning image copy when the blocked copy appears and center vertically the block icon to the error message, @WanderingHogan thoughts? 

We can add separate tickets to handle later or address on the PR.

jpanay (approved):
Slash icon could look better if it came before the text but I'll let that be your creative choice. Will be good to pull once you make the change sara requested

stavares843 (approved):
image

neat! 🎉

PR #311, opened 2021-11-24
implements lazy loading for glyphs - AP-101

Jekrimo:
**What this PR does** 📖
implements lazy loading for glyphs

AP-101

Fixes #

Can definetely see the lazy loading happening in the network calls and scrolling itself, we DO still have some failure api calls, 429 (Too many Requests) that will need another ticket, although, I think that is from a specific api or two as not every glyph group has that happen so may not be on our side but will most likely still require a fix from us.

**Additional comments** 🎤


components/tailored/marketplace/glyphs/Glyphs.vue
```diff
@@ -27,13 +27,38 @@ export default Vue.extend({
...
         view: GlyphMarketViewStatus.HOME,
         shopId: null,
       })
     },
+    getGlyphs() {
+
+    },
+    onScroll() {
+      const glyphs = this.$refs["marketGlyphs"]
+      if(glyphs) {
```
RetroPronghorn, line 51: can we nuke this?

components/tailored/marketplace/glyphs/Glyphs.vue
```diff
@@ -27,13 +27,38 @@ export default Vue.extend({
...
+    window.removeEventListener('scroll', this.onScroll)
+  },
   methods: {
     goBack() {
       this.$store.commit('ui/setGlyphMarketplaceView', {
         view: GlyphMarketViewStatus.HOME,
         shopId: null,
       })
     },
+    getGlyphs() {
```
RetroPronghorn, line 46: can we also nuke this?

RetroPronghorn (changes requested):

components/tailored/marketplace/glyphs/Glyphs.vue
Jekrimo: Good catch!

RetroPronghorn (approved):

PR #324, opened 2021-11-27
feat(webrtc): add low level abstractions for handling p2p connections and audio/video calls

iltumio:
**What this PR does** 📖
Adds abstractions for low level p2p connections

No issues, it's a new feature

I had troubles with the new eslint configuration. I had to add a specific rule for prettier to make it work.
The p2pt library that has been installed comes from our internal repo Satellite-im/P2PT. It should not cause issues, but it's better to know, just in case.
```yarn install``` command is required

**Additional comments** 🎤
A low level class named Wire can be used to connect two different peers on a specific channel in an easy way. It automatically makes the connection happen using Web Torrent signaling servers. An higher level class named Call can be used to initiate an audio/video conversation between two different peers. It makes use of a previously created Wire for the signaling part. Both classes are supposed to be combined together for creating the top level abstractions WebRTCUser and WebRTCManager.

### CODE REVIEW

File: libraries/WebRTC/Emitter.ts [hunk 1]
```diff
@@ -1,66 +1,64 @@
...
+    }
+
+    this._events[event]!.push(listener)
   }
 
   /**
    * @method off
    * @description Removes an event listener from the listener box
-   * @param event WebRTCEvent to unsubscribe from
+   * @param event event to unsubscribe from
```
stavares843, line 29: event 2x

libraries/WebRTC/Wire.ts
```diff
@@ -0,0 +1,242 @@
...
+  identifier: string
+  channel: string
+  sendIdentification: boolean
+
+  instance: P2PT // P2PT Instance used for connecting peers through Web Torrent signaling servers
+  peer?: Peer.Instance // Simple Peer object instantiated right after the connection occurred through p2pt library
+
+  /**
+   * @constructor
+   * @param originator Idendifier of the originator
```
stavares843, line 33: typo - Idendifier

RetroPronghorn (approved):

PR #327, opened 2021-11-29
feat(import account): 'Create Import Account Page, linked from Import Account button'

genie-magic:
**What this PR does** 📖
This PR is draft preparation PR for Import account bip39.
Create import account button and implement basic ui/ux
Jira ref: AP-15
@WanderingHogan 
1. Do we need to update InteractablesInputGroup component for showing limited bip39 words?
2. When we input all 12 phrases and click recover button, which function do I need to call ?
3. Do we need to add option to remove added phrases?


Jekrimo (approved):

WanderingHogan:
Per your notes:
1. We don't want to show the words, but we want to limit what the typeahead lets them click to be just bip39 words
2. I am not sure of the exact function - I think it is `loadAccount` from `store/accounts/actions.ts`, but we might need to add a method in here. If that is the case can you make another subtasks ticket for it?
3. Yes - the user needs to be able to remove words that they didn't intend to type. That word should be added back to the list of potential words they can pick

WanderingHogan (changes requested):
Great work, works well. One change - `middleware/authenticated.ts` forces you to be routed to /setup/disclaimer if you don't have a phrase and are not locked, update that to allow navigating to the import account page

WanderingHogan:
Also - I was wrong re: `1. We don't want to show the words, but we want to limit what the typeahead lets them click to be just bip39 words` but you wont have the data to make that until you do the bip39 service ticket, and the new AP-154 typeahead ticket

WanderingHogan (approved):

.eslintrc.js
```diff
@@ -16,7 +16,6 @@ module.exports = {
     'vue/script-setup-uses-vars': 0,
     'import/named': 0,
     'standard/no-callback-literal': 0,
-    'comma-dangle': ['error', 'never'],
```
RetroPronghorn, line 19: why was this removed?

components/tailored/messaging/message/actions/Actions.html
```diff
@@ -1,4 +1,5 @@
 <div class="message-actions">
+  <div class="reply-text">{{$dayjs(message.at).from()}}</div>
```
RetroPronghorn, line 2: Small change, please add a newline and add spacing after `{{`

RetroPronghorn (changes requested):
Can you rebase to dev, this is hard to read which changes are yours vs stuff from other PRs

PR #328, opened 2021-11-29
chore(textile-store): update textile convo to have timestamps on top level

WanderingHogan:
Added timestamps for user and convo on conversation object

#AP-125

**What this PR does** 📖
This adds two timestamps to the conversations textile object, which will make it easier for UI updates in the sidebar (sorting based on last message sent/received, show timestamp for last time friend responded, etc)

Fixes #
AP-125

No UI changes, you can see the object being updated in the debugger > application > local storage > satellite-store > and see the new lastMsgReceived and lastConvoUpdate keys in the textile conversations object.

**Additional comments** 🎤

### DIFF
--- a/store/textile/actions.ts
+++ b/store/textile/actions.ts
@@ -154,6 +154,7 @@ export default {
 
       commit('addMessageToConversation', {
         address: sender.address,
+        sender: 'other',
         message,
       })
     })
@@ -213,6 +214,7 @@ export default {
 
     commit('addMessageToConversation', {
       address: friend.address,
+      sender: 'self',
       message: result,
     })
   },
@@ -252,6 +254,7 @@ export default {
 
     commit('addMessageToConversation', {
       address: friend.address,
+      sender: 'self',
       message: result,
     })
   },
@@ -291,6 +294,7 @@ export default {
 
     commit('addMessageToConversation', {
       address: friend.address,
+      sender: 'self',
       message: result,
     })
   },
--- a/store/textile/actions.ts
+++ b/store/textile/actions.ts
@@ -4,6 +4,7 @@ import { ActionsArguments } from '~/types/store/store'
 import TextileManager from '~/libraries/Textile/TextileManager'
 import { TextileConfig } from '~/types/textile/manager'
 import { MailboxManager } from '~/libraries/Textile/MailboxManager'
+import { MessageRouteEnum } from '~/libraries/Enums/enums'
 import { Config } from '~/config'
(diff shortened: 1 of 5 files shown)


store/textile/actions.ts
```diff
@@ -154,6 +154,7 @@ export default {
 
       commit('addMessageToConversation', {
         address: sender.address,
+        sender: 'other',
```
RetroPronghorn, line 157: here, instead let's use ENUMS and instead of sender let's do something like 

route: MessageRoutes.INBOUND for messages from others and 
route: MessageROutes.OUTBOUND for messages we sent

store/textile/actions.ts
```diff
@@ -291,6 +294,7 @@ export default {
 
     commit('addMessageToConversation', {
       address: friend.address,
+      sender: 'self',
```
RetroPronghorn, line 297: route: MessageRoutes.OUTBOUND

store/textile/actions.ts
```diff
@@ -252,6 +254,7 @@ export default {
...
```
RetroPronghorn, line 257: route: MessageRoutes.OUTBOUND

store/textile/actions.ts
```diff
@@ -213,6 +214,7 @@ export default {
...
```
RetroPronghorn, line 217: route: MessageRoutes.OUTBOUND

store/textile/actions.ts
RetroPronghorn: route: MessageRoutes.INBOUND

store/textile/mutations.ts
```diff
@@ -49,6 +53,8 @@ const mutations = {
         messages: {},
         replies: {},
         reactions: {},
+        lastMsgReceived: 0,
```
RetroPronghorn, line 56: how about `lastInbound` and `lastUpdate` also let's comment (Hogan and I talked outside of this the verbiage still might not be enough of an explanation so let's add comment's explaining.

store/textile/types.ts
```diff
@@ -11,6 +11,8 @@ export interface TextileState {
       messages: MessagesTracker
       replies: RepliesTracker
       reactions: ReactionsTracker
+      lastMsgReceived: number // the last time a message was received by any member of conversation other than account owner
```
RetroPronghorn, line 14: Basically let's bring these comments to anywhere it's referenced. Better to have more comments since the build task prunes them anyways.

RetroPronghorn (changes requested):

RetroPronghorn (approved):

PR #457, opened 2021-12-03
fix(chatbar): replies style

molimauro:
**What this PR does** 📖
* Fix a style bug on the chatbar when a reply is in place;
* Remove a reply when switching conversations;
* Does not persist the reply in the store.

I noticed the bug while working, no JIRA ticket assigned.

**Additional comments** 🎤
We should persist the replies when switching conversations but this needs a refactor in the way we store replies. Please open a ticket to keep track of this, thanks!


Jekrimo (approved):

molimauro:
I have to focus the chatbar when hitting reply so please wait before merge, thanks!

stavares843:
converted to draft based on the last comment to avoid this being merged

PR #463, opened 2021-12-04
feat(ChangeLog): update changelog and related ui #AP-25

WanderingHogan:
**What this PR does** 📖
Mostly, this PR updates the build github action to generate the changelog and related UI updates to put it in the web app.
It also does a bunch of small changes to things I saw while I was updating the changelog:
In StatusBar and Global components
  - removed redundant methods to toggle modals on and off, added enums
In Global component
  - Split from single vue file to vue file with template
  - moved the check for updates logic to trigger showing/hiding the modal with changelog here instead of in the modal for the changelog so we could use that modal multiple places
Login Page
  - Made change modal show/hide by clicking the app version at the bottom
UpdateModal
  - removed logic to show/hide based on changes (moved to the global component) so this modal could be re-used more easily
  - commented out logic to upgrade until we can test it. It is clearing the local storage which we don't have a way for users to recover
ReleaseNotes utility
  - Took this fetch and moved it to a library so it can be re-used, and so we don't do more ajax requests than we need to
ConsoleWarning Utility
  - Moved the console warning from unlock page mounted to it's own utility class
DeveloperSettings Component
  - Chrome things all macs are intel macs, added webgl check to figure out if user is on arm cpu instead of apple intel
  - added changelog
Other
  - Updated inline method documentation I touched
  - Removed BuildNative GH-Action in favor of just one for changelog
  - Made modals use enums
  - changed showMarketplace modal name to marketplace

Fixes #AP-25

The UI updates can be seen clicking the version number on the unlock screen, if there is a version mismatch between when you last logged in and the new version, and in the settings for the developer.

**Additional comments** 🎤


components/ui/Global/Global.vue
```diff
@@ -1,92 +1,62 @@
...
+}
 
 export default Vue.extend({
   name: 'Global',
   computed: {
     ...mapState(['ui', 'media']),
+    ModalWindows: () => ModalWindows,
+  },
+  mounted() {
+    // This determines if we should show the
```
stavares843, line 20: if we should show the - seems is missing something 📖

components/ui/Global/Global.vue
```diff
@@ -1,92 +1,62 @@
...
+      lsVersion = this.$config.clientVersion
+      return
+    }
+
+    const [majorVersion, minorVersion, patchVersion] =
+      this.$config.clientVersion.split('.')
+    const [lsMajorVersion, lsMinorVersion, lsPatchVersion] =
+      lsVersion.split('.')
+
+    // // A update which requires resetting of the app has occured.
```
stavares843, line 34: not sure if the double // is on purpose 📖

stavares843 (approved):
pretty neat! 🎉 

very detailed PR description 🔨 

this will be tested after merging

RetroPronghorn (approved):

PR #466, opened 2021-12-06
Logger

RetroPronghorn:
**What this PR does** 📖

This method Logger be used for ALL logging you plan to leave in the application. 
 DO NOT use normal logging, the use of this class will make it easier to enable / disable different log levels.

### DIFF
--- a/utilities/Hounddog.ts
+++ b/utilities/Hounddog.ts
@@ -39,4 +39,9 @@ export default class Hounddog {
             return f.activeChat === true
         })
     }
+
+    friendExists(state: FriendsState, friend: Friend): Boolean {
+        const search = state.all.find((fr: Friend) => fr.address === friend.address)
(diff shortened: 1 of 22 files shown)


utilities/Hounddog.ts
```diff
@@ -0,0 +1,42 @@
...
+        return state.all.find(f => {
+            return f.name === identifier ||
+                f.address === identifier ||
+                f.account.accountId === identifier ||
+                f.textilePubkey === identifier || 
+                f.name === identifier
+        })
+    }
+
+    getActiveFriend(state: FriendsState): Friend | undefined {
```
WanderingHogan, line 37: why not load the friends from the store here instead of passing it in?

Jekrimo (approved):

stavares843 (approved):

PR #467, opened 2021-12-06
fix(upload): fix upload for large files

josephmcg:
nsfw check was hogging too much memory and crashing for large files. skips check above 8MB
closes #91

reopening previous PR. I messed the old branch up, recreated

**What this PR does** 📖

Fixes #

**Additional comments** 🎤

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -92,4 +92,8 @@ export const Config = {
   routingMiddleware: {
     prerequisitesCheckBypass: ['auth', 'setup'],
   },
+  uploadByteLimit: 1048576 * 8, // 8mb
+  regex: {
+    image: '^.*.(apng|avif|gif|jpg|jpeg|jfif|pjpeg|pjp|png|svg|webp)$',
+  },
 }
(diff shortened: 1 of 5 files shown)


config.ts
```diff
@@ -92,4 +92,8 @@ export const Config = {
   routingMiddleware: {
     prerequisitesCheckBypass: ['auth', 'setup'],
   },
+  uploadByteLimit: 1048576 * 8, // 8mb
```
RetroPronghorn, line 95: Is this the max we can check? I'd like to at least try to check the file size of a cell camera. Something like 70mb may be a safer bet, thoughts?

josephmcg: https://browserleaks.com/webgl
Mine was starting to crash around 15mb images.

We could check based on resolution rather than file size? I'm not sure where to draw the cutoff though

Max Viewport Dimensions | [16384, 16384]

RetroPronghorn: Hmm that's fair, we can just address the exact size later. I'm curious if other platforms even check.

RetroPronghorn (approved):

molimauro (approved):

stavares843 (approved):
fixed ✅

PR #474, opened 2021-12-06
fix(homeIconOnFiles): adjust home icon down to avoid icon clash

KemoPaw:
**What this PR does** 📖

https://satellite-im.atlassian.net/browse/AP-156

Fixes #

**Additional comments** 🎤

### DIFF
--- a/components/tailored/files/filepath/Filepath.less
+++ b/components/tailored/files/filepath/Filepath.less
@@ -1,7 +1,6 @@
 .breadcrumb {
   font-family: @primary-font;
   font-size: @text-size;
-  padding-top: @xlarge-spacing + @light-spacing;
   a {
     text-transform: capitalize;
     color: @primary-color;
@@ -23,4 +22,8 @@
     min-width: 100%;
     padding-top: @light-spacing;
   }
+}
+
+.blue-bg {
+  background-color: green;
 }
--- a/components/tailored/files/filepath/Filepath.less
+++ b/components/tailored/files/filepath/Filepath.less
@@ -11,7 +11,7 @@
   li + li::before {
     color: @text;
   }
-  
+
   li.is-active a {
     color: @text-muted;
   }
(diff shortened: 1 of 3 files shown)


Jekrimo (approved):

stavares843 (approved):
Captura de ecrã 2021-12-06, às 22 28 01

fixed ✅

WanderingHogan (changes requested):
This looks alright when it is collapsed, but when it is not the padding pushes everything down too far
Screen Shot 2021-12-06 at 4 05 19 PM
Screen Shot 2021-12-06 at 4 05 16 PM

Can you do something like watch for the left bar being collapsed, and only applying that padding if that is true?

KemoPaw:
@WanderingHogan That sounds like a good suggestion, will look into implementing it!

josephmcg:
It seems like this was formatted using eslint rather than prettier, is that because of husky? You can configure prettier to format files on save so you don't have to think about it.
image

I think template literals (`${}`) are redundant inside a v-bind.
https://vuejs.org/v2/api/#v-bind

Could be rewritten like this
image

stavares843:
yes husky is using `eslint` instead of prettier, but we look into setup that to use both

we added the `vue` plugin for `eslint` as well, do you feel those template literals (${}) that you mentioned are something that should have been captured by the linter? @josephmcg

stavares843:
@josephmcg added that here
https://github.com/Satellite-im/Core-PWA/pull/479

KemoPaw:
@josephmcg Thank you so much for explaining how to clean up the classes in files.vue + have my code format to Prettier on save!

josephmcg:
@stavares843 
Not necessarily, there is definitely a use case for template literals inside a v-bind. I just felt it was redundant this time.

It could be used to dynamically set attributes
<h3 v-bind:data="`haha-${test}`">{{ test }}</h3>

Thanks for checking!

stavares843:
thanks for clarifying! 🎉

josephmcg:
oh github messed with my code example.. one sec

WanderingHogan (approved):
Good job Kat

PR #565, opened 2021-12-08
feat(import account): recover account from phrases you input

genie-magic:
**What this PR does** 📖
Recover account from phrases you input

Fixes # AP-15

Test Case for reviewers:
1. create an account
2. while creating an account, you should save recover phrases
image
3. after you register and login.. clear localStorage and go to '/'
4. input pin and click import account
5. input recover phrases you saved
6. after clicking import account you can log in with your created account

### DIFF
--- a/pages/setup/InputAccount/index.vue
+++ b/pages/setup/InputAccount/index.vue
@@ -23,8 +23,21 @@ export default Vue.extend({
     },
   },
   methods: {
-    recoverAccount() {
-      /* recover account action will be implemented on bip39 service ticket */
+    /**
+     * @method recoverAccount DocsTODO
+     * @description recover account with 12 mnemonic recover phrases
+     * @param
+     * @example
+     */
+    async recoverAccount() {
+      try {
+        const mnemonic = this.phrases.join(' ')
+        await this.$store.dispatch('accounts/setRecoverMnemonic', mnemonic)
+        await this.$store.dispatch('accounts/loadAccount')
+        this.$router.replace('/chat/direct')
+      } catch (error: any) {
+        this.error = error.message
+      }
     },
     isOdd(num: number) {
       return num % 2
--- a/pages/setup/InputAccount/index.vue
+++ b/pages/setup/InputAccount/index.vue
@@ -26,8 +26,7 @@ export default Vue.extend({
     /**
      * @method recoverAccount DocsTODO
      * @description recover account with 12 mnemonic recover phrases
-     * @param
-     * @example
+     * @example this.recoverAccount()
      */
     async recoverAccount() {
       try {
--- a/pages/setup/InputAccount/index.vue
+++ b/pages/setup/InputAccount/index.vue
@@ -17,11 +17,6 @@ export default Vue.extend({
       bipList: bip39.wordlists.english,
     }
   },
-  computed: {
-    availableBipList() {
-      return this.bipList.filter((elm) => !this.phrases.includes(elm))
-    },
-  },
   methods: {
     /**
      * @method recoverAccount DocsTODO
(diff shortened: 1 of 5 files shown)


store/accounts/actions.ts
```diff
@@ -99,6 +99,28 @@ export default {
 
     commit('setEncryptedPhrase', encryptedPhrase)
   },
+  /**
+   * @method generateWallet DocsTODO
```
iltumio, line 103: wrong method name

pages/setup/InputAccount/index.vue
```diff
@@ -23,8 +23,21 @@ export default Vue.extend({
     },
   },
   methods: {
-    recoverAccount() {
-      /* recover account action will be implemented on bip39 service ticket */
+    /**
+     * @method recoverAccount DocsTODO
+     * @description recover account with 12 mnemonic recover phrases
+     * @param
```
iltumio, line 29: remove @param since is not needed

pages/setup/InputAccount/index.vue
```diff
@@ -23,8 +23,21 @@ export default Vue.extend({
...
+     * @example
```
iltumio, line 30: provide an example like
```
this.recoverAccount()
```

iltumio (changes requested):
The functionality is ok. There are some errors in the comments to be fixed.
From the UX point of view the component does not allow to paste a whole phrase at once. We should open a separate ticket for that

iltumio:
@RetroPronghorn we should allow users to paste the whole phrase at once otherwise they have to input every single word by hand that is not user friendly

stavares843 (changes requested):
Captura de ecrã 2021-12-08, às 15 36 29

on the seed, appeared the same word 2x, birth birth

but when trying to add birth 2x, is not allowed because is already added

check hogan's comment [here](https://discord.com/channels/773445116572073995/859670812856418314/918165565984632872) regarding how to address it

genie-magic:
> Captura de ecrã 2021-12-08, às 15 36 29
> 
> on the seed, appeared the same word 2x, birth birth
> 
> but when trying to add birth 2x, is not allowed because is already added
> 
> check hogan's comment [here](https://discord.com/channels/773445116572073995/859670812856418314/918165565984632872) regarding how to address it

I understand you.. yes. that is problem but not issue on this PR.. that is issue when creating account which populate same word.
What do you think abou that @WanderingHogan ?  We can make separate ticket or something but for this PR, we can skip that.

stavares843:
hogan already had replied that [here](https://discord.com/channels/773445116572073995/859670812856418314/918165565984632872)


to address that on this PR

WanderingHogan:
Alright - so changes -
please make the changes Manuel requested
please make it so when a word is selected it is no longer removed from the list of words, and so that there is no limit on the number of times it is selected in the words list up to the max number of words that can be selected

genie-magic:
@WanderingHogan , @iltumio , @stavares843 
I fixed all you mentioned. plz check and let me know.

stavares843:
great, thanks! 🎉 checking

stavares843:
Captura de ecrã 2021-12-08, às 22 54 42

getting the above error trying to import

WanderingHogan:
@stavares843 I was wrong - the words do have to be in order. We should update the text so the user knows to enter them in order. This works greats when they are in the correct order.

WanderingHogan:
Language changed to tell user to put words in order & error message added in the translations file.

stavares843:
thanks for clarifying!

stavares843:
thanks for adding the locales! 💥

stavares843 (approved):
approved ✅

iltumio (approved):

PR #570, opened 2021-12-08
chore(updates): minor updates throughout

WanderingHogan:
**What this PR does** 📖
Submodule updates stop saying there are updates to the submodules
Moved blankspace regex to config file
Added upload text to global because the same text was going to be in three different places
updated the way a class was updating in view so we didn't need to do a weird string concat thing

Fixes # n/a

**Additional comments** 🎤

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,12 @@
 [submodule "android"]
 	path = android
 	url = https://github.com/Satellite-im/android.git
+	ignore = all
 [submodule "electron"]
 	path = electron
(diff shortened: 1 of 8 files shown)


WanderingHogan:
also updated package.json to check for submodule updates when you start `yarn dev` so it should stay in sync every time you launch dev locally

Jekrimo (approved):

RetroPronghorn (approved):

stavares843:
neat! 🎉

PR #597, opened 2021-12-10
fix(ChatReply): disable links in reply bar if user has one in their name

WanderingHogan:
**What this PR does** 📖
If a user has a link in their name, it was being made linkable. This disables links in the chatbar reply indicator component.

Fixes #

Screen Shot 2021-12-09 at 10 06 05 PM

**Additional comments** 🎤

### DIFF
--- a/components/tailored/core/chatbar/reply/Reply.html
+++ b/components/tailored/core/chatbar/reply/Reply.html
@@ -1,6 +1,10 @@
 <div v-if="ui.replyChatbarContent.id" class="is-chatbar-reply">
-    <VueMarkdown :source="`**${$t('conversation.reply_to')} ${ui.replyChatbarContent.from}:** ${ui.replyChatbarContent.payload}`" class="markdown" />
-    <div class="reply-close" @click="setReplyChatbarContent">
-        <x-icon size="1x" />
-    </div>
+  <VueMarkdown
+    :linkify="false"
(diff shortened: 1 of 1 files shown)


components/tailored/core/chatbar/reply/Reply.html
```diff
@@ -1,6 +1,10 @@
 <div v-if="ui.replyChatbarContent.id" class="is-chatbar-reply">
-    <VueMarkdown :source="`**${$t('conversation.reply_to')} ${ui.replyChatbarContent.from}:** ${ui.replyChatbarContent.payload}`" class="markdown" />
-    <div class="reply-close" @click="setReplyChatbarContent">
-        <x-icon size="1x" />
-    </div>
-</div>
\ No newline at end of file
+  <VueMarkdown
+    :linkify="false"
+    :source="`**${$t('conversation.reply_to')} ${ui.replyChatbarContent.from}:** ${ui.replyChatbarContent.payload}`"
```
RetroPronghorn, line 4: I think this is fine, but maybe in the future we want to exclude the usernames from markdown completely. This way I can't name myself something like `Ma__tt__`

RetroPronghorn (approved):

stavares843 (approved):
fixed ✅ 

Captura de ecrã 2021-12-10, às 13 16 00

PR #604, opened 2021-12-11
fix(importaccount): entire page sits off center on mobile (AP-253)

nickjjang:
AP-253

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "Core-PWA",
-  "version": "0.1.1",
+  "version": "0.1.2",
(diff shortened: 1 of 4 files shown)


stavares843 (changes requested):
_

stavares843 (approved):
fixed

Captura de ecrã 2021-12-11, às 16 38 08

still gets a bit cut off on smaller screens but we can address that on other ticket

WanderingHogan (approved):

stavares843:
fixed on smaller screens as well 🎉

PR #613, opened 2021-12-13
fix: presuggestion not working on mobile devices if word start capitalized (AP-254)

nickjjang:
AP-254

### DIFF
--- a/components/ui/TypeHead/TypeHead.vue
+++ b/components/ui/TypeHead/TypeHead.vue
@@ -1,7 +1,11 @@
 <template src="./TypeHead.html"></template>
 <script lang="ts">
 import Vue, { PropType } from 'vue'
-import { InputSize, InputStyle } from '~/components/interactables/Input/types'
+import {
+  InputSize,
+  InputStyle,
+  InputTextTransform,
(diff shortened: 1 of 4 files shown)


stavares843 (approved):
fixed ✅

josephmcg (approved):

components/ui/TypeHead/TypeHead.vue
```diff
@@ -73,15 +84,22 @@ export default Vue.extend({
   },
   methods: {
     update() {
+      if (this.$props.textTransform === 'lowercase') {
```
WanderingHogan, line 87: Can you remove the pieces where you force lowercase input and instead favor just casting the input to lowercase to compare with the list? I think you can accomplis this in the update with something similar to ```      this.searchList = this.list.filter((item: any) => { 
        return item.indexOf(this.searchText.toLowerCase()) !== -1
      }```

WanderingHogan (changes requested):
Can you make this so the user can still input any case, but we cast it to lower case when comparing to the list?

WanderingHogan (approved):
works good!

stavares843 (approved):

PR #614, opened 2021-12-13
fix: persist the replies when switching conversations - AP-235

davidchan1219:
**What this PR does** 📖

Fixes # AP-235

**Additional comments** 🎤


stavares843 (changes requested):
fixed but this is introducing a bug that doesn't happen on dev


it doesn't allow sending out emojis


https://user-images.githubusercontent.com/29093946/145878678-a7ad28dd-ba3c-434f-ae3e-f818adfdaf50.mp4

stavares843 (changes requested):
always getting this error going to any DM on this branch

Captura de ecrã 2021-12-14, às 17 25 20

components/tailored/core/chatbar/Chatbar.vue
```diff
@@ -51,6 +51,13 @@ export default Vue.extend({
       type: Object as PropType<Friend>,
     },
   },
+  // updated() {
+  //   const chatMsgList = localStorage.getItem('chat-message-list')
+  //   let cMLObj = chatMsgList ? JSON.parse(chatMsgList) : []
+  //   if (this.$props.recipient && cMLObj.some((e: any) => e.id === this.$props.recipient.address)) {
+  //     this.value = cMLObj.find((item: any) => item.id === this.$props.recipient.address).value
+  //   }
+  // },
```
stavares843, line 60: we can burn this commented out code

davidchan1219: Removed that commented part

RetroPronghorn:
@davidchan1219 This approach is a bit too complex for what we need to acheive. Instead let's follow a pattern like so.


Make a new state section for message replies (create a Chat folder in store)

add the correct typings for it like 

```
type ChatState = {
    replies: {
        User.ID: string
    },
    text: {
       User.ID: string
    }
}
```

Make a mutation to add text based on the user we're chatting with and allow us to choose if it's a reply or generic chatbar text.

before the component unmounts call the mutation to store the text

```
beforeUnmount() {
   ...if we have an active reply being typed
       ...call chat mutation to store in reply state
  ...if we have active message in chatbar
       ...call chat mutation to store in text state
 }
 ```
 now when the component mounts, we can simply check the state to see if a reply is being made. We should also clear the reply from state when remounting.
 
 ```
 mounted() {
 ...if store.state.chat.replies[user]
    ...this.$data.reply = store.state.chat.replies[user]
    ...call mutation to clear
 ...if store.state.chat.text[user]
    ...this.store.state.chatbarContent = this.store.state.chat.text[user]
    ...call mutation to clear
}
```

this is psuedo code but please feel free to ask questions if you have any. I think we should start fresh from the dev branch on this. It will be a better approach in the long run.

stavares843:
the crash is fixed ✅ 

persisting message bug is fixed ✅ 

sending emojis works ✅

stavares843 (approved):

WanderingHogan (approved):

PR #618, opened 2021-12-13
feat: send images - images are sent to textile and display in conversations AP-78 / AP-249

Jekrimo:
**What this PR does** 📖
Allows users to send images within conversations.  Initializes textile buckets to store images.

AP-78 / AP-249

**Additional comments** 🎤
This is a start for our images, there will need to be tweaks to support file types other than images.


stavares843 (commented):
Captura de ecrã 2021-12-13, às 23 36 17

the download file icon seems pretty small

and not sure if the filename maybe would be better on the bottom instead on side, but that's a question for design team to take a look

stavares843 (changes requested):
https://user-images.githubusercontent.com/29093946/145906293-225308ff-40c8-4bcd-b493-65ee97f28f4c.mp4

when sending an image, the box with the send button doesn't go away after you click it to send it and if you keep clicking sends out the same image multiple times

so seems we need to
- dismiss the box right after user clicks on send 
- add some limit to avoid spamming requests

stavares843 (changes requested):
Captura de ecrã 2021-12-13, às 23 45 51

is missing a loading when the images are still loading

stavares843 (changes requested):
Captura de ecrã 2021-12-13, às 23 48 09

are we compacting the images after sending them on chat? I sent a 100MB image and the tab on the browser is very slow 

in case you need 100MB images NASA website has a bunch of those
https://visibleearth.nasa.gov/images/73751/july-blue-marble-next-generation-w-topography-and-bathymetry

components/tailored/core/fileupload/FileUpload.vue
```diff
@@ -28,19 +30,35 @@ export default Vue.extend({
...
   },
   data() {
     return {
       files: [] as Array<UploadDropItemType>,
       uploadStatus: false,
       count_error: false,
+      progress: 0,
+      ipfsHash: false,
+      selectedFile: false,
+      imageURL: String,
```
WanderingHogan, line 50: primitives should be lower case

components/tailored/core/fileupload/FileUpload.vue
```diff
@@ -125,6 +143,17 @@ export default Vue.extend({
...
       this.$data.count_error = false
     },
+    /**
+     * @method sendMessage
+     * @description Sends action to Upload the file to textile.
+     */
+    async sendMessage () {
+      this.$store.dispatch('textile/sendFileMessage', {
+        to: this.recipient.textilePubkey,
+        file: this.$data.files[0]
```
WanderingHogan, line 153: this is fine for this first pass but we should add a ticket to handle multiple file upload

components/ui/Embeds/File/File.vue
```diff
@@ -1,28 +1,38 @@
...
+  data() {
+    return {
+      fileUrl: String,
+      fileSize: '',
+    }
+  },
   computed: {
-    filesize() {
-      return filesize(this.$props.data.size)
+    getFileSize() {
```
WanderingHogan, line 24: where is this defined?

types/textile/manager.ts
```diff
@@ -1,5 +1,6 @@
 import { Client, Identity, Users } from '@textile/hub'
 import { SolanaWallet } from '../solana/solana'
+import * as t from "io-ts";
```
WanderingHogan, line 3: i don't think we are using this

components/tailored/core/fileupload/FileUpload.vue
stavares843: is this something expected to be burned by the linter?

components/ui/Embeds/File/File.vue
```diff
@@ -1,28 +1,38 @@
...
+    return this.bytesToSize(this.file.size)
     },
   },
+  methods: {
+    bytesToSize (bytes: number) {
+      const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB']
```
WanderingHogan, line 30: this is a neat way to do this

components/tailored/core/fileupload/FileUpload.vue
Jekrimo: 100%, I'll add a couple new tickets, this being one, but also one for handling other types of emails

store/textile/actions.ts
```diff
@@ -222,8 +223,53 @@ export default {
...
+    const path = `/${file.file.name}`
+    $TextileManager.bucketManager?.getBucket()
+    const result = await $TextileManager.bucketManager?.pushFile(
+      file.file,
+      path,
+      (progress: number) => {
+        progress
+      }
+    )
+    const imageURL = `https://hub.textile.io${result?.root}${path}`
```
WanderingHogan, line 246: can you add this to the config

components/ui/Embeds/File/File.vue
Jekrimo: I stole it from lega.... I mean, yeah I'm real smart

libraries/Textile/BucketManager.ts
```diff
@@ -0,0 +1,183 @@
...
+    this.textile = textile;
+    this.buckets = null;
+    this.bucketKey = null;
+    this.bucketName = 'v74files';
+    this.prefix = prefix;
+  }
+
+  private progressParse(uploaded: number, total: number) {
+    return uploaded / total * 100;
+  }
```
WanderingHogan, line 43: can you add the inline documentation here when you get a chance

WanderingHogan (approved):

components/tailored/core/fileupload/FileUpload.vue
WanderingHogan: I am not sure about the linter, an error should be shown in the ide though. Do you think anything that throws an error in the IDE should be in the linter?  I also think this will transpile fine but could potentially cause problems in the typescript, but is more of a best practices thing

string is a primitive, String is an object, so object assignment could get wonky in typescript, example:

Screen Shot 2021-12-15 at 4 26 52 PM

WanderingHogan:
@stavares843 going through your comments:

- `the download file icon seems pretty small` - this will be addressed in a ui ticket, AP-276
- `dismiss the box right after user clicks on send` - i think this is in? I see the reactions when you hover on the image
- add some limit to avoid spamming requests - button appears to be deactivated in recent pushes after you click the upload button
- `is missing a loading when the images are still loading`- a ticket Jeff will create when his internet comes back
- `are we compacting the images after sending them on chat...` we are not, we need to look more into the speed of uploading files to the encrypted textile buckets, see if speeds slow down for specific file sizes or if it's a distance thing or some other problem slowing uploads down

WanderingHogan:
related tickets are:
```
1. multi file upload (AP-274)
2. handling other file types (AP-275)
3. UI updates (AP-276)
4. Download button (tbd)
5. Downloading indicator (tbd)
6. Adding glyphs to image posts (tbd)
```

stavares843 (commented):
first time sending the file crashed

https://user-images.githubusercontent.com/29093946/146287629-914ec175-2ada-4023-995f-dea403f64e69.mp4

non reproducible

Chrome, macOS

stavares843:
pretty neat! 🎉 

image

stavares843 (changes requested):
if you try to send 4 images, only sends 1

is that expected? if is something to be added later?

stavares843 (changes requested):
Captura de ecrã 2021-12-16, às 00 55 49

it says upload blocked but has the send button activated and if you click, it sends the image

on dev doesn't have the send button on a blocked image

Captura de ecrã 2021-12-16, às 01 00 33

WanderingHogan:
@stavares843 multi file upload  will be addressed in (AP-274)

stavares843:
ohh thanks 😂

stavares843:
and thanks for clarifying the above points! 💥

stavares843:
added tickets here

[AP-285](https://satellite-im.atlassian.net/browse/AP-285)
[AP-286](https://satellite-im.atlassian.net/browse/AP-286)

stavares843 (approved):

PR #623, opened 2021-12-14
feat: [AP-65] - emoji chat render

KemoPaw:
**What this PR does** 📖
Have emojis not render as tiny as they were previously

https://satellite-im.atlassian.net/browse/AP-65

Fixes #

**Additional comments** 🎤

### DIFF
--- a/components/tailored/core/chatbar/Chatbar.vue
+++ b/components/tailored/core/chatbar/Chatbar.vue
@@ -245,6 +245,10 @@ export default Vue.extend({
       sel?.selectAllChildren(messageBox)
       sel?.collapseToEnd()
     },
+    makeOnlyEmojiLarger(text: any, originalText: any) {
+       
+        return text
+    },
     /**
      * @method sendMessage
      * @description Sends message by calling the sendMessage action with current data and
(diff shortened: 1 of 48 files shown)


stavares843 (commented):
@KemoPaw the emojis are awesome! 🎉 

Captura de ecrã 2021-12-14, às 00 09 34

stavares843 (commented):
screenshot with a couple variations @WanderingHogan 

Captura de ecrã 2021-12-14, às 00 10 55

stavares843 (approved):

components/tailored/core/chatbar/Chatbar.vue
```diff
@@ -245,13 +246,50 @@ export default Vue.extend({
...
+            let wrapAround = `<span class="mediumEmoji">${emojiFound}</span>`
+            newValue += wrapAround
+            i++
+          } else {
+            newValue += firstChar
+          }
+        }
+        return newValue
+      }
+      if (!newInput) {
```
WanderingHogan, line 280: don't need this if, the return statement in the if above will kick you out of the method before you get here

components/tailored/core/chatbar/Chatbar.vue
```diff
@@ -245,13 +246,50 @@ export default Vue.extend({
...
+            /[\p{Emoji}\u200d]+/gu,
+          )
+            ? firstChar + secondChar
+            : false
+
+          if (emojiFound) {
+            let wrapAround = `<span class="mediumEmoji">${emojiFound}</span>`
+            newValue += wrapAround
+            i++
+          } else {
```
WanderingHogan, line 274: never else, use a guard clause here

WanderingHogan (changes requested):
couple changes in comments

josephmcg (changes requested):
https://jsfiddle.net/qu43byLn/7/
I was talking with Matt about this and he thought we should check the input string in a different way. The for loop may cause performance issues down the line. See if you can draw inspiration from the jsfiddle

You can also add regex rules to the config file at:
Core-PWA/config.ts

KemoPaw:
https://github.com/Satellite-im/Core-PWA/pull/649

PR #670, opened 2021-12-16
feat(importAccount): user should be able to select presuggestion by using arrow keyboard

nickjjang:
AP-258


WanderingHogan (approved):

josephmcg:
I was able to select an option that's offscreen, beyond the 10th choice. typed something with many options like 'a' and keep hitting down

Jekrimo (approved):

PR #677, opened 2021-12-17
fix(friends): friends - account id should be cleared after friend request is sent

genie-magic:
**What this PR does** 📖
friends - account id should be cleared after friend request is sent

Fixes # AP-294


stavares843 (approved):
fixed ✅

WanderingHogan (approved):

PR #752, opened 2021-12-22
feat(cache): use service worker to cache ipfs responses

josephmcg:
**What this PR does** 📖
- service worker caches ipfs responses like glyphs and mute/deafen/video mp3
- using StaleWhileRevalidate since ipfs sends opaque responses. If we enable CORS on ipfs, we could look into cacheFirst

N/A

https://pwa.nuxtjs.org/workbox/#enabled
- documentation recommends testing by running prod locally. I couldn't get that to work, so I just switch enabled to true in nuxt.config
- you will need to check application tab of chrome dev tools to see details about service worker

**Additional comments** 🎤

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -153,7 +153,7 @@ export default defineNuxtConfig({
       runtimeCaching: [
         {
           urlPattern: 'https://satellite.mypinata.cloud/ipfs/*',
-          handler: 'cacheFirst',
+          handler: 'StaleWhileRevalidate',
           method: 'GET',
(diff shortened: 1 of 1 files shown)


stavares843 (approved):

iltumio (approved):

iltumio (changes requested):
needs to be rebased before merging

iltumio (approved):

iltumio:
what is the downside of keeping the `enabled` option set to true even running the app in localhost?

PR #798, opened 2021-12-24
fix(signup): fix sign up with just spaces on username

nickjjang:
AP-297


WanderingHogan (approved):

stavares843 (commented):
Captura de ecrã 2021-12-27, às 14 12 51

fixed

but if you add 5 spaces and then any char, returns error to the user

we can improve that on other ticket

stavares843 (approved):

PR #809, opened 2021-12-27
fix(Themes): add additional theme support to inputs

RetroPronghorn:
**What this PR does** 📖
This PR cleans up the layout of the audio page and adds a new component to add reusability in the settings pages. I have not added all of the components into this new layout we can address a new ticket for this. This ticket also covers various theme changes @stavares843 had requested.


RetroPronghorn:
@stavares843 there are some small things in search that are not yet themed (and profile) but leaving them for now as I think we'll need to refactor those components later anyways as we build out the app.

josephmcg (commented):
I feel like normalizing the use of !important is a dangerous approach. It quickly gets sloppy and can only be overridden by another !important

RetroPronghorn:
> I feel like normalizing the use of !important is a dangerous approach. It quickly gets sloppy and can only be overridden by another !important

Agreed but it's required for themes since youre overriding everything that exists already in the app. I think usage of the important tag outside of a theme though should never be allowed (except the rare case were overriding plug-ins or module themes). Sadly because of locally scoped less files for theming I haven't found a better solution without a major overhaul of how we do things and I don't want to do that pre alpha.

josephmcg:
> Agreed but it's required for themes since youre overriding everything that exists already in the app. I think usage of the important tag outside of a theme though should never be allowed (except the rare case were overriding plug-ins or module themes). Sadly because of locally scoped less files for theming I haven't found a better solution without a major overhaul of how we do things and I don't want to do that pre alpha.

I see, a necessary evil. Will review now!

josephmcg (approved):
Looks good, a couple comments/questions:

- Once mic or camera is blocked, can we prompt again or do they have to update in the browser?
- Flair only makes a difference for Moonless Night, yeah?
- I'd imagine Lime and Sunflower don't have enough contrast
- Is this a missed translation?
image

stavares843:
Captura de ecrã 2021-12-28, às 18 24 30

should these blue numbers be themed?

stavares843:
Captura de ecrã 2021-12-28, às 18 24 36

default color when going to profile > search language

stavares843:
Captura de ecrã 2021-12-28, às 18 26 24

should those be themed?

stavares843:
Captura de ecrã 2021-12-28, às 18 27 15

should the checkmarks on the filters and the add to cart icons be themed?

stavares843:
Captura de ecrã 2021-12-28, às 18 30 15

stavares843 (approved):
pretty neat! 🎉

added couple comments

WanderingHogan (approved):
I can't actually see where this is changing anything, but it doesn't seem to be breaking anything.

WanderingHogan:
oh i see it now, thanks, @josephmcg's comment

stavares843:
ticket added here with the missing components - [AP-384](https://satellite-im.atlassian.net/browse/AP-384)

PR #814, opened 2021-12-28
feat(indexeddb): Added dexie to support indexeddb storage AP-366

josephmcg:
**What this PR does** 📖
- uses dexie for better indexeddb api. I researched other options (idb,localforage) too, but dexie seems like the best
- on message load - if nothing is stored in indexeddb, fetch every message from textile and set in indexeddb
- on message load - if indexeddb has info, pull any new messages from textile based on last inbound message timestamp. combine indexeddb messages with new textile messages and set in store.
- on message send/edit - set in indexeddb. Edit replaces message in indexeddb similar to how textile would return an edited message

#AP-366

- indexeddb data is found in the chrome dev tools Application tab
- since textile stores message times in unix nanoseconds, the textile fetch will return the latest message, even if nothing is new. We may want to pull the lastInbound value directly from textile rather than set it ourselves using Date.now()

**Additional comments** 🎤


stavares843:
@josephmcg is this still a WIP or its ready? 🔨

Jekrimo (approved):

josephmcg:
@stavares843 WIP -  I'm thinking of adding some additional functionality to it first. Right now, it's just storing conversation info every time its fetched

stavares843:
@josephmcg just checking, is this ready to review/test? 🔨

josephmcg:
not yet, I'm taking a couple vacation days. I'm hoping to finish it up tomorrow

stavares843:
neat, thanks 🎉

store/textile/mutations.ts
```diff
@@ -106,6 +107,32 @@ const mutations = {
...
+    // indexeddb work below
+
+    // don't store editing message
+    if (message.editingAt) {
+      return
+    }
+
+    // replace old message with new edited version
+    if (message.editedAt) {
+      db.conversations.get(address).then((convo) => {
```
iltumio, line 120: I know that technically is not the mutation to be async, but it's side effect in this case. Anyway we should not put any async logic inside mutations in my opinion.
https://vuex.vuejs.org/guide/mutations.html#mutations-must-be-synchronous

Can we move this inside the action or eventually in a store plugin?
https://vuex.vuejs.org/guide/plugins.html#plugins

iltumio (changes requested):
the most important part is to move the indexedDb async logic from the mutation to an action or into a  store plugin. See comments

WanderingHogan (approved):
this is neat

josephmcg:
Manuel gave me some good feedback regarding the data structure. I'm going to finish refactoring within a day or two

iltumio (approved):

PR #840, opened 2021-12-29
fix(nsfw): don't allow nsfw images to be sent

Jekrimo:
**What this PR does** 📖
Blocks sending images deemed nsfw and shows an alert notifying user that file/s were not sent that were nsfw

AP-285

Alert should pop up on it's own and auto close after 5ish seconds, the user can also click the close on the alert to close before then.

### DIFF
--- a/components/ui/Chat/Image/Image.html
+++ b/components/ui/Chat/Image/Image.html
@@ -18,19 +18,6 @@
     v-on:click='() => $data.showfull = !$data.showfull' />
   <div class='file-info'>
     <div class='download-container'>
-      <image-icon class='icon' />
-      <TypographySubtitle class='name-text' :size='6' :text='`${image.name}`' />
-
-      <a class='dl-text link' href='#' :download='image.url'>
-        <download-icon size='1x' class='icon' />
-        <TypographyText class='name-text text link' :text="$t('controls.download')" />
-      </a>
-    </div>
-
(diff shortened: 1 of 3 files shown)


phillsatellite:
Tested and never had a alert pop up, the image would be blured when uploading and send button was disabled 

testnsfw

Jekrimo:
> Tested and never had a alert pop up, the image would be blured when uploading and send button was disabled
> 
> testnsfw

Fixed! I'm a big dummy and was always uploading multiple files to run the check and totally didn't account for a singular nsfw file lol

phillsatellite (approved):
Boom alert message is showing up now!

components/views/files/upload/Upload.html
```diff
@@ -10,6 +11,17 @@
       <TypographyError :text="$t('errors.chat.drop_file_count')" />
       <x-icon size="1x" class="has-text-danger" @click="cancelUpload" />
     </div>
+    <div v-if="alertNsfw" class="file-error">
+      <TypographyError :text="$t('errors.chat.contains_nsfw')" />
```
stavares843, line 15: Captura de ecrã 2021-12-30, às 16 34 07

we need to add the string into the locales

stavares843 (changes requested):

stavares843 (approved):

WanderingHogan (approved):

PR #973, opened 2022-01-11
fix(reaction): fixes emoji reacts for reply messages of inner messages

josephmcg:
**What this PR does** 📖
Pretty specific use case, but I noticed something was still buggy after mine and David's fixes earlier.

Emoji reacts to a reply of a message which _is not_ the first message of a group were still not working.

With this, two reactions for 'reply test' will now appear

image


RetroPronghorn (approved):

stavares843 (approved):
fixed ✅

PR #988, opened 2022-01-12
fix(toolbar): show wallet and market buttons in mobile view AP-218

davidchan1219:
**What this PR does** 📖

Made to show marketplace and wallet button in mobile

AP-218


WanderingHogan:
So - you did a good job, i just didn't realize in my head how much of a disaster it would look like. I think we should close this for now and come back when things are re-designed. What do you think?

davidchan1219:
> So - you did a good job, i just didn't realize in my head how much of a disaster it would look like. I think we should close this for now and come back when things are re-designed. What do you think?

For now, I think that is not so big problem. let's remain that.

stavares843:
closing this one based on the comments above

PR #991, opened 2022-01-12
fix(ui): sidebar stretching outside view

vimercati-samir:


vimercati-samir:
branch name do not correspond with right Jira task (AP-383)

PR #1014, opened 2022-01-14
fix(ui): missing components on themes

vimercati-samir:
Screenshot 2022-01-14 at 11 56 09
Screenshot 2022-01-14 at 11 56 25
Screenshot 2022-01-14 at 11 56 52
Screenshot 2022-01-14 at 11 57 11
Screenshot 2022-01-14 at 11 57 28
Screenshot 2022-01-14 at 11 57 58

### DIFF
--- a/assets/styles/themes/moonless_night.less
+++ b/assets/styles/themes/moonless_night.less
@@ -149,7 +149,7 @@
     box-shadow: @flair-glow !important;
   }
   .multiselect__tags {
-    background: var(--flair-color) !important;
+    background: var(--flair-color);
     border: var(--flair-color);
 
     .multiselect__input {
@@ -369,7 +369,7 @@
 
   .tag {
     .tag-number {
-      background: var(--flair-color) !important;
+      background: var(--flair-color);
     }
   }
 
--- a/assets/styles/themes/moonless_night.less
+++ b/assets/styles/themes/moonless_night.less
@@ -151,6 +151,11 @@
   .multiselect__tags {
     background: var(--flair-color) !important;
     border: var(--flair-color);
+
+    .multiselect__input {
+      background: var(--flair-color);
+      box-shadow: @flair-glow;
+    }
   }
 
   .chip-item {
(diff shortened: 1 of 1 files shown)


assets/styles/themes/moonless_night.less
```diff
@@ -361,4 +366,22 @@
   .new-message-alert {
     background: var(--flair-color) !important;
```
iltumio, line 367: is there a way to avoid `!important`?

assets/styles/themes/moonless_night.less
```diff
@@ -361,4 +366,22 @@
...
   }
+
+  .tag {
+    .tag-number {
+      background: var(--flair-color) !important;
```
iltumio, line 372: same as above

assets/styles/themes/moonless_night.less
```diff
@@ -151,6 +151,11 @@
   .multiselect__tags {
     background: var(--flair-color) !important;
```
iltumio, line 152: same as below

iltumio (dismissed):

assets/styles/themes/moonless_night.less
vimercati-samir: resolved in new commit

assets/styles/themes/moonless_night.less
vimercati-samir: resolved in new commit

assets/styles/themes/moonless_night.less
vimercati-samir: new-message-alert style has not been affected by AP-384, nonetheless, should I investigate?

WanderingHogan:
I believe this is good to go in, and the !important from before is something we should come back to in a larger ticket where we review all of the style rules/overrides

WanderingHogan (approved):

WanderingHogan:
Sorry - not ready to go in, but ready for QA

phillsatellite (approved):

PR #1037, opened 2022-01-19
feat(filesystem): add new fuzzySearch function AP-479

davidchan1219:
**What this PR does** 📖

Added new search function

AP-479

### DIFF
--- a/libraries/Files/FileSystem.ts
+++ b/libraries/Files/FileSystem.ts
@@ -3,6 +3,7 @@ import { DIRECTORY_TYPE } from './types/directory'
 import { Fil } from './Fil'
 import { Item } from './abstracts/Item.abstract'
 import { FileSystemExport, FILESYSTEM_TYPE } from './types/filesystem'
+import { matchSorter } from 'match-sorter'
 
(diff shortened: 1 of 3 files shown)


libraries/Files/FileSystem.ts
```diff
@@ -301,6 +302,21 @@ export class FileSystem {
...
+   */
+  public fuzzySearch(partial: string): Item[] {
+    const itemList: Item[] = this.findAllItems(this.checkString, this.root)
+    return matchSorter(itemList, partial, { keys: ['_name', '_type'] })
+  }
+
+  checkString(item: Item): boolean {
+    if (item._name.includes('')) return true
+    return false
+  }
```
molimauro, line 318: `includes()` returns already a boolean, maybe can we rewrite this to:
```
return item._name.includes('')
``` 
What do u think?

davidchan1219: Okay, That seems good. I will change the code.

molimauro (approved):

stavares843 (approved):

PR #1062, opened 2022-01-21
fix(call): fix ui issue when user hang up / deny call

genie-magic:
**What this PR does** 📖

Right now if a user disconnects/deny call, the other user doesn’t get any feedback via sound or ui update. The call should end/ui should exit the call instead of leaving it open

AP-524

### DIFF
--- a/components/ui/Global/Global.vue
+++ b/components/ui/Global/Global.vue
@@ -85,8 +85,8 @@ export default Vue.extend({
 
       peer?.call.deny()
 
-      // peer?.send('SIGNAL', { type: 'CALL_DENIED' })
-      // this.$store.dispatch('webrtc/denyCall')
+      peer?.send('SIGNAL', { type: 'CALL_DENIED' })
+      this.$store.dispatch('webrtc/denyCall')
     },
   },
 })
--- a/store/webrtc/actions.ts
+++ b/store/webrtc/actions.ts
@@ -74,12 +74,12 @@ export default {
       )
     })
 
-    // peer?.communicationBus.on('RAW_DATA', (message) => {
-    //   if (message.data.type === 'CALL_DENIED') {
-    //     peer?.call.hangUp()
-    //     dispatch('hangUp')
-    //   }
-    // })
+    peer?.communicationBus.on('RAW_DATA', (message) => {
+      if (message.data.type === 'CALL_DENIED') {
+        peer?.call.hangUp()
+        dispatch('hangUp')
+      }
+    })
 
     peer?.call.on('INCOMING_CALL', (data) => {
       // if incoming call is activer call return before toggling incoming call


molimauro:
Sorry to bother, do you know why Jason commented those lines in the first place?

phillsatellite:
https://user-images.githubusercontent.com/93608357/150801082-bc8b2b24-4350-4103-aa15-42b3ad93cd69.mov

Only thing I noticed when testing, if a user A calls User B and User B declines that call, if user A tries to call again call will not go through. I can start a new ticket for it if you didn't wanna throw it in this pr just lemme know 😄

genie-magic:
> endingcall.mov 
> Only thing I noticed when testing, if a user A calls User B and User B declines that call, if user A tries to call again call will not go through. I can start a new ticket for it if you didn't wanna throw it in this pr just lemme know 😄

We can do that with new ticket...

genie-magic:
> Sorry to bother, do you know why Jason commented those lines in the first place?

I don't know that.. Do you know?

molimauro:
> > Sorry to bother, do you know why Jason commented those lines in the first place?
> 
> I don't know that.. Do you know?

unfortunately no. Let's ask him @jpanay :)

jpanay:
> > > Sorry to bother, do you know why Jason commented those lines in the first place?
> > 
> > 
> > I don't know that.. Do you know?
> 
> unfortunately no. Let's ask him @jpanay :)

I wrote those lines but someone else must've commented them out, so not completely sure why either. What it looks like to me is that deny call was switched from triggering that raw data listener to triggering the HANG_UP listener which is why raw data was commented out. @iltumio Is raw data the way to do it or should we work around the way it's being done now?

genie-magic:
> > > > Sorry to bother, do you know why Jason commented those lines in the first place?
> > > 
> > > 
> > > I don't know that.. Do you know?
> > 
> > 
> > unfortunately no. Let's ask him @jpanay :)
> 
> I wrote those lines but someone else must've commented them out, so not completely sure why either. What it looks like to me is that deny call was switched from triggering that raw data listener to triggering the HANG_UP listener which is why raw data was commented out. @iltumio Is raw data the way to do it or should we work around the way it's being done now?

I think current code have no problem... but plz test on your side too

jpanay (approved):

iltumio:
I checked the code. Whenever you call the method `peer?.call?.deny()` a "REFUSE" message is sent via the communicationBus. Whenever a "REFUSE" message is received from the other peer, the `onClose` method (that involves the hangUp method too) is called. Looking at the code, I'm quite sure that the issue can happen on the local peer only. Am I right? In this case we should have a look at the `deny()` method inside `Call.ts` instead of using the "RAW_DATA" message listener

Schermata 2022-01-28 alle 15 50 17

Schermata 2022-01-28 alle 15 50 43

phillsatellite:
removing QA Tested due to @iltumio changes requested

genie-magic:
@molimauro  you are right. Anyway I will close this PR and create new one with rebased.

PR #1065, opened 2022-01-21
feat(timezone): add timezone picker to profile setting

genie-magic:
**What this PR does** 📖
add timezone picker to profile setting

AP-518


stavares843 (changes requested):
I think it's somewhat strange for UI/UX regarding we can click but doesn't work, the user is obligated to use the enter in order to save the selection

https://user-images.githubusercontent.com/29093946/150611882-4508f99b-5451-463b-a65b-22298f7a2d15.mov

stavares843 (changes requested):
Captura de ecrã 2022-01-21, às 23 10 37

also, this all says location, but what the user is really selecting is the timezone, and timezone and location are different things, I can be anywhere in Portugal but selecting Lisbon as the timezone

stavares843 (changes requested):
Captura de ecrã 2022-01-21, às 23 11 09

after selecting a timezone, the text inside is not vertically aligned

stavares843 (changes requested):
https://user-images.githubusercontent.com/29093946/150611882-4508f99b-5451-463b-a65b-22298f7a2d15.mov


when typing, the font color is dark but then after selecting a timezone is white

mock/profile.ts
```diff
@@ -46,3 +45,354 @@ export const recommendLocations = [
...
+  'Indian/Kerguelen',
+  'Asia/Bangkok',
+  'Asia/Dushanbe',
+  'Pacific/Fakaofo',
+  'Asia/Dili',
+  'Asia/Ashgabat',
+  'Africa/Tunis',
+  'Pacific/Tongatapu',
+  'Europe/Istanbul',
+  'America/Port_of_Spain',
```
stavares843, line 353: can we edit the ones with an underscore in order to avoid showing Port_of_Spain on the user side?

mock/profile.ts
```diff
@@ -46,3 +45,354 @@ export const recommendLocations = [
...
+
+export const timezones = [
+  'Europe/Andorra',
+  'Asia/Dubai',
+  'Asia/Kabul',
+  'Europe/Tirane',
+  'Asia/Yerevan',
+  'Antarctica/Casey',
+  'Antarctica/Davis',
+  'Antarctica/DumontDUrville',
```
stavares843, line 57: seems the correct name is Dumont-d'Urville

mock/profile.ts
```diff
@@ -46,3 +45,354 @@ export const recommendLocations = [
...
+  'Atlantic/Faroe',
+  'Europe/Paris',
+  'Europe/London',
+  'Asia/Tbilisi',
+  'America/Cayenne',
+  'Africa/Accra',
+  'Europe/Gibraltar',
+  'America/Godthab',
+  'America/Danmarkshavn',
+  'America/Scoresbysund',
```
stavares843, line 192: seems the correct name is Scoresby Sund

genie-magic:
> I think it's somewhat strange for UI/UX regarding we can click but doesn't work, the user is obligated to use the enter in order to save the selection
> 
>  Gravacao.do.ecra.2022-01-21.as.23.11.45.mov 
> also asked the design team if they have any ideas
> 
> design team replied it shouldn't be like that, typically both the user clicking the option or hitting 'enter' key will work just fine.

@stavares843  Problem is that multiselect package we are using for select does not support click option... cc: @WanderingHogan

WanderingHogan:
@genie-magic can you switch these to the normal dropdowns, the user has no reason to select more than one anyways, and also make the other updates Sara found

WanderingHogan:
can you store the offset, too, and provide a way to access it

WanderingHogan (changes requested):
please make sara's changes, switch to a dropdown, and create that way to get the offset

stavares843:
cc @genie-magic

genie-magic:
Sorry for late response, @WanderingHogan , @stavares843 .
I will do it after finishing current tasks.

genie-magic:
@stavares843  , @WanderingHogan  I will create new PR with rebased code.

PR #1075, opened 2022-01-24
fix(chat): enforce the 256 chars when sending

genie-magic:
**What this PR does** 📖
enforce the 256 chars when sending

AP-512

### DIFF
--- a/components/views/chat/chatbar/Chatbar.vue
+++ b/components/views/chat/chatbar/Chatbar.vue
@@ -35,7 +35,6 @@ export default Vue.extend({
     return {
       text: '',
       showEmojiPicker: false,
-      maxChars: Config.chat.messageMaxChars,
       recipientTyping: false,
       showFilePreview: false,
       nsfwUploadError: false,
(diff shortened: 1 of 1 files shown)


phillsatellite:
https://user-images.githubusercontent.com/93608357/150798239-5e504ea6-6e1d-42c5-ada8-034a3b2b5862.mov

https://user-images.githubusercontent.com/93608357/150798245-d62f0ff2-8e20-4992-a804-638c2d908981.mov

only two things I noticed while testing, when user has reached the 256 limit, if they try to keep adding more whatever they add will appear in the beginning of the message. Second issue was, if user has a message over 256 chars and they start to delete some of their text, it'll freeze up at 257 characters and user will need to click the text again so they can continue to delete. 👍

genie-magic:
> adding.more.when.limit.is.reached.mov 
>  delete.when.reached.257.mov 
> only two things I noticed while testing, when user has reached the 256 limit, if they try to keep adding more whatever they add will appear in the beginning of the message. Second issue was, if user has a message over 256 chars and they start to delete some of their text, it'll freeze up at 257 characters and user will need to click the text again so they can continue to delete. 👍

@phillsatellite  my changes is only for sending maximum 256 characters when sending message... The issues you mentioned there , you can create individual ticket..  My code is only for enforce 256 chars when sending.

Jekrimo (approved):

phillsatellite (approved):

phillsatellite:
> > adding.more.when.limit.is.reached.mov
> > delete.when.reached.257.mov
> > only two things I noticed while testing, when user has reached the 256 limit, if they try to keep adding more whatever they add will appear in the beginning of the message. Second issue was, if user has a message over 256 chars and they start to delete some of their text, it'll freeze up at 257 characters and user will need to click the text again so they can continue to delete. 👍
> 
> @phillsatellite my changes is only for sending maximum 256 characters when sending message... The issues you mentioned there , you can create individual ticket.. My code is only for enforce 256 chars when sending. Those issues are different content with ticket content.

Will do 👍 thank you!

WanderingHogan (approved):

PR #1082, opened 2022-01-24
add(jest): 100% coverage for SoundManager unit test

drepram:
**What this PR does** 📖

Add unit tests (100% coverage) for the SoundManager unit test.

AP-139

The `this.sounds[sound].play()` returns two weird `console.error`:

image

I did some googling and came across [this](https://stackoverflow.com/questions/51829319/how-to-mock-video-pause-function-using-jest/51831366), to which I tried to add to the code but it didn't work so I just commented it out for further debugging!

image

I think it's a particular problem to` .play()` because other units in the test suite are written identically and yet only `.play()` have those two console errors:

image

**Additional comments** 🎤

### DIFF
--- a/libraries/SoundManager/SoundManager.test.ts
+++ b/libraries/SoundManager/SoundManager.test.ts
@@ -1,0 +1,90 @@
+import { Howl } from 'howler'
+
+import { Sounds } from './SoundManager'
+import SoundManager from './SoundManager'
+
+describe('init', () => {
+    it('should pass',() => {
+        expect(Sounds).toMatchSnapshot(`
+    Object {
+    "CALL": "call",
+    "CONNECTED": "connected",
+    "DEAFEN": "deafen",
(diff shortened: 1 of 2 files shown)


josephmcg:
I don't think I know enough about Jest to give meaningful feedback here

WanderingHogan:
Jest uses jsdom, which doesn't have HTMLMediaElement, so you will have to fake it:
https://bharat-tiwari.gitbook.io/our-tech-journal/react/jest-setup-file

We had to do something similar for canvas. In this case you should be able to get away with the first little 4 line snippet from that link. For canvas we loaded something to mock it in dev dependencies, but I don't see it anymore

WanderingHogan (changes requested):
check out the comment to mock htmlaudio

WanderingHogan (approved):

stavares843 (approved):

PR #1141, opened 2022-01-25
add(live): add active and close chat for live component items AP-538

davidchan1219:
**What this PR does** 📖

added chat active and close function in live component

AP-538


components/views/navigation/sidebar/live/Live.html
```diff
@@ -1,12 +1,15 @@
 <div class="live">
-  <div class="name-and-icon">
+  <div class="name-and-icon" v-on:click="() => activeChat()">
```
josephmcg, line 2: @click, also don't think you need the arrow function

components/views/navigation/sidebar/live/Live.html
```diff
@@ -1,12 +1,15 @@
...
+    <TypographySubtitle :size="6" :text="ui.senderInfo.name || ''" />
   </div>
-  <InteractablesButton 
+  <InteractablesButton
     type="danger"
     size="small"
-    :text="$t('controls.end_stream')">
+    :text="$t('controls.end_stream')"
+    :action="() => hangUp()"
+    v-if="this.$store.state.friends.all.find((friend) => friend.address === this.$store.state.webrtc.activeCall)"
```
josephmcg, line 11: same extra arrow function here.
the v-if maybe be better off in a computed. You shouldn't include `this` if you keep it here

components/views/navigation/sidebar/live/Live.vue
```diff
@@ -10,6 +11,44 @@ export default Vue.extend({
...
+      const activeFriend = this.$Hounddog.getActiveFriend(
+        this.$store.state.friends,
+      )
+      if (!activeFriend) return
+      const identifier = activeFriend.address
+
+      if (!this.webrtc.connectedPeer) {
+        await this.$store.dispatch('webrtc/createPeerConnection', identifier)
+        if (!this.webrtc.connectedPeer) return
+      }
```
josephmcg, line 38: confused by the guard clause on 28 if the same condition is being checked on 35 to dispatch the connection, could you explain?

store/ui/mutations.ts
```diff
@@ -322,4 +322,7 @@ export default {
   updateFlair(state: UIState, flair: Flair) {
     state.theme.flair = flair
   },
+  setSenderInfo(state: UIState, senderInfo: Object) {
+    state.senderInfo = senderInfo
+  },
```
josephmcg, line 327: declare this as User, let's be kind to our future selves

store/ui/types.ts
```diff
@@ -167,4 +167,5 @@ export interface UIState {
     base: Theme
     flair: Flair
   }
+  senderInfo: Object
```
josephmcg, line 170: same ^

josephmcg (changes requested):

components/views/navigation/sidebar/live/Live.vue
davidchan1219: This function was made by other developer a few months ago. I think line 25 is not necessary.

store/ui/mutations.ts
davidchan1219: I changed function name and object variable name. like setSelectedUserInfo, selectedUserInfo

components/views/navigation/sidebar/live/Live.html
davidchan1219: changed

components/views/navigation/sidebar/live/Live.html
davidchan1219: changed

store/ui/types.ts
davidchan1219: changed the variable name

store/ui/mutations.ts
josephmcg: Hey David, thanks for the quick fixes. I wasn't concerned with the variable name, more so the declaration as type Object. Since we know it will be a User, we should declare it as such. Sorry if my comment wasn't clear.

davidchan1219: @josephmcg I just fixed the type

josephmcg (approved):

phillsatellite (approved):

WanderingHogan (approved):

WanderingHogan (changes requested):
I think the UI is a little off

WanderingHogan:
Screen Shot 2022-01-26 at 11 01 10 PM
I think that green icon should not be visible when not in a call

davidchan1219:
> I think the UI is a little off

Fixed. Please check again

WanderingHogan:
image
If you look close, the live component still shows this horizontal rule. It makes it the container too tall.

If you can hide the entire live component it looks way more balanced
image


When you make this change, if you have to fix conflicts, please rebase and not merge

WanderingHogan (changes requested):
see comments

PR #1142, opened 2022-01-25
feat(profilePic): block nsfw profile picture uploads AP-503

josephmcg:
**What this PR does** 📖
- block nsfw profile pictures on account registration
- block pictures that are too large to be NSFW scanned. I think hogan said we may ultimately break the nsfw check off to a separate service to allow for larger files to be scanned. Although, profile pic ultimately gets cropped and scaled down quite a bit anyway, so allowing larger files here doesn't seem mission critical.
- fix minor linter suggestion on config file

AP-503

**Additional comments** 🎤

### DIFF
--- a/components/views/user/create/Create.vue
+++ b/components/views/user/create/Create.vue
@@ -38,12 +38,29 @@ export default Vue.extend({
      * @param e
      * @example
      */
-    selectImage(e: Event) {
+    async selectImage(e: Event) {
       const target = e.target as HTMLInputElement
       if (target.value === null) return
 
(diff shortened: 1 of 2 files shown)


phillsatellite (approved):

WanderingHogan (approved):

stavares843:
Captura de ecrã 2022-01-27, às 20 51 15

maybe we could just say file or image, instead of file/s because thats only 1 image there

besides that, pretty neat 💥

stavares843 (approved):

PR #1143, opened 2022-01-25
fix(user): fix timestamp for user list AP-118

davidchan1219:
**What this PR does** 📖

added correct timestamp for last message

AP-118


components/views/user/User.vue
```diff
@@ -23,7 +23,17 @@ export default Vue.extend({
...
+    src(): string {
+      const hash = this.user?.profilePicture
+      return hash ? `${this.$Config.textile.browser}/ipfs/${hash}` : ''
+    },
+    getLastBoundary(): String {
+      const curUserMInfo = this.textile.conversations[this.user.address]
+      return curUserMInfo && curUserMInfo.lastInbound
+        ? this.$dayjs(curUserMInfo.lastInbound).from()
+        : ''
+    },
```
josephmcg, line 36: Is `curUserMInfo && curUserMInfo.lastInbound` necessary? why not just `curUserMInfo.lastInbound`

I believe this should be fromNow(). from() expects arguments
https://day.js.org/docs/en/display/from-now

I'll fix the dayjs type errors soon so this will be more apparent

josephmcg (changes requested):

components/views/user/User.vue
```diff
@@ -23,7 +23,17 @@ export default Vue.extend({
...
   },
   props: {
     user: {
```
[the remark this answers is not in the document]
davidchan1219: moved computed after data

components/views/user/User.vue
davidchan1219: changed from() to fromNow()

josephmcg (approved):
thanks David

phillsatellite (approved):

WanderingHogan (approved):

stavares843:
@davidchan1219 only missing conflict 🔨

davidchan1219:
> @davidchan1219 only missing conflict 🔨

@stavares843 fixed the conflict.

stavares843:
thanks! 🔨

stavares843:
Captura de ecrã 2022-01-27, às 20 43 42

my account doesn't have a timestamp

those messages are older than 24h ago

davidchan1219:
> Captura de ecrã 2022-01-27, às 20 43 42
> 
> my account doesn't have a timestamp
> 
> those messages are older than 24h ago

Are there messages from the friends of the sidebar?
If there is no any message from the friend, can't display the timestamp.

stavares843:
checking 🔨

stavares843:
the previous screenshot was my own messages, which I thought would appear when it was the messages from the current user as well

couple questions

Captura de ecrã 2022-01-27, às 22 11 27

should that say ''a few moments ago'' when the other user sent a message? the ticket says

''Once #AP-125 goes in, you will want to make the time in components/tailored/core/user/User.vue/html use the same timestamp transformation from the message box, but using the time data from the textile state object, lastInboundkey for the appropriate user'' - so seems it would be somewhat on the format of 12:00 or similar right? cc @WanderingHogan let me know if I misunderstood 

Captura de ecrã 2022-01-27, às 22 11 35


also, when the ''a few moments ago'' updated on the chat itself saying ''a minute ago'', but not on the sidebar so this is lacking a real-time update here

Captura de ecrã 2022-01-27, às 22 13 57


after switching chats, then updated properly, but I thought this was expected to say the exact timestamp such as 11:28 instead of 3 minutes ago

davidchan1219:
@stavares843 I fixed the timestamp issue you mentioned above. I made to load the message's last update time.

components/views/user/User.vue
```diff
@@ -79,6 +80,23 @@ export default Vue.extend({
       })
       this.$store.commit('ui/setUserProfile', this.user)
     },
+    getLastBoundary() {
+      const curUserMInfo =
```
stavares843, line 84: maybe currentUserInfo would be easier to understand compared to curUserMInfo

davidchan1219: @stavares843 renamed that

stavares843:
Captura de ecrã 2022-01-31, às 00 04 25

neat 💥

davidchan1219:
@stavares843 fixed the videoscreenshot's issue

WanderingHogan (changes requested):
with lastInbound added to indexeddb, is it possible to use that instead of what's going on in the getLastBoundary method?

PR #1156, opened 2022-01-25
feat(textile): add lastInbound to indexeddb records AP-598

davidchan1219:
**What this PR does** 📖

added lastInbound to indexeddb records

AP-598

### DIFF
--- a/store/textile/actions.ts
+++ b/store/textile/actions.ts
@@ -68,6 +68,8 @@ export default {
       return convo?.conversation ?? []
     })
 
+    let rLastInbound = 0
+
     //  if nothing stored in indexeddb, fetch entire conversation
     if (!dbMessages.length) {
       conversation = await $MailboxManager.getConversation({
@@ -84,6 +86,8 @@ export default {
         lastInbound,
       })
 
+      rLastInbound = lastInbound
+
       // use textileMessages as primary source. this way, edited messages will use the newest version
       const ids = new Set(textileMessages.map((d) => d.id))
       conversation = [
@@ -96,9 +100,7 @@ export default {
     const dbData: DexieMessage = {
       conversation,
       key: address,
-      lastInbound: conversation.length
-        ? conversation[conversation.length - 1].at
-        : 0,
+      lastInbound: rLastInbound,
     }
     db.conversations.put(dbData)
 
@@ -489,7 +491,7 @@ export default {
    * glyph to be sent, and pack name
    */
   async storeMessage(
-    {}: ActionsArguments<TextileState>,
+    { rootState }: ActionsArguments<TextileState>,
     {
       address,
       message,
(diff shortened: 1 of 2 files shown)


store/textile/actions.ts
```diff
@@ -512,6 +515,9 @@ export default {
     db.conversations
       .where('key')
       .equals(address)
-      .modify((convo) => convo.conversation.push(message))
+      .modify((convo) => {
+        convo.conversation.push(message)
+        convo.lastInbound = message.at
+      })
```
josephmcg, line 521: This method is for outbound messages. you should not update lastInbound based on an outbound message

store/textile/actions.ts
```diff
@@ -96,6 +96,9 @@ export default {
     const dbData: DexieMessage = {
       conversation,
       key: address,
+      lastInbound: conversation.length
+        ? conversation[conversation.length - 1].at
+        : 0,
```
josephmcg, line 101: This has the potential to set lastInbound based on an outboundMessage. I would suggest the lastInbound in textile store. you can move the scope up one level outside of the else.

image

josephmcg (changes requested):

phillsatellite (approved):

store/textile/actions.ts
davidchan1219: updated

store/textile/actions.ts
davidchan1219: removed that part

store/textile/actions.ts
josephmcg: David, please read this method documentation, understand what it is doing, and where it is called. It is only called when sending messages, you should not be adjusting lastInbound at all here.

store/textile/actions.ts
josephmcg: Rather than juggling two variables, you should just declare that const lastInbound outside of the else

store/textile/actions.ts
davidchan1219: StoreMessage is called when the user sends and receives the messages.
And I made to register incoming message's lastbound.

store/textile/actions.ts
davidchan1219: I updated the code with your suggestion.

store/textile/actions.ts
josephmcg: I see, thank you for helping me understand. It is being called for incoming messages inside of the subscribeToMailbox function.

davidchan1219: fixed

store/textile/actions.ts
```diff
@@ -68,6 +68,10 @@ export default {
       return convo?.conversation ?? []
     })
 
+    const lastInbound = rootState.textile.conversations.length
+      ? rootState.textile.conversations[address].lastInbound
+      : 0
+
```
josephmcg, line 74: this is always evaluating false because conversations is an object
If there are no inbound messages, it will be 0 anyway. I believe you can take advantage of ts optional chaining
`rootState.textile.conversations[address]?.lastInbound ?? 0`

image

josephmcg (changes requested):
1 more bug, then I think it's good to go

store/textile/actions.ts
davidchan1219: fixed

Jekrimo (approved):

josephmcg (approved):

WanderingHogan (approved):

PR #1184, opened 2022-01-27
feat(sidebar): sort the friend list by last updated time AP-119

davidchan1219:
**What this PR does** 📖

Sorted the friend list by last updated time

AP-119


Jekrimo (approved):

phillsatellite:
Tested and the only little issue I could find was if User A friend request User B, once they're accepted if User A starts to message User B they're messages won't be moved up to the top until User A has clicked into the chat. If that's how it should be just lemme know! 

https://user-images.githubusercontent.com/93608357/151453857-ec844089-0d3a-4163-adac-335685b7ff19.mov

stavares843:
what phil mentioned should be addressed 🔨

davidchan1219:
@phillsatellite I fixed the issue you found.

components/views/navigation/sidebar/Sidebar.vue
```diff
@@ -26,6 +32,11 @@ export default Vue.extend({
     MessageSquareIcon,
     MenuIcon,
   },
+  data() {
+    return {
+      userList: [],
+    }
+  },
```
molimauro, line 39: are you using this?

components/views/navigation/sidebar/Sidebar.vue
```diff
@@ -76,6 +87,20 @@ export default Vue.extend({
...
+      this.$store.commit('friends/sortFriends', textileObj)
+    },
+  },
+  watch: {
+    textile: {
+      handler(newValue) {
+        this.sortUserList(newValue)
+      },
+      deep: true,
+    },
```
molimauro, line 100: we can add `immediate: true` and remove the `mounted()` function

components/views/navigation/sidebar/Sidebar.vue
```diff
@@ -76,6 +87,20 @@ export default Vue.extend({
         this.$router.push({ path: '/friends/list' })
       }
     },
+    sortUserList(textileObj: any) {
+      this.$store.commit('friends/sortFriends', textileObj)
+    },
+  },
+  watch: {
+    textile: {
```
molimauro, line 95: here can we watch only `'textile.conversations'` since we're using only that property in the `sortUserList`?

components/views/navigation/sidebar/Sidebar.vue
davidchan1219: @molimauro Thanks for your review. I've just added that

components/views/navigation/sidebar/Sidebar.vue
davidchan1219: @molimauro No, I'm not using that. removed.

phillsatellite:
Retested, seems like user is unable to accept a friend request. Tried accepting same request on Dev and worked fine so assuming it might be an issue with this branch

https://user-images.githubusercontent.com/93608357/151815948-cbea7245-083e-4dfd-a32a-1623c000f4eb.mov

.

molimauro:
@phillsatellite i also noticed this problem on another branch, so not sure it is related to this branch. I think we should create a bug ticket and assign it to the backend team to understand why this is happening

phillsatellite:
> @phillsatellite i also noticed this problem on another branch, so not sure it is related to this branch. I think we should create a bug ticket and assign it to the backend team to understand why this is happening

Awesome thank ya!

molimauro (approved):

phillsatellite (approved):

PR #1194, opened 2022-01-28
fix(heic): Make .heic a sendable file type

genie-magic:
**What this PR does** 📖
Make .heic a sendable file type so that can preview on html5

AP-320

### DIFF
--- a/components/views/files/upload/Upload.vue
+++ b/components/views/files/upload/Upload.vue
@@ -6,8 +6,10 @@ import Vue, { PropType } from 'vue'
 import { mapState } from 'vuex'
 import { Config } from '~/config'
 import { PropCommonEnum } from '~/libraries/Enums/enums'
+import { isHeic } from '~/utilities/Heic'
 import { UploadDropItemType, FileType } from '~/types/files/file'
 import { Friend } from '~/types/ui/friends'
+const converter = require('heic-convert')
 
 declare module 'vue/types/vue' {
   interface Vue {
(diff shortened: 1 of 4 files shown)


components/views/files/upload/Upload.vue
```diff
@@ -72,13 +74,33 @@ export default Vue.extend({
...
         this.$parent.$data.showFilePreview = files.length > 0
         if (files.length > 8) {
           this.$data.count_error = true
           return
         }
         this.$data.count_error = false
+        for (let i = 0; i < files.length; i++) {
+          /* checking .heic file needs file array buffer because sometimes its file type return empty string */
+          const buffer = new Uint8Array(await files[i].arrayBuffer())
+          const isHeicType = await isHeic(buffer)
```
molimauro, line 87: why are u awaiting here?

utilities/Heic.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export const isHeic = (uint8Buffer: any) => {
+  if (!uint8Buffer || uint8Buffer.length < 24) return false
+  return (
+    uint8Buffer[20] === 0x68 &&
+    uint8Buffer[21] === 0x65 &&
+    uint8Buffer[22] === 0x69 &&
+    uint8Buffer[23] === 0x63
+  )
+}
```
molimauro, line 16: is this check taken from some library?

components/views/files/upload/Upload.vue
genie-magic: arrayBuffer is async function... and about isHeic.. I removed await and commited again. @molimauro

utilities/Heic.ts
genie-magic: I referenced some packages and made that. No need to install npm package for that kind of small function. @molimauro

components/views/files/upload/Upload.vue
molimauro: yeah sorry i meant only about isHeic. Perfect however thanks!

utilities/Heic.ts
molimauro: Yep, totally agree. Thanks!

molimauro (approved):

Jekrimo (approved):

phillsatellite (approved):

PR #1233, opened 2022-01-31
feat(user): hide the non-message user in the sidebar AP-116

davidchan1219:
**What this PR does** 📖

Hidden the users has no messages in the sidebar

AP-116

### DIFF
--- a/components/views/user/User.vue
+++ b/components/views/user/User.vue
@@ -81,14 +81,11 @@ export default Vue.extend({
     },
     existMessage(textileObj: Conversation) {
       const currentUserInfo = textileObj[this.user.address]
-      if (
+      this.$data.existConversation =
         currentUserInfo?.lastUpdate <= 0 &&
         this.user.account.from === this.user.address
-      ) {
-        this.$data.existConversation = false
-      } else {
-        this.$data.existConversation = true
-      }
+          ? false
+          : true
     },
   },
   watch: {
--- a/components/views/user/User.vue
+++ b/components/views/user/User.vue
@@ -79,10 +79,10 @@ export default Vue.extend({
       this.$store.commit('ui/setUserProfile', this.user)
     },
     existMessage(textileObj: any) {
-      const curUserMInfo = textileObj.conversations[this.user.address]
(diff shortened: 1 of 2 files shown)


components/views/user/User.vue
```diff
@@ -79,6 +79,26 @@ export default Vue.extend({
...
+    existMessage(textileObj: Conversation) {
+      const currentUserInfo = textileObj[this.user.address]
+      if (
+        currentUserInfo?.lastUpdate <= 0 &&
+        this.user.account.from === this.user.address
+      ) {
+        this.$data.existConversation = false
+      } else {
+        this.$data.existConversation = true
+      }
```
molimauro, line 91: here we can substitute the if-else with a guard ;)

davidchan1219: @molimauro I've just simplify the code

components/views/user/User.vue
```diff
@@ -79,6 +79,23 @@ export default Vue.extend({
       })
       this.$store.commit('ui/setUserProfile', this.user)
     },
+    existMessage(textileObj: Conversation) {
+      const currentUserInfo = textileObj[this.user.address]
+      this.$data.existConversation =
+        currentUserInfo?.lastUpdate <= 0 &&
+        this.user.account.from === this.user.address
+          ? false
+          : true
```
molimauro, line 88: at this point go this way: 
```
this.$data.existConversation = !(currentUserInfo?.lastUpdate <= 0 &&
this.user.account.from === this.user.address)
```

stavares843:
added changes requested based on mauro's review

components/views/user/User.vue
davidchan1219: @molimauro Thanks for your comment

molimauro: np!! :)

davidchan1219:
@stavares843 fixed

phillsatellite:
Tested: Seems like if you add a friend and the friend accepts it, they will appear in the Sidebar

https://user-images.githubusercontent.com/93608357/151889121-3da16e99-81d2-4649-9228-61ace533b3b6.mov

@davidchan1219

PR #1234, opened 2022-01-31
fix(actions): remove more option button AP-419

davidchan1219:
**What this PR does** 📖

Removed more option button

AP-419

### DIFF
--- a/components/views/chat/message/actions/Actions.html
+++ b/components/views/chat/message/actions/Actions.html
@@ -31,10 +31,4 @@
   >
     <edit-icon size="1x" :class="'control-icon'" />
   </div>
-  <!-- <div
-    class="reply-command has-tooltip has-tooltip-primary"
-    :data-tooltip="$t('ui.more')"
-  >
-    <more-vertical-icon size="1x" :class="'control-icon'" />
-  </div> -->
 </div>
--- a/components/views/chat/message/actions/Actions.html
+++ b/components/views/chat/message/actions/Actions.html
@@ -31,10 +31,10 @@
   >
     <edit-icon size="1x" :class="'control-icon'" />
   </div>
-  <div
+  <!-- <div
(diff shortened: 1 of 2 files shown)


components/views/chat/message/actions/Actions.html
```diff
@@ -31,10 +31,10 @@
...
     <edit-icon size="1x" :class="'control-icon'" />
   </div>
-  <div
+  <!-- <div
     class="reply-command has-tooltip has-tooltip-primary"
     :data-tooltip="$t('ui.more')"
   >
     <more-vertical-icon size="1x" :class="'control-icon'" />
-  </div>
+  </div> -->
```
Jekrimo, line 39: Can we just delete this whole commented code? Or are we planning to uncomment for later use?

Jekrimo (changes requested):

components/views/chat/message/actions/Actions.html
davidchan1219: @Jekrimo I removed that code.

stavares843:
i don't think the idea was to remove the button

i think the idea was something more

if there are options show the button

if not, hide the button

cc @WanderingHogan thoughts?

WanderingHogan:
@davidchan1219 like @stavares843 said - this isn't meant to remove the ellipses, just hide them if there are not things to show in the ellipses, more likely a v-if

WanderingHogan (changes requested):
hide if nothing in it, show if something in it, not remove code

davidchan1219:
@WanderingHogan There is no any item in the settings, so I made to hide the more icon.

stavares843:
@davidchan1219 feel free to remove changes requested label and add missing code review label when you address the feedback

stavares843:
and thanks for updating 🔨

phillsatellite (approved):

WanderingHogan (approved):

PR #1245, opened 2022-01-31
fix(call): dismiss incoming call when click deny button

genie-magic:
**What this PR does** 📖
fix(call): dismiss incoming call when click deny button

AP-524

### DIFF
--- a/components/ui/Global/Global.vue
+++ b/components/ui/Global/Global.vue
@@ -86,7 +86,7 @@ export default Vue.extend({
       peer?.call.deny()
 
       // peer?.send('SIGNAL', { type: 'CALL_DENIED' })
-      // this.$store.dispatch('webrtc/denyCall')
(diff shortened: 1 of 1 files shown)


phillsatellite:
Tested: Everything worked good only issue is it seems like if friend denies call, user is unable to call them again

https://user-images.githubusercontent.com/93608357/151887436-007a7c91-42ee-4770-a146-74bddb94573a.mov

@genie-magic

stavares843:
we should address phil's comment above

genie-magic:
> Tested: Everything worked good only issue is it seems like if friend denies call, user is unable to call them again
> 
>  secondcall.mov 
> @genie-magic

@phillsatellite   I think about that issue, we can make separate ticket.

iltumio (approved):

WanderingHogan (approved):

PR #1256, opened 2022-02-01
fix(edit): remove message edit option for files and glyphs, fix indxeddb bug for files AP-364

josephmcg:
**What this PR does** 📖
- remove edit option for files and glyphs
- fix bug where files were not being stored in indexeddb

AP-364

**Additional comments** 🎤


phillsatellite:
Tested: Everything seems to be working, the only question I have is I see conversation in the indexeddb but wasn't seeing the files. Wasn't sure if I was looking over it or if its an issue. Just lemme know 😄 
Screen Shot 2022-02-01 at 12 31 51 PM

phillsatellite:
Tested: Everything worked great, was going through to check that files were stored in Indexeddb and only saw the conversation. Wasn't sure if I was looking over it or if there was an issue. Just lemme know, thank you! 😄
Screen Shot 2022-02-01 at 10 22 27 AM

josephmcg:
Thanks for checking Phill, you need to expand that conversation object and check that every message is stored (specifically, the file or image messages). initially, the method to store in indexeddb wasn't being called correctly for file messages. you can compare the 'key' to the address from the url during a chat

phillsatellite:
> Thanks for checking Phill, you need to expand that conversation object and check that every message is stored (specifically, the file or image messages). initially, the method to store in indexeddb wasn't being called correctly for file messages. you can compare the 'key' to the address from the url during a chat

Awesome I see it now and its working as should. Thanks Joe!

phillsatellite (approved):

WanderingHogan (approved):

stavares843 (approved):

PR #1270, opened 2022-02-02
feat(user): fix timestamp for user list AP-118

davidchan1219:
**What this PR does** 📖

added correct timestamp for last message

AP-118


WanderingHogan (approved):
This is good right now, we will want to revisit when we make improvements to indexeddb

phillsatellite (approved):

PR #1328, opened 2022-02-04
fix(imageUpload): fix pending upload images stay in chat

KemoPaw:
… chat

**What this PR does** 📖
https://satellite-im.atlassian.net/browse/AP-113

Upload modal updates when you switch chats. 

**In the Index.vue file there is a `resetAllUploadedFiles`, not sure if this is the best practice. Other attempted methods did not work such as adding it in mutationsBlacklist in `persist.ts`**

**Additional comments** 🎤

### DIFF
--- a/components/views/files/upload/Upload.vue
+++ b/components/views/files/upload/Upload.vue
@@ -38,7 +38,7 @@ export default Vue.extend({
     },
     recipient: {
       type: Object as PropType<Friend>,
-      default: () => {},
+      required: true,
     },
   },
   data() {
@@ -59,9 +59,17 @@ export default Vue.extend({
     }
   },
   computed: {
-    ...mapState(['ui', 'friends']),
+    ...mapState(['ui', 'friends', 'chat']),
     activeFriend() {
       return this.$Hounddog.getActiveFriend(this.$store.state.friends)
+    },
+    uploadedFiles: {
+      get() {
+        return this.chat.uploadedFiles[this.$props.recipient?.address]
+      },
+      set(uploadFile) {
+        this.$store.commit('chat/addUploadedFile', uploadFile)
+      },
     },
   },
   methods: {
@@ -101,7 +109,7 @@ export default Vue.extend({
             )
           }
         }
-        this.$data.files = await Promise.all(
+        const tempFiles = await Promise.all(
           [...files].map(async (file: File) => {
             const uploadFile = {
               file,
@@ -111,7 +119,7 @@ export default Vue.extend({
             return uploadFile
           }),
         )
-        this.$data.files.every(async (uploadFile: UploadDropItemType) => {
+        tempFiles.every(async (uploadFile: UploadDropItemType) => {
           if (uploadFile.file.size <= Config.uploadByteLimit) {
             uploadFile.nsfw.checking = true
             try {
@@ -124,7 +132,6 @@ export default Vue.extend({
             }
             uploadFile.nsfw.checking = false
           }
-
           this.loadPicture(uploadFile)
         })
         this.$data.uploadStatus = true
@@ -145,6 +152,10 @@ export default Vue.extend({
       reader.onload = function (e: Event | any) {
         if (e.target) item.url = e.target.result
       }
+      this.uploadedFiles = {
+        file: item,
+        address: this.recipient.address,
+      }
       reader.readAsDataURL(item.file)
     },
     /**
@@ -154,15 +165,18 @@ export default Vue.extend({
      * @example @click="cancelUpload"
      */
     cancelUpload() {
-      this.$data.files = []
+      this.$store.commit('chat/clearUploadedFiles', this.recipient.address)
+      this.$data.files = this.uploadedFiles
       document.body.style.cursor = PropCommonEnum.DEFAULT
       this.$data.uploadStatus = false
       this.$data.count_error = false
       this.$parent.$data.showFilePreview = false
     },
     removeUploadItem(index: number) {
-      this.$data.files.splice(index, 1)
-      this.$data.files = [...this.$data.files]
+      this.$store.commit('chat/removeUploadedFile', {
+        address: this.recipient.address,
+        index: index,
+      })
       if (this.$data.files.length === 0) {
         document.body.style.cursor = PropCommonEnum.DEFAULT
         this.$data.uploadStatus = false
@@ -245,6 +259,19 @@ export default Vue.extend({
         this.dispatchFile(file)
       })
     },
+  },
+  watch: {
+    recipient() {
+      this.$store.commit(
+        'chat/initUploadedFiles',
+        this.$props.recipient?.address,
+      )
+      this.$data.files = this.uploadedFiles
+    },
+  },
+  mounted() {
+    this.$store.commit('chat/initUploadedFiles', this.$props.recipient?.address)
+    this.$data.files = this.uploadedFiles
   },
 })
 </script>
--- a/components/views/files/upload/Upload.vue
+++ b/components/views/files/upload/Upload.vue
@@ -65,11 +65,20 @@ export default Vue.extend({
     },
     uploadedFiles: {
       get() {
-        return this.chat.uploadedFiles[this.$props.recipient?.address]
+        return this.chat.uploadedFiles[this.$props.recipient?.address] ?? []
       },
       set(uploadFile) {
         this.$store.commit('chat/addUploadedFile', uploadFile)
       },
+    },
+  },
+  watch: {
+    recipient() {
+      this.$store.commit(
+        'chat/initUploadedFiles',
+        this.$props.recipient?.address,
+      )
+      this.$data.files = this.uploadedFiles
     },
   },
   methods: {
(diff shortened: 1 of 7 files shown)


components/views/files/upload/Upload.vue
```diff
@@ -246,6 +260,19 @@ export default Vue.extend({
       })
     },
   },
+  watch: {
+    recipient() {
+      this.$store.commit(
```
WanderingHogan, line 265: move these up, above methods. If you have ESLint installed in your IDE you should be able to 'quickfix' it
Screen Shot 2022-02-04 at 9 36 46 PM

pages/index.vue
```diff
@@ -55,6 +56,10 @@ export default Vue.extend({
       return
     }
     this.loadAccount()
+  },
+   destroyed(){
```
WanderingHogan, line 60: good idea

types/chat/chat.ts
```diff
@@ -7,3 +9,5 @@ export type ChatTextObj = {
   userId: string
   value: string
 }
+
+export interface UploadedFiles extends UploadDropItemType {}
```
WanderingHogan, line 13: Why did you extend this instead of just importing the type directly in store/chat/types.ts?

WanderingHogan (changes requested):
Minor eslint update and a question. 

Great job Kat

components/views/files/upload/Upload.vue
```diff
@@ -111,7 +119,7 @@ export default Vue.extend({
             return uploadFile
           }),
         )
-        this.$data.files.every(async (uploadFile: UploadDropItemType) => {
+        tempFiles.every(async (uploadFile: UploadDropItemType) => {
```
josephmcg, line 122: this should probably be a forEach since the boolean returned by every isn't being used

josephmcg, line 122: this should probably be a forEach since the boolean returned by every isn't being used

pages/index.vue
```diff
@@ -55,6 +56,10 @@ export default Vue.extend({
...
+    //Clear store of all pending photo uploads on close
+    this.$store.commit('chat/resetAllUploadedFiles')
```
josephmcg, line 62: I think we typically set this inside `plugins/thirdparty/persist.ts`

store/chat/mutations.ts
```diff
@@ -13,13 +14,48 @@ const mutations = {
...
+    state.uploadedFiles[address]
+      ? state.uploadedFiles[address].push(file)
+      : (state.uploadedFiles[address] = [file])
+  },
+  resetAllUploadedFiles(state: ChatState) {
+    state.uploadedFiles = {}
+  },
+  clearUploadedFiles(state: ChatState, address: string) {
+    state.uploadedFiles[address].splice(0, state.uploadedFiles[address].length)
+  },
```
josephmcg, line 43: I think it would be better to just delete uploadedFiles[address]
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

store/chat/mutations.ts
```diff
@@ -13,13 +14,48 @@ const mutations = {
...
+      address: string
+    },
+  ) {
+    state.uploadedFiles[address]
+      ? state.uploadedFiles[address].push(file)
+      : (state.uploadedFiles[address] = [file])
+  },
+  resetAllUploadedFiles(state: ChatState) {
+    state.uploadedFiles = {}
+  },
```
josephmcg, line 40: After configuring the anti-persist stuff, can delete

components/views/files/upload/Upload.vue
```diff
@@ -246,6 +260,19 @@ export default Vue.extend({
...
+        'chat/initUploadedFiles',
+        this.$props.recipient?.address,
+      )
+      this.$data.files = this.uploadedFiles
+    },
+  },
+  mounted() {
+    this.$store.commit('chat/initUploadedFiles', this.$props.recipient?.address)
+    this.$data.files = this.uploadedFiles
+  },
```
josephmcg, line 275: Could you explain your thought process for setting the empty array in `initUploadedFiles`? I feel like this could be avoided by using optional chaining when you're accessing the properties

components/views/files/upload/Upload.vue
```diff
@@ -59,10 +59,18 @@ export default Vue.extend({
...
   },
   computed: {
-    ...mapState(['ui', 'friends']),
+    ...mapState(['ui', 'friends', 'chat']),
     activeFriend() {
       return this.$Hounddog.getActiveFriend(this.$store.state.friends)
     },
+    uploadedFiles: {
+      get() {
+        return this.chat.uploadedFiles[this.$props.recipient?.address]
```
josephmcg, line 68: try this instead: 
`return this.chat.uploadedFiles[this.$props.recipient?.address] ?? []`

If there's an object, it returns it. If not, return whatever is after the `??`

josephmcg (changes requested):
Good job, let me know if you want to work through these changes together

types/chat/chat.ts
josephmcg: That's my fault, I thought it would be good to have the state related interface in that file to keep things consistent

components/views/files/upload/Upload.vue
josephmcg: also can you explain why you created the tempFiles variable?

store/chat/mutations.ts
cmckelvy42: We tried that initially but reassigning to a new pointer was causing all kinds of headaches. This was done to maintain the same reference across the app, as creating a new empty array prevents the modal from clearing properly after a successful send

cmckelvy42: > I think it would be better to just delete uploadedFiles[address], or just set to []. With some of my other feedback below though, I think delete is better https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

We tried that initially but reassigning to a new pointer was causing all kinds of headaches. This was done to maintain the same reference across the app, as creating a new empty array prevents the modal from clearing properly after a successful send

components/views/files/upload/Upload.vue
cmckelvy42: > also can you explain why you created the tempFiles variable?

Initially it used this.$data.files, but since that now functions solely as a reference to an array in the store, it can never be directly assigned, as that will cause an error due to modifying the store outside of a mutation. the tempFiles array exists only to be passed to loadPicture where the files are put into the array in the store

components/views/files/upload/Upload.vue
cmckelvy42: In the Upload.html, it references files.length
If there isn't already an instanced array, it will throw an error as soon as it tries to render because you're referring .length on an object that doesn't exist

pages/index.vue
cmckelvy42: We tried that, but it seems, at least as far as I can tell, that doing that simply prevents the store from persisting _when the listed mutations are made_ it does NOT stop the results of those mutations from being stored when a subsequent, non blacklisted mutation occurs. May be wrong on the reasoning, but at the very least, using the blacklist did not serve to actually wipe the files from the store on page refresh like we needed.

components/views/files/upload/Upload.vue
josephmcg: Reference the comment below

components/views/files/upload/Upload.vue
cmckelvy42: > try this instead: `return this.chat.uploadedFiles[this.$props.recipient?.address] ?? []`
> 
> If there's an object, it returns it. If not, return whatever is after the `??`

this doesn't work as it creates a new array outside of the store

pages/index.vue
josephmcg: I don't believe it would clear anything that was added before the non-persist config. I had to manually clear first, then it was working for me

components/views/files/upload/Upload.vue
josephmcg: ah I see, you could try returning a deep copy of the array in the getter rather than the files array itself.
https://lodash.com/docs/4.17.15#cloneDeep

types/chat/chat.ts
KemoPaw: > Why did you extend this instead of just importing the type directly in store/chat/types.ts?

Should I still do these changes?

WanderingHogan: no, i was just curious

WanderingHogan (approved):

josephmcg (approved):
Going to refactor later

WanderingHogan:
thank you all for sharing your thought process and putting this together

stavares843 (changes requested):
Captura de ecrã 2022-02-08, às 04 11 29

- go to a chat
- copy paste/drag 1 image
- change chats
- works

if you do the following
- go to a chat
- copy paste/drag 4 images
- change chat
- always crashes

doesn't happen on dev

WanderingHogan:
Good work, but we will come back to this after alpha. Closing for now, but please keep the branch

PR #1333, opened 2022-02-04
feat(bucket): add methods in Bucket retrieve Files Manager exports

josephmcg:
**What this PR does** 📖
- TextEncoder, TextDecoder don't exist by default in jsdom. needed to add or test would fail
- fix import and improve typing for FS export
- tests laid out in JIRA
- add init to classLoader for project

AP-468

- typescript has its own FileSystem interface, so we either need to rename or be careful with imports

image

**Additional comments** 🎤

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -1,0 +1,3 @@
+import { TextEncoder, TextDecoder } from 'util'
+global.TextEncoder = TextEncoder
+global.TextDecoder = TextDecoder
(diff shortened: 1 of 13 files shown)


stavares843:
@josephmcg pretty neat! thanks for adding the jest setup stuff and adding the test 🔨

stavares843:
about the FileSystem naming, maybe it should be renamed to avoid confusion

josephmcg:
> about the FileSystem naming, maybe it should be renamed to avoid confusion

agreed, will do that now

PR #1356, opened 2022-02-07
add(crypto, jest): functioning utilities and unit test initialization

drepram:
**What this PR does** 📖

Add utilities that resolve the error and the following unit test (`ReferenceError: TextEncoder is not defined`, `Crypto Instance not initialized`)

AP-334

> Not fully fixed, but it's AP-334. I suggest writing from scratch the unit test because... it's a bit weird? Some of the non-base58 errors were because the string passed into the function as parameters had `i`, ` ` (spaces), and other characters that weren't valid in base58.

This is a draft, I think I would need to spend more time on this; if it is considered worthy to be reworked!

**Additional comments** 🎤

### DIFF
--- a/libraries/Crypto/Crypto.ts
+++ b/libraries/Crypto/Crypto.ts
@@ -1,3 +1,4 @@
+import { TextEncoder, TextDecoder } from 'util'
 import { Keypair, PublicKey } from '@solana/web3.js'
 import ed2curve from 'ed2curve'
 import { sharedKey, signMessage } from 'curve25519-js'
--- a/libraries/Crypto/Crypto.test.ts
+++ b/libraries/Crypto/Crypto.test.ts
@@ -113,6 +113,8 @@ describe.skip('initializeRecipient', () => {
   })
 
   test('0', async () => {
+    inst.init(web3.Keypair.generate())
+    inst.signMessage(new Uint8Array([]), '')
     const param1: any = new web3.PublicKey(1)
     await inst.initializeRecipient(param1)
   })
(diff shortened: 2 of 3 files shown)


stavares843 (approved):

drepram:
I removed the draft label because I think this PR can be merged, and the urge to rewrite the Crypto unit test can be done in another PR and another ticket. What do you think?

libraries/Crypto/Crypto.ts
```diff
@@ -1,3 +1,4 @@
+import { TextEncoder, TextDecoder } from 'util'
```
WanderingHogan, line 1: do we need this for testing specifically?

stavares843: joe added this here
https://github.com/Satellite-im/Core-PWA/pull/1333/files#diff-3d96081f4e4a75728a1aec82be2ac7de4b46cb7fdad5b21f7d4b5fdcd98d3353

@drepram can you try to check if without the import will error anyway?

but yes if we don't add the TextEncoder no where, will return error saying ''ReferenceError: TextEncoder is not defined''

phillsatellite:
Tested and got this when trying to log in
Screen Shot 2022-02-08 at 3 36 38 PM

WanderingHogan:
I think we will need to remove the TextEncoder/TextDecoder import from the Crypto file, and use this:
https://github.com/facebook/jest/issues/9983#issuecomment-626489847

```
import util from 'util'
import * as web3 from '@solana/web3.js'
...

if (typeof TextEncoder === 'undefined') {
  global.TextEncoder = util.TextEncoder
}
```

drepram:
Hi, please check again. Thanks!

WanderingHogan (approved):

phillsatellite (approved):

PR #1393, opened 2022-02-09
add(jest): 100% code coverage for Peer.test.ts unit test

drepram:
**What this PR does** 📖

self-explanatory

AP-149

Something weird is happening in the send() function. I managed to cover the function while also mocking it by making two calls; 1 call that is before the function is mock, and 1 that is after the function is mock. Let me know if there is a better way.

**Additional comments** 🎤

### DIFF
--- a/libraries/WebRTC/Peer.test.ts
+++ b/libraries/WebRTC/Peer.test.ts
@@ -1,0 +1,18 @@
+import * as Peer from './Peer'
+
+describe('test alert methods', () => {
+  const AlertsConstructor = new Peer.Peer('originator', 'identifier', 'channel')
+  const inst = AlertsConstructor
+
+  test('get identifier', () => {
+    const result: any = inst.identifier
+    expect(result).toEqual('identifier')
+  })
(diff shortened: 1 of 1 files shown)


stavares843 (approved):

stavares843:
image

@phillsatellite noticed this

and this also happens on my side

because we use an M1 and we can't install that module lol

https://github.com/node-webrtc/node-webrtc/issues/698

this will fail on any local setup that is a M1, even though is not your fault 😂 

is there a way of adding tests for this without using that module directly?

stavares843 (changes requested):
added comment

drepram:
I'm not sure there is a way of writing unit tests without using that module at the moment. With that in mind, I think we need to put plans to unit testing WebRTC for later; what do you think?

stavares843:
agree

will close this one

stavares843:
Captura de ecrã 2022-06-10, às 23 43 10

no longer returns the old error but returns this one

can check out this test again and fix it?  🔨 

@drepram

PR #1420, opened 2022-02-09
feat(chatImageUpload): have images persist in each chat

cmckelvy42:
**What this PR does** 📖
https://satellite-im.atlassian.net/browse/AP-113
Upload modal changes when switching between chats and preserves pending uploads. No longer uses store due to store size limitations

changed the way uploads work so that pressing the upload button again with pending uploads will add to the list instead of replacing the list of pending uploads. However, line 99 of Upload.vue checks if over 8 files have been selected and throws an error if this is the case, and I'm not sure if this should be updated to check the count of files already in the list as well.
**Additional comments** 🎤

### DIFF
--- a/components/views/files/upload/Upload.vue
+++ b/components/views/files/upload/Upload.vue
@@ -43,6 +43,7 @@ export default Vue.extend({
   },
   data() {
     return {
+      filesDB: {},
       files: [] as Array<UploadDropItemType>,
       uploadStatus: false,
       count_error: false,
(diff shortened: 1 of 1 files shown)


stavares843:
> However, line 99 of Upload.vue checks if over 8 files have been selected and throws an error if this is the case, and I'm not sure if this should be updated to check the count of files already in the list as well.

yes, we can only send 8 at this moment, so only 8 as max number of images should be allowed on the chat bar section

cmckelvy42:
> > However, line 99 of Upload.vue checks if over 8 files have been selected and throws an error if this is the case, and I'm not sure if this should be updated to check the count of files already in the list as well.
> 
> yes, we can only send 8 at this moment, so only 8 as max number of images should be allowed on the chat bar section

this should be implemented now

components/views/files/upload/Upload.vue
```diff
@@ -64,6 +65,16 @@ export default Vue.extend({
       return this.$Hounddog.getActiveFriend(this.$store.state.friends)
     },
   },
+  watch: {
+    recipient() {
+      this.$data.filesDB[this.recipient.address] ||= []
+      this.$data.files = this.$data.filesDB[this.recipient.address]
```
WanderingHogan, line 71: I think you can chain assignment here, eg `this.$data.files = this.$data.filesDB[this.recipient.address] ||= []`, but this is okay, too

WanderingHogan (approved):
This works well, thank you, Chris

stavares843 (approved):

PR #1426, opened 2022-02-10
feat(files): Files structure and implementation AP-469

josephmcg:
I know this isn't necessarily within the scope of the ticket, but Hogan and I thought it would be much easier to plug in textile after updating the UI to use the new classes/object shapes. Will plug it into textile after this gets merged/

**What this PR does** 📖
- implement new file system class
- allow creation of folders and dummy files on files page
- traverse directories
- liked/shared status for folders/files. Somewhat hacky solution though since Map has reactivity issues in Vue2

AP-469 (kind of. needs additional textile work, but this PR was getting too big)

- since the bucket needs a fileSystem in the constructor, I thought we may as well just use that one. let me know if this isn't what you had in mind. Maybe the bucket should just store the import instead, but then we should probably remove the required FS from the constructor.
- we may consider switching the children map on Directory to an array for better reactivity with Vue2. I've implemented hacky workarounds for any issues so far, but may not scale well (depending on how many reactive updates these items need to support). Vue3 supports Map once we upgrade
https://stackoverflow.com/questions/37130105/does-vue-support-reactivity-on-map-and-set-data-types

**Additional comments** 🎤
- How can we take advantage of the errors being thrown inside the classes? for instance, right now you can't create another child with the same name as an existing child. I was having trouble catching this error, then showing a corresponding message
- I needed to target a ref of a child component in `components/views/files/controls/Controls.vue`, let me know if my technique is bad


components/views/files/controls/Controls.vue
```diff
@@ -40,8 +39,24 @@ export default Vue.extend({
...
+      }
+      if (this.input.type === FilesViewEnum.FOLDER) {
+        // add folder
+        this.$store.commit('bucket/createDirectory', this.text)
+        this.input.show = false
+        this.text = ''
+        this.$emit('forceRender')
+        return
+      }
+      // add file todo - gather description
```
stavares843, line 54: added ticket - [AP-762](https://satellite-im.atlassian.net/browse/AP-762)

libraries/Files/Fil.ts
```diff
@@ -52,6 +52,11 @@ export class Fil extends Item {
     return new Fil(`${this.name} copy`, this._description, this._hash)
   }
 
+  // todo store file size after textile upload
```
stavares843, line 55: added ticket - [AP-763](https://satellite-im.atlassian.net/browse/AP-763)

libraries/Files/FilSystem.ts
```diff
@@ -190,20 +190,20 @@ export class FilSystem {
...
+  //   console.log(
+  //     `\n[${this.currentDirectoryPath.join('/')}]:` +
+  //       (this.currentDirectory.content
+  //         .map(
+  //           (item) =>
+  //             `\n[${item.constructor.name.substring(0, 1)}]-> ${item.name}`,
+  //         )
+  //         .join('') || '\n(empty)'),
+  //   )
+  // }
```
stavares843, line 206: do we need this block commented-out? 🔨

pages/files/browse/index.vue
```diff
@@ -36,100 +42,34 @@ export default Vue.extend({
...
+    // },
+    // loadPicture(file: File) {
+    //   if (!file) return
+    //   const self = this
+    //   const reader = new FileReader()
+    //   reader.onload = function (e: Event | any) {
+    //     if (e.target) self.$data.url = e.target.result
+    //   }
+    //   reader.readAsDataURL(file)
+    // },
```
stavares843, line 72: do we need this commented-out? 🔨  or is this a TODO thats not ready yet? 💥

phillsatellite:
This feature is super cool, only I could things I noticed when testing: 
Should the file icon and folder icon be different? 
Screen Shot 2022-02-10 at 8 35 47 PM

and the other thing is if you select new folder/file and then press enter right after it'll close the pop up for the new file/folder
https://user-images.githubusercontent.com/93608357/153526305-1e31c4dd-c9a0-4d2d-a67f-63606bc55fef.mov

libraries/Files/FilSystem.ts
josephmcg: I wanted to confirm with Matt if we can remove it. It just prints to the console. I'll remove and he will say something in review if he wants to keep it

pages/files/browse/index.vue
josephmcg: I'm gonna work on the file upload feature next, so I left it there as a note of the various steps I need to take (nsfw check, show thumbnail of upload if image). I can add a todo to make that more clear

josephmcg:
@phillsatellite 
Thanks Phill, fixing icons right now.

The reason the input is disappearing is because the add folder/file button is still focused after click (we can't see it due to css outlines being hidden, which is something we will fix in the [accessibility ticket](https://satellite-im.atlassian.net/browse/AP-521)). If you hit tab once or twice, you should see the focus switch to the input box. I'll work on a fix to automatically switch focus

josephmcg:
will reopen when it's ready

PR #1457, opened 2022-02-11
fix(result): fix the result count show issue AP-768

davidchan1219:
**What this PR does** 📖

Show correct counts info when search messages

AP-768

### DIFF
--- a/components/interactables/Search/Result/Result.vue
+++ b/components/interactables/Search/Result/Result.vue
@@ -146,7 +146,7 @@ export default Vue.extend({
       handler(newResult) {
         if (newResult && newResult.data) {
           this.$data.groupList.map((item: any) => {
-            if (item.title === 'Messages') {
+            if (item.type === SearchResultGroupType.Messages) {
               item.count = newResult.data.totalRows
             }
             return item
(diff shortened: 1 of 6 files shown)


components/interactables/Search/Result/Result.vue
```diff
@@ -143,6 +142,18 @@ export default Vue.extend({
         }
       },
     },
+    result: {
+      handler(newResult) {
+        if (newResult && newResult.data) {
+          this.$data.groupList.map((item: any) => {
```
molimauro, line 148: Please dont use `.map` if not returning anything. Just go for `for/forEach` much cleaner and efficient. Thanks

components/interactables/Search/Result/Result.vue
```diff
@@ -143,6 +142,18 @@ export default Vue.extend({
...
+            if (item.title === 'Messages') {
```
molimauro, line 149: Will this `if` work if we change language?

components/interactables/Search/Result/Result.vue
davidchan1219: Yes. I know that. That loop has return value. So I used map.

components/interactables/Search/Result/Result.vue
davidchan1219: I changed the condition as "type".

phillsatellite:
Tested: Seems like whenever user tries to use search bar they get error: "_item$payload.toLowerCase is not a function. Tested Locally when I got this crash

Uploading crash.mov…


https://user-images.githubusercontent.com/93608357/153944962-2dccdb2c-1ef6-4e39-a572-1189f1a7ab03.mov

davidchan1219:
@phillsatellite I just fixed the code. Could you test the above case with new code again?

WanderingHogan:
Closing, we will try to incorporate these in the search branch

PR #1460, opened 2022-02-11
fix(indexeddb): update sort logic AP-709

davidchan1219:
**What this PR does** 📖

Implement sorting on basic search library

AP-709

### DIFF
--- a/libraries/IndexedDB/index.ts
+++ b/libraries/IndexedDB/index.ts
@@ -9,7 +9,7 @@ export const searchMessage = async (
   page: number = 1,
 ) => {
   console.log('queryOptions: ', queryOptions)
-  const { friends, queryString, dateRange } = queryOptions
+  const { friends, queryString, dateRange, orderBy } = queryOptions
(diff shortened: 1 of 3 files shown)


libraries/IndexedDB/index.ts
```diff
@@ -32,7 +32,13 @@ export const searchMessage = async (
   })
 
   const result = newResult
-    .sort((item1, item2) => item2.at - item1.at)
+    .sort((item1, item2) =>
+      orderBy === 'new'
+        ? item2.at - item1.at
+        : orderBy === 'old'
+        ? item1.at - item2.at
+        : item2.at - item1.at,
```
molimauro, line 40: It seems that the only diff case is `orderBy ===  'old'`. Can this be refactored to `orderBy === 'old' ? item1.at - item2.at : item2.at - item1.at`?

davidchan1219: Thanks.

WanderingHogan:
Closing, we will try to incorporate these in the search branch

PR #1495, opened 2022-02-15
fix(search): fix "x" in search bar is cutoff

genie-magic:
**What this PR does** 📖
fix "x" in the search bar is cutoff

AP-770


KemoPaw (approved):
Works like a charm! 👍

phillsatellite (approved):

PR #1500, opened 2022-02-15
fix(encryption): encrypt group chat messages

Jekrimo:
**What this PR does** 📖

Encrypts and decrypts messages sent to group chats

Ap 721

**Additional comments** 🎤

I think testing would just be making sure messages are still sending, unsure though if there is a better way

### DIFF
--- a/libraries/Textile/GroupChatManager.ts
+++ b/libraries/Textile/GroupChatManager.ts
@@ -1,7 +1,9 @@
-import { PublicKey, ThreadID, Update } from '@textile/hub'
+import { Identity, PrivateKey, PublicKey, ThreadID, Update } from '@textile/hub'
 import { Query } from '@textile/threads-client'
 import { isRight } from 'fp-ts/lib/Either'
 import { v4 as uuid } from 'uuid'
+import Vue from 'vue'
+import { Keypair } from '@solana/web3.js'
 import { messageEncoder } from './encoders'
 import {
   ConversationQuery,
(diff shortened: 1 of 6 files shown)


libraries/Textile/GroupChatManager.ts
```diff
@@ -91,7 +101,18 @@ export class GroupChatManager {
       groupChatID,
       groupChatQuery,
     )
-    const promises = messages.map<Promise<Message>>(this.decodeMessage)
+    const group = Groups.find((group) => {
+      if (group.address === groupChatID) {
+        return group.encryptionKey
```
WanderingHogan, line 106: returns the same thing

WanderingHogan: Actually, the way it's set up it returns the first object every time

libraries/Textile/GroupChatManager.ts
```diff
@@ -91,7 +101,18 @@ export class GroupChatManager {
...
+      }
+      return undefined
```
WanderingHogan, line 108: you don't need undefined, but doesn't hurt https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#return_value

phillsatellite:
I noticed img previewer works on group chats now! Looks awesome 💥  should they send as well or is that to be added in future pr? 
Screen Shot 2022-02-16 at 2 34 32 PM

Jekrimo:
@phillsatellite I'm workin on a separate pr for getting our other message types right now that will hopefully be done today!

stavares843:
thanks @Jekrimo 🔨

phillsatellite (approved):

iltumio (approved):

stavares843 (approved):

PR #1509, opened 2022-02-16
feat(files): File system functionality AP-469

josephmcg:
**What this PR does** 📖
- implement new file system class
- create folders and files. When uploading files, it's currently just ripping some data (name, size) and saving that. Actual file needs to be uploaded to textile and then store the hash.
  - nsfw filter for file upload
  - proper error handling/notification if you try to add a file with the same name
  - went with new upload design to handle multiple files more easily and improve ux (confirmed with Liz)
- traverse directories
- liked status toggle for folders/files. Map has funky reactivity for Vue2, so I had to do a workaround to force render

AP-469, AP-640

- removed the fileSystem from the Bucket class. It seemed unnecessary, since we already have the FileSystem global we can use. Bucket will store the export
- switched up the Fil constructor and fs.createFile to make it easier to use (objects rather than regular params). updated tests accordingly
- we may consider switching the children map on Directory to an array for better reactivity with Vue2. I've implemented hacky workarounds for any issues so far, but may not scale well (depending on how many reactive updates these items need to support). Vue3 supports Map once we upgrade
https://stackoverflow.com/questions/37130105/does-vue-support-reactivity-on-map-and-set-data-types

**Additional comments** 🎤


josephmcg:
@RetroPronghorn Right on, just added

RetroPronghorn (approved):

WanderingHogan:
@stavares843 @phillsatellite there isn't much to test here other than the app doesn't crash, with the uploading and management of files coming soon

stavares843 (changes requested):
image

crash's when uploading a .heic file, standard image on iOS

stavares843 (changes requested):
Captura de ecrã 2022-02-16, às 18 11 22

same crash, with the .TIFF file

stavares843:
i know its all the same crash but added in case it was useful info 🔨

josephmcg:
Thanks Sara, I'll work on supporting these file types

josephmcg:
This is happening because nsfwjs can only scan a select few image types. The chat image upload is running into the same problem, but its wrapped in a try catch that marks any image unable to be scanned as nsfw. I think I'll refactor the nsfw scan function to only check supported file types to avoid a runtime error and provide proper support for people who want to sent/upload tiff

image

image

josephmcg:
@stavares843 ready for testing again. We won't allow tiff for now, but I will resolve that in AP-807 later. heic should work though. And we should block the upload if the file is an image that can't be scanned for any reason

stavares843:
thanks 🔨

PR #1527, opened 2022-02-17
feat(friendRequest): cancel sent friend request AP-701

josephmcg:
**What this PR does** 📖
- cancel a friend request that hasn't been accepted yet

AP-701

- I noticed it doesn't automatically disappear for the receiver. I think we may need something else in the `libraries/Solana/FriendsProgram/FriendsProgram.ts`, but I'm not entirely sure how that works.
  - should this be resolved in this ticket or would this be considered a separate issue?

**Additional comments** 🎤


josephmcg:
closed for now. probably better after Manuels PR goes through

PR #1559, opened 2022-02-18
fix(imageUpload): fixes uploaded images being added to the wrong chat when changing during NSFW check

cmckelvy42:
**What this PR does** 📖
assigns the address in the store for uploaded images to be the one from the time the upload is initiated, not the one at the time the file has finished processing.
Previously, changing chats while an image is in the NSFW check phase would cause it to get added to the wrong chat's files array. This fixes this.

**Additional comments** 🎤
the store solution @josephmcg used for upload persists is a nice one that solves some problems I couldn't figure out how to work around, and investigating how his solution worked is what led me to finding this bug that would've affected both our solutions

### DIFF
--- a/components/views/files/upload/Upload.vue
+++ b/components/views/files/upload/Upload.vue
@@ -74,6 +74,7 @@ export default Vue.extend({
     },
   },
   mounted() {
+    console.log('test')
     this.files = cloneDeep(this.chat.files?.[this.recipient.address]) ?? []
   },
   methods: {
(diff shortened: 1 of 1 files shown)


molimauro (approved):

molimauro:
Can you add this line `this.$parent.$data.showFilePreview = this.files.length > 0` in the recipient watcher (line 73)? This is needed because if you go from a chat with files to another with 0 files the chatbar corners remain sharp. Thanks!

josephmcg (approved):

PR #1563, opened 2022-02-18
fix(timezone): fix timezone cut issues on mobile

genie-magic:
**What this PR does** 📖
fix timezone cut issues on mobile

AP-727

### DIFF
--- a/components/views/settings/profile/card/Card.less
+++ b/components/views/settings/profile/card/Card.less
@@ -48,8 +48,8 @@
         &-timezone {
           .timezone-list {
             .items {
-              width: calc(100vw - 5rem) !important;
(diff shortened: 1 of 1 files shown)


genie-magic:
@molimauro timezone is very long so.. I just set 25rem for wide screen so that user can read more as they can...
image
image

molimauro:
Ok now i understand. Can we put `100%` instead of `calc(100vw - 5rem)`? I think that we should then revisit this part after alpha from a UI perspective if we want to keep long strings. Thank you @genie-magic

genie-magic:
@molimauro  I updated.

molimauro (approved):

stavares843 (approved):

PR #1565, opened 2022-02-18
fix(video): video track appear black on iOS

vimercati-samir:
What this PR does 📖
Fix black video track on iOS

AP-416

Additional comments 🎤


WanderingHogan (approved):

stavares843:
if can address that as well would be neat 💥

vimercati-samir:
I'm looking into it 👍

phillsatellite (approved):

components/views/media/user/User.vue
```diff
@@ -38,6 +38,38 @@ export default Vue.extend({
...
+      this.playVideo()
+    }
+  },
+  methods: {
+    playVideo() {
+      const video = document.querySelector(
+        `#${this.isLocal ? 'local' : 'remote'}-video`,
+      ) as HTMLVideoElement
+
+      video.play()
```
molimauro, line 71: ref is not working?

components/views/media/user/User.vue
```diff
@@ -38,6 +38,38 @@ export default Vue.extend({
...
+  },
+  watch: {
+    propsToWatch({ videoStream }) {
+      if (videoStream) {
+        this.$nextTick(() => {
+          this.playVideo()
+        })
+      }
+    },
+  },
```
molimauro, line 59: If your goal is to watch for props change, you can remove the computed and watch directly for `videoStream(value)` instead of `propsToWatch({ videoStream })`

components/views/media/user/User.vue
vimercati-samir: ref doesn't work in this specific case because the video element is inside a parent that is rendered conditionally with v-if

components/views/media/user/User.vue
vimercati-samir: I did it like this because I thought it was a pattern (I found this pattern in other instances in the code), should I change it as you suggested?

components/views/media/user/User.vue
molimauro: Ok then! Can you add a check (or `?`) before calling the `.play` function?

components/views/media/user/User.vue
molimauro: Yes please! I found the same pattern on just another file in `views/media/Media.vue` can you change it there as well? thanks a lot

components/views/media/user/User.vue
vimercati-samir: Yes, It should never return null but it's a good practice to check either way

components/views/media/user/User.vue
vimercati-samir: Sure 👍

components/views/media/user/User.vue
vimercati-samir: changed accordingly in commit 975cb2e1cebe58eedbf64d4224becf4366c91f4c

WanderingHogan (approved):

molimauro (approved):

phillsatellite (approved):

PR #1578, opened 2022-02-18
chore(media): when a track is found, called one stops blinking

ThomBos:
**What this PR does** 📖
Adds back again the "connecting animation", whenever a user got connected it stops the animation.
Fixed media width (now it's fullwidth)

AP-794

I know for sure there's a cleaner code... If someone has any idea, let me know that i change my code! 👍 
By fixing media width, i found out the blink was still on after the other user joined the call, so i removed that too. (my bad i didn't open a new ticket for that)
**Additional comments** 🎤

### DIFF
--- a/components/views/media/Media.html
+++ b/components/views/media/Media.html
@@ -20,7 +20,7 @@
       :videoStream="remoteVideoStream"
       :audioStream="remoteAudioStream"
       :audioMuted="remoteAudioMuted"
-      :calling="remoteTracks"
+      :calling="!remoteTracks"
     />
 
     <div
--- a/components/views/media/Media.html
+++ b/components/views/media/Media.html
@@ -20,6 +20,7 @@
       :videoStream="remoteVideoStream"
       :audioStream="remoteAudioStream"
       :audioMuted="remoteAudioMuted"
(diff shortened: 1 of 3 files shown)


components/views/media/Media.html
```diff
@@ -20,6 +20,7 @@
       :videoStream="remoteVideoStream"
       :audioStream="remoteAudioStream"
       :audioMuted="remoteAudioMuted"
+      :calling="remoteTracks"
```
molimauro, line 23: Based on the refactor made by Jason on this component i guess this line should be changed to `calling` and nothing else. You can remove all the other code about `remoteTracks()`

molimauro (changes requested):

molimauro:
@ThomBos can you please verify that my assumptions are correct? Change the code and test again thank you! Hit me up for any clarifications

molimauro:
Ok, after a quick call btw me and @ThomBos we managed to resolved the issue. This PR resolves 2 issues: the width as reported in the ticket and the blinking around the user circle to stop when the user answered the call.

molimauro (approved):

phillsatellite (approved):

PR #1581, opened 2022-02-21
chore(jest): add test

stavares843:
**What this PR does** 📖

Adds store/bucket/state.test.ts


josephmcg:
I'm actually removing this from store in my current branch. Opting to use the global instead

stavares843:
thanks 😂

PR #1673, opened 2022-02-23
feat(global): routes to user when accepting call

jpanay:
**What this PR does** 📖
When accepting a call routes user to chat for whoever is calling
AP-710

**Additional comments** 🎤

### DIFF
--- a/components/ui/Global/Global.vue
+++ b/components/ui/Global/Global.vue
@@ -83,6 +83,9 @@ export default Vue.extend({
       if (this.$route.path !== callingPath) {
         this.$router.push(callingPath)
       }
+      if (this.ui.showSettings) {
+        this.$store.commit('ui/toggleSettings', { show: false })
+      }
     },
     /**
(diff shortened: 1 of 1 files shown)


molimauro (approved):

Jekrimo (approved):

phillsatellite:
Tested: If a user is in Settings and they answer a incoming call, they can answer but they will still be in Settings, is that intended? If thats intended feature works great 👍 

https://user-images.githubusercontent.com/93608357/155385475-45885d78-acab-4b74-ba2e-1f6fe30bf3f5.mov

jpanay:
> Tested: If a user is in Settings and they answer a incoming call, they can answer but they will still be in Settings, is that intended? If thats intended feature works great 👍
> 
>  RPReplay_Final1645641332.2.mov

It should exit settings when accepting I'll change that now

phillsatellite (approved):

stavares843 (approved):

PR #1682, opened 2022-02-24
fix(webrtc): hangs up call when accepting a new call

jpanay:
**What this PR does** 📖
Stops users from being able to be in 2 calls by hanging up current call when accepting a new call
AP-750

**Additional comments** 🎤

### DIFF
--- a/components/views/media/Media.vue
+++ b/components/views/media/Media.vue
@@ -36,6 +36,12 @@ export default Vue.extend({
   },
   computed: {
     ...mapState(['ui', 'accounts', 'friends', 'webrtc']),
+    isActiveCall() {
+      return this.friends.all.find(
+        (friend: any) =>
+          friend.activeChat && friend.address === this.webrtc.activeCall,
+      )
+    },
     localAudioMuted() {
       return this.webrtc.localTracks.audio.muted
     },
--- a/components/ui/Global/Global.vue
+++ b/components/ui/Global/Global.vue
@@ -14,7 +14,7 @@ declare module 'vue/types/vue' {
 export default Vue.extend({
   name: 'Global',
   computed: {
-    ...mapState(['ui', 'media']),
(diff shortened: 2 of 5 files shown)


phillsatellite:
Tested: Tested a couple times and tried starting from a new account but was still able to be in 2 calls at once 

https://user-images.githubusercontent.com/93608357/155577737-87998717-75be-4169-909e-91510d888aa3.mov

jpanay:
> Tested: Tested a couple times and tried starting from a new account but was still able to be in 2 calls at once 
> 
> 
> 
> https://user-images.githubusercontent.com/93608357/155577737-87998717-75be-4169-909e-91510d888aa3.mov
> 
> 
> 
> 
> 
> 

Hmm, I'll do some investigating

stavares843:
@jpanay any news? 💥

stavares843:
thanks 🎉

components/views/media/Media.vue
```diff
@@ -36,6 +36,12 @@ export default Vue.extend({
   },
   computed: {
     ...mapState(['ui', 'accounts', 'friends', 'webrtc']),
+    isActiveCall() {
+      return this.friends.all.find(
+        (friend: any) =>
+          friend.activeChat && friend.address === this.webrtc.activeCall,
```
WanderingHogan, line 42: Can you move the 'friend.activeChat' outside of the find method, no sense in running the find if activeChat is already ''

WanderingHogan (changes requested):
please check the comments and let me know

components/views/media/Media.vue
jpanay: We could do if activechat === activecall but it's necessary to have both

phillsatellite (approved):

WanderingHogan (approved):

stavares843 (approved):

PR #1686, opened 2022-02-24
feat(jest): initialize unit test for store/webrtc/actions

drepram:
**What this PR does** 📖

Working on this wouldn't be too hard, but as Sara knows: WebRTC doesn't quite run in M1 machines. Hence, My current work can just be added so that when the time comes where that bug has been patched, development can be started again.

No ticket at the moment.

**Additional comments** 🎤


phillsatellite:
tested: seems like user gets error at  FAIL  store/webrtc/actions.test.ts

stavares843:
as discussed, even if the test is skipped it will require the module that doesn't have m1 support

closing this for now

PR #1715, opened 2022-02-25
feat(filesLimit): handle if user goes over size limit

josephmcg:
**What this PR does** 📖
- add getter to calculate the total size of all tracked files
- update files aside to track your files usage (amount and progress bar)
- prevent further uploads if you're out of room

AP-899, AP-910

- Originally wanted to use a computed for the progress bar, but it would cause an infinite render issue. opted for watch instead

**Additional comments** 🎤

### DIFF
--- a/components/ui/Progress/Progress.vue
+++ b/components/ui/Progress/Progress.vue
@@ -1,4 +1,4 @@
-<template src="./Progress.html" />
+<template src="./Progress.html"></template>
 <script lang="ts">
 import Vue from 'vue'
 
--- a/components/views/files/aside/Aside.html
+++ b/components/views/files/aside/Aside.html
@@ -11,7 +11,7 @@
   <UiProgress :progress="progress" />
   <div>
     <span class="usage-text">
-      <span class="red">{{totalSize}}</span> of 4GB
+      <span class="red">{{totalSize}}</span> of {{sizeLimit}}
     </span>
   </div>
   <UiSpacer :height="10" />
--- a/components/views/files/aside/Aside.html
+++ b/components/views/files/aside/Aside.html
@@ -12,7 +12,7 @@
   <UiProgress :progress="progress" />
   <div>
     <span class="usage-text">
-      <span class="red">{{totalSize}}</span> of {{sizeLimit}}
+      <span :class="sizeColor">{{totalSize}}</span> of {{sizeLimit}}
     </span>
   </div>
   <UiSpacer :height="10" />
--- a/components/views/files/aside/Aside.html
+++ b/components/views/files/aside/Aside.html
@@ -8,9 +8,11 @@
       :action="() => {}"
     />
   </div>
-  <UiProgress :progress="80" />
+  <UiProgress :progress="progress" />
   <div>
-    <span class="usage-text"> <span class="red">3.8GB</span> of 4GB </span>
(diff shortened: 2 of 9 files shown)


KemoPaw (commented):
Very cool to see the progress bar updating upon file uploads! 🎉  Sadly do not currently have a 4GB file to test for if a user can continue to upload past that point.

josephmcg:
I didn't either, I tested it by changing the limit to a smaller number

phillsatellite (approved):

phillsatellite:
Tested: works really well, the only bug I was able to find is if a user has stuff uploaded and the refresh they're page, all the files will still be there but the progress bar will be empty until a user uploads another file

https://user-images.githubusercontent.com/93608357/156032800-0d4641a2-5726-4ceb-be26-2f12398e66e4.mov

josephmcg:
@phillsatellite Thanks Phil! should be fixed now

stavares843:
thanks @josephmcg 🔨

josephmcg:
@stavares843 I noticed something I need to fix here, will cause bugs after file type support gets merged

josephmcg:
ok, now its good

stavares843:
Captura de ecrã 2022-03-01, às 14 25 42

I uploaded couple of files, but the bar is the same


I also wonder if when is such a small number or even 0 if the storage used number maybe shouldn't be red

Captura de ecrã 2022-03-01, às 14 24 41

stavares843:
Captura de ecrã 2022-03-01, às 14 29 18

now that I uploaded a couple more files, the progress bar is updated. But if you only update small files, as in the previous screenshots the progress bar stays empty.

stavares843:
also, seems if I try to upload more than the current storage available, doesn't upload, doesn't return an error to the user, and also doesn't allow the user to try to upload any more files unless the user refreshes the page

https://user-images.githubusercontent.com/29093946/156188356-dff2f3f8-8f81-489d-98ab-5041c1cbda98.mov

josephmcg:
- 4GB is 4e+9 bytes. A few bytes won't 'move the needle', so to speak. I'm not sure what the threshold is for a visible indication on <progress>. We could potentially add Math.ceil to it if you'd like to always see a bar, but that may have unintended results past 1%.
- What should the threshold be for turning the text red? >80% filled?
- I've never tried uploading several GB at once, I'll look into that

stavares843:
that's true! 

maybe we could ask Liz about the 2 first questions as they are more UI/UX related

josephmcg:
@stavares843 issues above should be fixed. I was setting the loading state after checking filesize. It was working correctly in your video, it just takes a LONG time for files that large
It seems like the filesize package we're using changed their default byte base from 2 to 10. I added some config to change it to 2 again, now numbers should be more accurate

stavares843:
thanks @josephmcg 

we could ask if there's any preference, either decimal or binary

molimauro (approved):

phillsatellite:
@josephmcg hello!! Last little problem I found on this branch, after updates MB and GB changed to MiB and GiB
Screen Shot 2022-03-03 at 2 39 30 PM

josephmcg:
yes, they were using binary. just switched it to decimal

stavares843 (approved):

PR #1716, opened 2022-02-25
fix(files): show image if possible and css cleanup

josephmcg:
**What this PR does** 📖
- show image preview on file view if possible.
- clean up css

AP-902, AP-900

**Additional comments** 🎤

### DIFF
--- a/components/views/files/view/View.vue
+++ b/components/views/files/view/View.vue
@@ -33,9 +33,11 @@ export default Vue.extend({
   },
   computed: {
     ...mapState(['ui']),
-
     path(): string {
       return this.$Config.textile.browser + this.file.hash
+    },
+    isImage() {
+      return this.file.name.match(this.$Config.regex.image)
     },
   },
   methods: {
--- a/components/views/files/view/View.vue
+++ b/components/views/files/view/View.vue
@@ -36,8 +36,8 @@ export default Vue.extend({
     path(): string {
       return this.$Config.textile.browser + this.file.hash
     },
-    isImage() {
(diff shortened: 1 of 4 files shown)


phillsatellite:
Tested: @josephmcg should .heic images preview as well? 
Screen Shot 2022-02-25 at 12 17 48 PM

josephmcg:
heic cannot be displayed in browser

components/views/files/view/View.vue
```diff
@@ -33,10 +33,12 @@ export default Vue.extend({
   },
   computed: {
     ...mapState(['ui']),
-
     path(): string {
       return this.$Config.textile.browser + this.file.hash
     },
+    isImage() {
+      return this.file.name.match(this.$Config.regex.image)
```
molimauro, line 40: Since is a boolean can we return `Boolean(this.file.name.match(this.$Config.regex.image))`? :)

molimauro:
Only one super minor thing, other than that we're good!

components/views/files/view/View.vue
josephmcg: Thanks Mauro, good call. I added it to some other computed as well

WanderingHogan:
@josephmcg I just tried uploading an heic to box and google drive and both preview it. you can create another ticket if Sara and Phil are okay with it to address it

WanderingHogan (approved):

josephmcg:
We should probably handle HEIC in a separate ticket, I'm still not sure what the best course of action is

molimauro (approved):

phillsatellite (approved):

PR #1724, opened 2022-02-25
fix(chatbar): fixed persistent "is typing" bug

ThomBos:
**What this PR does** 📖
Fixes persistent "is typing" bug
AP-830

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/chatbar/Chatbar.vue
+++ b/components/views/chat/chatbar/Chatbar.vue
@@ -52,7 +52,6 @@ export default Vue.extend({
       recipientTyping: false,
       showFilePreview: false,
       nsfwUploadError: false,
-      chatEventsTimer: 5,
     }
   },
   computed: {
@@ -143,8 +142,9 @@ export default Vue.extend({
       handler() {
         const activeFriend = this.$Hounddog.getActiveFriend(this.friends)
 
-        this.$data.recipientTyping =
-          activeFriend?.typingState === PropCommonEnum.TYPING
+        if (activeFriend)
+          this.$data.recipientTyping =
+            activeFriend.typingState === PropCommonEnum.TYPING
       },
       deep: true,
     },
--- a/components/views/chat/chatbar/Chatbar.vue
+++ b/components/views/chat/chatbar/Chatbar.vue
@@ -51,10 +51,11 @@ export default Vue.extend({
       recipientTyping: false,
       showFilePreview: false,
       nsfwUploadError: false,
+      chatEventsTimer: 5,
     }
   },
   computed: {
-    ...mapState(['ui', 'friends', 'chat', 'textile']),
+    ...mapState(['ui', 'friends', 'webrtc', 'chat', 'textile']),
     activeFriend() {
       return this.$Hounddog.getActiveFriend(this.friends)
     },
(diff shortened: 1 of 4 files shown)


components/views/chat/chatbar/Chatbar.vue
```diff
@@ -183,6 +183,16 @@ export default Vue.extend({
         }
       }
     },
+    /**
+     * @method clearTyping
+     * @description Clears typing event immediatly
```
stavares843, line 188: immediately

ThomBos: > immediately

i go correct immediatly, sorry 🤣

stavares843: 🤣

WanderingHogan (changes requested):
This is very clean and a great attempt, but i dont think it solves the issue. I think we need to look at where this.debounceTypingStop(this) should go, so on the receiving side it tries to end the typing message if it doesn't hear anything back because the person you are chatting with closed the chat in the middle of typing

ThomBos:
> This is very clean and a great attempt, but i dont think it solves the issue. I think we need to look at where this.debounceTypingStop(this) should go, so on the receiving side it tries to end the typing message if it doesn't hear anything back because the person you are chatting with closed the chat in the middle of typing
Trying doing it this way! 👍 
By the way the main problem is when CTRL or META key are pressed... seems like the tab get detached and it doesn't receive any state update. Since we emit the chat event from the one who texts, we don't get the update and it stays "is typing".

ThomBos:
> This is very clean and a great attempt, but i dont think it solves the issue. I think we need to look at where this.debounceTypingStop(this) should go, so on the receiving side it tries to end the typing message if it doesn't hear anything back because the person you are chatting with closed the chat in the middle of typing

What if i add a setTimeout (which calls the debounce function) to the typing function?

iltumio:
we should reset the typing state even when we receive a disconnection event. On the recipient side we should automatically remove the typing notification if we do not receive a typing event within a certain amount of time

ThomBos:
> we should reset the typing state even when we receive a disconnection event. On the recipient side we should automatically remove the typing notification if we do not receive a typing event within a certain amount of time

that's exactly what this modification does (after 3s the typing state is cleared), but it can't be done from the recipient since we manage the typing state from the sender

phillsatellite (approved):

ThomBos:
Need to wait for Hogan's update for the peer connections to merge this PR

stavares843:
thanks for the update @ThomBos 🔨

ThomBos:
@stavares843 can you please check if there're still conflict with cypress? 
By the way, i started back from dev and applied new update following the Hogan's one

WanderingHogan (approved):

phillsatellite:
Tested again after rebase and it seems like if the user types very fast, it'll stop recognizing that they are typing and will just say connected until the user sends message and starts typing a new one 

https://user-images.githubusercontent.com/93608357/157933802-9daa3b1b-a083-43a1-90e3-6d419303c3aa.mov

ThomBos:
> Tested again after rebase and it seems like if the user types very fast, it'll stop recognizing that they are typing and will just say connected until the user sends message and starts typing a new one
> 
>  fast.typing.mov

i don't think that is't possible to have the perfect solution for this with the webrtc state we're currently using... 
The only thing i can do, is increasing the timer that turns off the typing state and decreasing the debounce function that sets off the typing state from sender's side (for updating more times the state while typing and having a more precise state).

Actual settings are:

sender                  receiver
0.5s                       5s   (inactivity.... or too much activity, like this case)

The main problem is that the sender doesn't update his state when is typing too fast and the receiver starts the inactivity timer if doesn't detect state changes (this is because refreshing page and disconnections are slow to detect... about 1 minute).

If you have any suggestion, i'd like to read that!! Anything that can improve this code! 💥

components/views/chat/chatbar/Chatbar.vue
```diff
@@ -221,22 +185,18 @@ export default Vue.extend({
...
+     * @method throttleTyping
+     * @description Throttles the typing event so that we only send the typing once every two seconds
      */
-    debounceTypingStop: debounce(function (ctx) {
-      ctx.$data.typing = false
-      ctx.typingNotifHandler(PropCommonEnum.NOT_TYPING)
-    }, 500),
+    throttleTyping: throttle(function (ctx) {
+      ctx.typingNotifHandler(PropCommonEnum.TYPING)
+    }, 2000),
```
molimauro, line 193: I'd put this `2000` in config

store/webrtc/actions.ts
```diff
@@ -85,13 +86,23 @@ const webRTCActions = {
...
       )
-      dispatch('subscribeToMailbox')
+      clearTimeout(prevTimeout)
+      prevTimeout = setTimeout(() => {
+        commit(
+          'friends/setTyping',
+          { id: peerId, typingState: PropCommonEnum.NOT_TYPING },
+          { root: true },
+        )
+      }, 2000)
```
molimauro, line 104: And then multiply here the config const * 3. Normally triple time is a good sign that connection dropped

stavares843 (changes requested):
https://user-images.githubusercontent.com/29093946/158424764-86533ba4-5936-4a64-a5ec-8b3020d56af9.mov

user A writes to user B
user A stops writing
still shows on user B for couple seconds that user A is typing but it isn't

molimauro:
@stavares843 we send a "is_typing" event once every 2 seconds and if within triple this time (so 6 seconds of course) we did not receive another event we assume the communication stopped, thats why is not immediate. We can lower the 6 seconds threshold its just a matter of preference. Sorry for not having specified this before

stavares843:
thanks for clarifying @molimauro 🔨

ThomBos:
what should i change here?

stavares843 (approved):

PR #1741, opened 2022-02-28
update(fileClose): update emit event rather than pass function as a prop

josephmcg:
**What this PR does** 📖
title
- while passing function as a prop may be convenient, I believe emitting from the child is best practice

N/a

- this event closes the file preview fullscreen onclick of the x. Maybe we could add another ticket, also closing on esc key

image

**Additional comments** 🎤


josephmcg:
We may want to update the PR title checker to include all of our commit types

molimauro (approved):

tooshel (approved):

stavares843 (approved):

PR #1742, opened 2022-02-28
feat(files): file type support for common image, archive types

josephmcg:
🔴 **this will remove any current filesystem data since import is based on a valid type**

**What this PR does** 📖
- type support for common file types
- switch type check on import to support more dynamic enums
  - previous check was only checking the key, rather than value. not scalable due to MIME types structure
- add basic file types to enum

AP-906

- this will NOT work retroactively. type is registered upon upload

**Additional comments** 🎤

### DIFF
--- a/libraries/Files/FilSystem.ts
+++ b/libraries/Files/FilSystem.ts
@@ -166,13 +166,13 @@ export class FilSystem {
    * @description recursively adds files and directories from JSON export
    */
   public importChildren(item: ExportItem) {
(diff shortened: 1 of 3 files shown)


molimauro (approved):

stavares843 (changes requested):
this branch

image

dev.satellite.one

image

preview is not showing for .PNG images but works on dev

stavares843 (approved):

PR #1753, opened 2022-03-01
feat(files): add support for last modified timestamp

josephmcg:
**What this PR does** 📖
- store modified in Item class
- dynamically set modified for a Directory based on latest modified value of a child
- update modified on like/share
- refresh every minute on list view

AP-907

**Additional comments** 🎤

### DIFF
--- a/components/views/files/list/List.vue
+++ b/components/views/files/list/List.vue
@@ -32,6 +32,16 @@ export default Vue.extend({
       type: Array as PropType<Array<Item>>,
       required: true,
     },
+    /**
+     * counter to force reactivity for Map
+     */
+    counter: {
+      type: Number,
+      required: true,
+    },
+  },
+  created() {
+    setInterval(this.forceRender, this.$Config.chat.timestampUpdateInterval)
   },
   methods: {
     /**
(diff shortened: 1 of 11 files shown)


stavares843:
the /rebase will not work due to the conflict

josephmcg:
never tried it before lol I was wondering what the issue was

stavares843:
works as long there's no conflict 😂 and sorry about the cypress conflict

components/views/files/list/List.vue
```diff
@@ -30,8 +30,18 @@ export default Vue.extend({
...
+    /**
+     * counter to force reactivity for Map
+     */
+    counter: {
+      type: Number,
+      required: true,
+    },
+  },
+  created() {
+    setInterval(this.forceRender, this.$Config.chat.timestampUpdateInterval)
```
molimauro, line 44: Can we save the interval and then clear it upon component destruction?

Jekrimo (approved):

components/views/files/list/List.vue
josephmcg: Sure, could you explain the benefits of doing so? I don't understand

stavares843:
is possible to have invalid date here? @josephmcg 

i haven't saw any invalid date but i was wondering if we have code to prevent that

also, does this appear with the specific date when is more than 24h?

minor questions, merging

stavares843 (approved):

PR #1765, opened 2022-03-02
fix(files): svg preview

josephmcg:
**What this PR does** 📖
- svg preview in grid view and full screen file view
- added utility function to convert a textile svg url to a usable data string

AP-970

this was more complicated than expected.. the more 'anonymous' url for an svg upload to a bucket looks like this
https://hub.textile.io/ipfs/bafkreifzm3k7qg3trv3qyfsxvyli5engm42if3vrezhr36oicxberyv4oa

This link can't be used as an `` src, I needed to fetch the svg then run some string replacement so it could be used as a data string for src.

We don't run into this issue in chat. Those urls maintain their file extension (textile.io/.../test.svg), but you lose anonymity. Anyone with that link can backtrack and see all the files in the bucket

I also experimented using the link as a data attribute for `<object>`, but the styling was painful and would be super hacky to fix (like direct DOM manipulation)

**Additional comments** 🎤

### DIFF
--- a/utilities/EncodeSvg.ts
+++ b/utilities/EncodeSvg.ts
@@ -1,0 +1,26 @@
+/**
+ * @function encode
+ * @description fetches svg from textile and converts to a data string
+ */
+export default async function encode(url: string): Promise<string> {
+  const svg = await fetch(url).then((res) => res.text())
+
+  return (
+    'data:image/svg+xml,' +
+    svg
+      .replace(
+        '<svg',
+        ~svg.indexOf('xmlns')
+          ? '<svg'
+          : '<svg xmlns="http://www.w3.org/2000/svg"',
+      )
+      .replace(/"/g, "'")
+      .replace(/%/g, '%25')
+      .replace(/#/g, '%23')
+      .replace(/{/g, '%7B')
+      .replace(/}/g, '%7D')
+      .replace(/</g, '%3C')
+      .replace(/>/g, '%3E')
+      .replace(/\s+/g, ' ')
+  )
+}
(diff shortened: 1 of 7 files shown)


phillsatellite (approved):

stavares843:
Captura de ecrã 2022-03-02, às 16 54 07

when downloading shows like the above
https://hub.textile.io/ipfs/bafkreie6ysqw5cwzlvjzduonvdkzfa24c5vptx3df5pmveq4yoq6ip37qy


instead of 

Captura de ecrã 2022-03-02, às 16 54 25


same image

should we ditch the download option for SVG's?

stavares843 (approved):

Jekrimo (approved):

josephmcg:
That's the content of the file. If you open an svg in a text editor, you will see the markup for it

stavares843:
and should we still have a download button for these?

josephmcg:
I would say yes, a user probably wants to download their uploaded file at some point

utilities/EncodeSvg.ts
```diff
@@ -0,0 +1,26 @@
...
+      )
+      .replace(/"/g, "'")
+      .replace(/%/g, '%25')
+      .replace(/#/g, '%23')
+      .replace(/{/g, '%7B')
+      .replace(/}/g, '%7D')
+      .replace(/</g, '%3C')
+      .replace(/>/g, '%3E')
+      .replace(/\s+/g, ' ')
+  )
```
molimauro, line 25: How about this: `return `data:image/svg+xml,${encodeURIComponent(svg)}``? :)

molimauro: with backticks of course

josephmcg: brilliant, thank you!

josephmcg:
Closing this due to upcoming file system changes regarding encryption. This would create a few conflicts for me. I'll make sure svgs work in my new encrypted branch

PR #1769, opened 2022-03-02
fix(media): close call on tab reload

ThomBos:
**What this PR does** 📖
Closes the call after reloading the tab
AP-802

**Additional comments** 🎤

### DIFF
--- a/components/views/media/Media.vue
+++ b/components/views/media/Media.vue
@@ -37,17 +37,6 @@ export default Vue.extend({
   computed: {
     ...mapState(['ui', 'accounts', 'friends', 'webrtc']),
     isActiveCall() {
-      const peer = this.$WebRTC.getPeer(this.webrtc.activeCall)
-      const remoteTracksLength =
-        peer && Object.keys(peer?.call?.tracksManager?.tracks).length
-
-      if (peer && remoteTracksLength === 0) {
-        peer?.call.hangUp()
-        this.$store.dispatch('webrtc/hangUp')
-        this.$store.commit('ui/fullscreen', false)
-        return
-      }
-
       return this.friends.all.find(
         (friend: any) =>
           friend.activeChat && friend.address === this.webrtc.activeCall,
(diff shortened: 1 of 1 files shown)


Jekrimo (approved):

phillsatellite:
So is the expected behavior supposed to be call is reconnected when user reloads or call should close out? When I tested I closed tab and refresh  and was brought back into video call I was in but video feed wasn't working (on other users side video feed was still working but I was frozen on they're side once I refreshed) and other user would get kicked out of video chat moments after

https://user-images.githubusercontent.com/93608357/156438689-ed8453e1-8138-424a-8d8c-89fd654b5e59.mov

stavares843:
also, @ThomBos is this expected to fix the red dot? 🤣

ThomBos:
> So is the expected behavior supposed to be call is reconnected when user reloads or call should close out? When I tested I closed tab and refresh and was brought back into video call I was in but video feed wasn't working (on other users side video feed was still working but I was frozen on they're side once I refreshed) and other user would get kicked out of video chat moments after
> 
>  clsoe.tab.2.mov

Oh.... if this video is recorded on this branch it's not working as it is supposed to. 
It should close the call for both participants when one get disconnected or refreshes the page 😢

ThomBos:
> also, @ThomBos is this expected to fix the red dot? 🤣

i mean, the red dot is more like a feature... we'd probably lose many users if we remove that

stavares843:
any news? 🔨

ThomBos:
> any news? 🔨

i'm closing another pr, probably going to work on this one, tomorrow late afternoon

phillsatellite:
Hi Thomas! I tested again and this time frozen video feed didnt pop up 👍  but it seemed like it took a while for the call to hangup after refreshing the page, but if the user who hung up was to type their pin in and login the call would end instantly on user B's side. Should the call end as soon as the PIN page is loaded or when the user has entered back into account

https://user-images.githubusercontent.com/93608357/157968072-89fe85c5-9c21-492a-b9ec-129270002227.mov

?

ThomBos:
> Hi @ThomBos ! I tested again and this time frozen video feed didnt pop up 👍 but it seemed like it took a while for the call to hangup after refreshing the page, but if User A (the user to refresh they're page) was to type their pin in and login the call would end instantly on user B's side. Should the call end as soon as the PIN page is loaded or when the user has entered back into account
> 
>  calling.mov 
> ?

Hi @phillsatellite and thanks for your feedbacks! The call should end as soon as possible, when a disconnection event is detected from the user whose still in call. So the main problem here is that we don't get the timeout event.
Taking your example (user A calls B) and making A crash, B needs to wait the time you saw.
If A reloads page, there's not time to wait if he has stored the pin or if he logs back in fast, because we take the event and close the opened call.

phillsatellite (approved):

stavares843:
this was using an old cypress version, I'm updating now, should be green next time
 
@ThomBos

stavares843:
Captura de ecrã 2022-03-21, às 15 56 22

all green now 🎉

PR #1774, opened 2022-03-02
fix(contextMenu): fix contextMenu error when clicking out of bounds

aewing:
**What this PR does** 📖
Fixes an error with the context menu when clicking outside of the viewport

Not sure if there is a specific issue for this, but I fixed this while testing [AP-973](https://satellite-im.atlassian.net/browse/AP-973)

I was seeing an error in the console when clicking on dev tools or clicking back into my code editor and this change resolved it
**Additional comments** 🎤

### DIFF
--- a/components/interactables/ContextMenu/ContextMenu.vue
+++ b/components/interactables/ContextMenu/ContextMenu.vue
@@ -56,6 +56,7 @@ export default Vue.extend({
      */
     handleOverflow() {
       const contextMenu = this.$refs.contextMenu as HTMLElement
+      if (!contextMenu) return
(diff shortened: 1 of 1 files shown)


stavares843:
do you know what specific error was on the console? @aewing to see if we had a ticket for it

aewing:
`contextMenu is undefined`
context-undefined

aewing:
And it looks like this can be reproduced by right clicking in most places in the application, so this may not be a great solution long term.

WanderingHogan (approved):

stavares843 (approved):

PR #1790, opened 2022-03-04
feat(files): encrypted bucket for personal files

josephmcg:
🔴 Breaking change - need to open a new account or files won't be encrypted

**What this PR does** 📖
- encrypt bucket
- store thumbnails of image files as base64 in files index
  - This allows image previews without downloading the whole image
  - images are scaled down using skaler package before conversion to base64
- download file and store as File in fileSystem class on more fullscreen view. show loader until dl is ready
- fix issue where a user could change folders mid upload, causing bugs

AP-984, AP-955, AP-970

- appreciate everyone's feedback, I was able to make the whole process much smoother
- disabled sharing functionality until I figure out how to handle that. It seems like we probably have to move files to an unencrypted bucket to share them
- temporarily removing FileSystem tests after adding skaler. jest doesn't play well with canvas, unsure of how to fix it
- we won't use things like ipfs hash for encrypted buckets, but we may for unencrypted. I left it for now, may remove later

**Additional comments** 🎤


libraries/Files/FilSystem.ts
```diff
@@ -179,10 +189,10 @@ export class FilSystem {
    * @param {FileSystemExport} fs
    * @description sets global file system based on parameter. will be fetched from Bucket
    */
-  public import(fs: FileSystemExport) {
-    fs.content.forEach((item) => {
-      this.importChildren(item)
-    })
+  public async import(fs: FileSystemExport) {
+    for (const item of fs.content) {
+      await this.importChildren(item)
```
molimauro, line 194: I was wondering if with a `Promise.all/Promise.allSettled` we can improve this importing. Or the import should happen sequentially?

Jekrimo (approved):

WanderingHogan:
@josephmcg Can you generate thumnails, and only pull/decrypt those on a per folder basis so the slow feeling for the user is reduced as much as possible, and don't decrypt any full size files until the user goes to download, or, in the future, share them?

libraries/Files/FilSystem.ts
josephmcg: that would speed things up a lot. it's currently sequential so that all the files were added as children in the correct folder. It could probably be reworked to improve speed

josephmcg:
@WanderingHogan Also a good idea. I'll play around with it today

libraries/Files/remote/textile/Bucket.ts
```diff
@@ -57,22 +43,24 @@ export class Bucket extends RFM implements RFMInterface {
...
-
-      this._index = await fetch(Config.textile.browser + hash).then((res) =>
-        res.json(),
+      )) {
+        data.push(bytes)
+      }
+      this._index = JSON.parse(
+        await new Blob(data, {
+          type: 'application/json',
+        }).text(),
```
josephmcg, line 63: let me know if there's a better way to do this. AsyncIterableIterators are painful to deal with
https://textileio.github.io/js-textile/docs/hub.buckets.pullpath

libraries/Files/TextileFileSystem.ts
```diff
@@ -38,4 +45,52 @@ export class TextileFileSystem extends FilSystem {
...
+      return
+    }
+    const buffer = new Uint8Array(await file.arrayBuffer())
+    if (isHeic(buffer)) {
+      const outputBuffer = await convert({
+        buffer,
+        format: 'JPEG',
+        quality: 1,
+      })
+      return await this._fileToData(
```
molimauro, line 70: `return await` is redundant, you can only `return` :)

libraries/Files/TextileFileSystem.ts
```diff
@@ -38,4 +45,52 @@ export class TextileFileSystem extends FilSystem {
...
+        await skaler(
+          new File([outputBuffer.buffer], file.name, {
+            type: 'image/jpeg',
+          }),
+          { width: 400 },
+        ),
+      )
+    }
+    return await this._fileToData(await skaler(file, { width: 400 }))
```
molimauro, line 79: same

phillsatellite:
Tested:From what I've checked so far the only thumbnail that appeared were .png and .webp, is that expected? Others I tested were .heic, .mov, .bmp, .svg



https://user-images.githubusercontent.com/93608357/157350772-89516d1b-63f3-4440-800d-597be52cbd92.mov

josephmcg:
Thanks Phill,

I made a mistake on my guard clause, so svg and heic were both broken. should be better now

I've had issues with that svg in particular as well. I think there's some sort of encoding issue in the header. I adjusted a line in the file then in started working. shared updated version below
I've tried other svgs as well (like the one below) and it's been working fine. 

card

Test

josephmcg:
at the moment, we're only supporting the [embeddable image types](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#supported_image_formats) as well as heic since everyone and their mother has an iphone.

We could certainly look into adding a thumbnail for videos, maybe even pdfs in a different ticket

phillsatellite:
Hi Joe! I tested again and found something kinda strange, so I uploaded a AVIF and the thumbnail/preview was a still picture, I think uploaded the same file to Dev.Satellite.one and the thumbnail and preview were both animated 

https://user-images.githubusercontent.com/93608357/157565341-92b05ebc-b5d7-4899-bf1f-8c73feb9e9bc.mov

josephmcg:
good catch, that seems like a minor issue we could address in a separate ticket. I'd love to get this merged so I can work on additional features

components/views/files/filepath/Filepath.less
```diff
@@ -12,17 +12,24 @@
     &:hover {
       &:extend(.font-accent);
     }
+    &.disabled {
```
WanderingHogan, line 15: we have a global disabled class in `assets/styles/base.less`

components/views/files/view/View.less
```diff
@@ -22,6 +22,9 @@
     flex-direction: column;
     text-align: center;
     margin: auto;
+    &.disabled {
```
WanderingHogan, line 25: same global disabled class note

WanderingHogan (approved):
good job Joe, nice to see this stuff coming together

phillsatellite (approved):

stavares843:
added 2 tickets for andre to take a look at the jest related things

[1061](https://satellite-im.atlassian.net/browse/AP-1061)
[1062](https://satellite-im.atlassian.net/browse/AP-1062)

libraries/Files/Fil.ts
```diff
@@ -96,6 +106,38 @@ export class Fil extends Item {
...
+   * @setter file
+   * @param {File} file file object
+   */
+  set file(file: File | undefined) {
+    this._file = file
+  }
+
+  /**
+   * @getter url
+   * @returns link of localally stored File for image preview and downloads
```
stavares843, line 127: localally - typo

stavares843 (approved):

PR #1795, opened 2022-03-04
fix(profile): fix close button issue on ios device

genie-magic:
**What this PR does** 📖
Close icon in profile is sitting in mobile status bar and causes user from not being able to close profile modal

AP-991

### DIFF
--- a/components/views/user/profile/Profile.less
+++ b/components/views/user/profile/Profile.less
@@ -94,7 +94,7 @@
     width: @full;
     height: @full;
     .close-button {
-      top: 2.5rem !important;
+      top: 2.5rem;
     }
     .profile-body {
       padding: @large-spacing;
@@ -125,7 +125,7 @@
   /* CSS specific to iOS devices */
   #profile {
     .close-button {
-      top: 2.5rem !important;
+      top: 2.5rem;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


Jekrimo (approved):

components/views/user/profile/Profile.less
```diff
@@ -93,7 +93,9 @@
   #profile {
     width: @full;
     height: @full;
-
+    .close-button {
+      top: 2.5rem !important;
```
iltumio, line 97: is there a way to avoid the use of `!important` directive?

components/views/user/profile/Profile.less
```diff
@@ -118,3 +120,12 @@
     }
   }
 }
+
+@supports (-webkit-touch-callout: none) {
+  /* CSS specific to iOS devices */
+  #profile {
+    .close-button {
+      top: 2.5rem !important;
```
iltumio, line 128: is there a way to avoid the use of `!important` directive?

iltumio (changes requested):

components/views/user/profile/Profile.less
genie-magic: @iltumio  I removed it.

components/views/user/profile/Profile.less
genie-magic: @iltumio  I removed it.

phillsatellite (approved):

stavares843 (approved):

PR #1801, opened 2022-03-04
fix(friend): fix undefined name issue

davidchan1219:
**What this PR does** 📖

Fixed the null name issue

AP-968


phillsatellite (approved):

WanderingHogan:
This PR does not address the ticket, but the change is fine. 

I cannot replicate the issue either, and it seems to be intermittent. @stavares843 If you want to keep the ticket open in the backlog, and un-assign it in case it comes back, or if you want to mark the ticket as closed and re-open if the issue resolves either is fine once this goes in.

WanderingHogan (approved):

PR #1812, opened 2022-03-08
fix(connectedPeers): fix issue showing connected peers

WanderingHogan:
**What this PR does** 📖
Before we were storing a single peer id in the state, so the last peer connection you set, if it was active, showed all of your connections as active. If that connection closed, it showed all of your dm's as not connected. Sorry the video is so long.

This update stores that data as an array. As connections are added or removed, that connection is added or removed from a list, so when you go to different users it shows the correct webrtc connection status, per user.

To test, on dev I created three users:

User A (regular Brave browser)
User B (incognito Brave Browser)
User C (regular chrome browser)

After this update you should see users appear online independent from each other.

Please use the local webtorrent when testing for the best results.

https://user-images.githubusercontent.com/2993032/157169478-8733616b-6aa0-4e37-be6b-5b6d8d9ffbc1.mov

No Ticket

**Additional comments** 🎤

### DIFF
--- a/store/webrtc/actions.ts
+++ b/store/webrtc/actions.ts
@@ -36,6 +36,12 @@ const webRTCActions = {
 
     commit('setInitialized', true)
   },
+  /**
+   * @method getActivePeers
+   * @description returns an array of the user id's that have an open/active webtorrent signal
+   * @example
+   * this.$store.dispatch('webrtc/getActivePeers') // ['userid1', 'userid2']
+   */
   async getActivePeers() {
     const $WebRTC: WebRTC = Vue.prototype.$WebRTC
     const peersArray: string[] = []
--- a/store/webrtc/actions.ts
+++ b/store/webrtc/actions.ts
@@ -45,7 +45,7 @@ const webRTCActions = {
   async getActivePeers() {
     const $WebRTC: WebRTC = Vue.prototype.$WebRTC
     const peersArray: string[] = []
-    if ($WebRTC.peers) {
+    if ($WebRTC.peers.size > 0) {
       $WebRTC.peers.forEach((e) => {
         if (Object.keys(e.communicationBus.instance.peers).length !== 0) {
           peersArray.push(e.identifier)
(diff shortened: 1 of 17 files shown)


store/webrtc/actions.ts
```diff
@@ -23,18 +23,38 @@ export default {
...
+   * @method getActivePeers
+   * @description returns an array of the user id's that have an open/active webtorrent signal
+   * @example
+   * this.$store.dispatch('webrtc/getActivePeers') // ['userid1', 'userid2']
+   */
+  async getActivePeers() {
+    const $WebRTC: WebRTC = Vue.prototype.$WebRTC
+    const peersArray: string[] = []
+    if ($WebRTC.peers) {
+      $WebRTC.peers.forEach((e) => {
```
josephmcg, line 49: WebRTC.peers is a Map right? that should always eval true, unless there was some initialization bug.
Maybe guard clause if `!$WebRTC.peers.size`, return empty array

josephmcg (commented):
I haven't worked with webrtc yet, so maybe I'm misinterpreting your intention with the `if` check

store/webrtc/actions.ts
WanderingHogan: thank you

josephmcg (approved):

molimauro:
Great! Just one thing, there is a control on `this.webrtc.connectedPeer` in `sidebar/live/Live.vue` that should be updated accordingly, other than that 💣

WanderingHogan:
Thanks Mauro & Joe, updates made

phillsatellite:
Tested: Had some issues here and there with it. When the Brave account refreshed, neither of the Chrome Accounts updated saying that Brave was offline, when User B refreshed the Brave account updated but User A did not. When User A refresh both User B and Brave updated fine. Sorry for the long video lmao

https://user-images.githubusercontent.com/93608357/157348566-13b3c183-159b-42ab-a22c-1e7abc4d3ccf.mov

WanderingHogan:
@phillsatellite To me it just looks like the connection isn't timing out right away. If you give it some more time the connection should timeout

phillsatellite:
Tested again, I was curious how long do you think it should take to update? Tried waiting a little longer but the biggest problem seems to be Chrome. Brave will update pretty quick. Github wont let me attach the video so Im gonna send it to your email @WanderingHogan

WanderingHogan:
@phillsatellite do you see the WebRTC Error in the console on chrome?

phillsatellite (approved):

PR #2283, opened 2022-03-10
fix(jest): resolved Skaler error import jest

drepram:
**What this PR does** 📖

Resolves Skaler import error by applying what was written on https://stackoverflow.com/questions/70829503/jest-gives-an-error-syntaxerror-unexpected-token-export-for-node-modules-mi

AP-1061

**Additional comments** 🎤

### DIFF
--- a/store/textile/actions.test.ts
+++ b/store/textile/actions.test.ts
@@ -1,32 +1,59 @@
 import Vue from 'vue'
 import { TextileError } from './types'
+import { DIRECTORY_TYPE } from '~/libraries/Files/types/directory'
 import * as actions from '~/store/textile/actions'
 import * as Logger from '~/utilities/Logger'
 import TextileManager from '~/libraries/Textile/TextileManager'
 import { TextileFileSystem } from '~/libraries/Files/TextileFileSystem'
 import { Config } from '~/config'
+import { Fil } from '~/libraries/Files/Fil'
+import { FilSystem } from '~/libraries/Files/FilSystem'
 
 Vue.prototype.$Config = Config
 
 const DefaultLogger = Logger.default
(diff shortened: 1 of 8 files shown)


store/textile/actions.test.ts
```diff
@@ -1,32 +1,59 @@
...
 
 describe('actions.default.initialize', () => {
   test('', async () => {
     try {
+      // Mocks so we have a fs.export
+      const mockFileData = {
+        name: 'TestFile.png',
+        hash: '0x0aef',
+        size: 42345,
+        descrption: 'Test file description',
```
stavares843, line 27: typo - descrption

stavares843 (approved):

stavares843 (changes requested):
Captura de ecrã 2022-03-10, às 19 39 54

phil saw this warning when compiling

doesn't happen on dev

can you take a look pls? 🔨

WanderingHogan:
You can import default exports as any name you want, so instead of `import { skaler as SkalerImported } from 'skaler'`, it would be `import SkalerImported from 'skaler'`, but why do you need to change this in the library?

josephmcg:
I pulled this down, reverted the import in TextileFileSystem back to the original and the tests were working.
`import skaler from 'skaler'`

Thanks for getting something working

stavares843 (approved):

stavares843:
updated

tests are green and no longer returning warning when compiling

josephmcg (approved):
not sure if Andre wants to work on the other test in this same PR. current stuff looks great! I can re-review later if needed

drepram:
Looks good to me. Can this be merged?

stavares843 (approved):

PR #2293, opened 2022-03-11
fix(nsfw): nsfw check valid files, allow unscannable/unembeddable files through

josephmcg:
**What this PR does** 📖
- only try nsfw scan if its a valid format(embeddable images or videos)
- fix false positives for TIFF and other un-embeddable images
- use async await in favor of promises

AP-807

- I know we typically avoid else, but it saves us from repetition in this case
- `const model = await nsfwjs.load()` is declared in both the if and else. If I placed it above, I would get weird errors

**Additional comments** 🎤

### DIFF
--- a/utilities/NSFW.ts
+++ b/utilities/NSFW.ts
@@ -5,7 +5,7 @@ import { FILE_TYPE } from '~/libraries/Files/types/file'
  * @method isNSFW
  * @description Checks if an image is NSFW using nsfwjs
  * @param {File} file File object to be scanned
- * @returns boolean of nsfw status
+ * @returns {Promise} nsfw status
  */
 export default async function isNSFW(file: File): Promise<boolean> {
   const vidTypes = [FILE_TYPE.MP4, FILE_TYPE.WEBM, FILE_TYPE.OGV]
--- a/utilities/NSFW.ts
+++ b/utilities/NSFW.ts
@@ -8,7 +8,7 @@ import { FILE_TYPE } from '~/libraries/Files/types/file'
  * @returns {Promise} nsfw status
  */
 export default async function isNSFW(file: File): Promise<boolean> {
-  const vidTypes = [FILE_TYPE.MP4, FILE_TYPE.WEBM, FILE_TYPE.OGV]
+  const vidTypes = [FILE_TYPE.MP4, FILE_TYPE.MOV, FILE_TYPE.WEBM, FILE_TYPE.OGV]
   const imgTypes = [
     FILE_TYPE.APNG,
     FILE_TYPE.AVIF,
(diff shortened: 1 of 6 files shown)


utilities/NSFW.ts
```diff
@@ -1,38 +1,49 @@
...
- * @returns Boolean based on predictionResults
- * @example
+ * @param {File} file File object to be scanned
+ * @returns {Promise} nsfw status
  */
-export const isNSFW = (file: File) => {
-  const fileTypePrefix = file.type.split('/')[0]
-  if (fileTypePrefix !== 'image') {
+export default async function isNSFW(file: File): Promise<boolean> {
+  const vidTypes = [FILE_TYPE.MP4, FILE_TYPE.WEBM, FILE_TYPE.OGV]
```
molimauro, line 11: Why we dont also scan, for example, for MOV videos? just try to understand :) thanks

molimauro:
@josephmcg is there already a ticket for hosting our own model in the future? https://github.com/infinitered/nsfwjs#host-your-own-model

utilities/NSFW.ts
josephmcg: good question, I couldn't find a comprehensive list of video types that can be embedded via the `<video>` tag at the time. After some additional research, I found this
https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs#common_codecs

josephmcg:
@molimauro not yet, that's a good idea. Seems like we can also save it to indexeddb if that would be helpful

utilities/NSFW.ts
josephmcg: I think files larger than 10mb skip the scan for now so the user doesn't run out of RAM.
Scaling down an image for a scan is easy, but I'm not sure what the solution is for large videos.

josephmcg:
the threshold should still be the same. If it wasn't a scannable image format like tiff, for instance, it would get allowed through. all tiff were getting marked as nsfw before because the scan was crashing

utilities/NSFW.ts
josephmcg: I didn't realize MOV was supported in browser. added that to the list to check for

josephmcg:
video scan was broken before. it was returning the same value for every video. Should be working now, but it will only scan videos smaller than 8mb, I'm going to work on that size limit in a separate ticket. if you need to remove that for testing - comment out these lines(components/views/files/controls/Controls.vue)
image

utilities/NSFW.ts
molimauro: Ok thank you very much!

molimauro (approved):

josephmcg:
Still blocked? I can check out nsfw file size limits after this

stavares843:
i added blocked because was checking something, is not related to any previous comments

stavares843 (approved):

PR #1676, opened 2022-02-23
fix(chat): multiple files upload

AlexZakablukov:
**What this PR does** 📖
posibility to upload a bulk of files

**Which issue(s) this PR fixes** 🔨
[AP-788](https://satellite-im.atlassian.net/browse/AP-788)

**Special notes for reviewers** 🗒️

**Additional comments** 🎤

PR #2303, opened 2022-03-11
fix(chat): focus on input after file upload

AlexZakablukov:
**What this PR does** 📖
focus on input after file upload

[AP-924](https://satellite-im.atlassian.net/browse/AP-924)

**Additional comments** 🎤

### DIFF
--- a/components/ui/Chat/Scroll/Scroll.vue
+++ b/components/ui/Chat/Scroll/Scroll.vue
@@ -2,8 +2,11 @@
 
 <script>
 import Vue from 'vue'
+import { mapState } from 'vuex'
 
 import { ChevronDownIcon } from 'satellite-lucide-icons'
+
+import { User } from '~/types/ui/user'
 
 export default Vue.extend({
   name: 'Scroll',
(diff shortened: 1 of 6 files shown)


WanderingHogan (approved):

phillsatellite:
Tested: Enter functionality works good, noticed that messages will not automatically move up in chat so user has to scroll after sending each message

https://user-images.githubusercontent.com/93608357/157915356-64c4678b-750f-4935-a57d-de4bb9ad61e1.mov

AlexZakablukov:
> Tested: Enter functionality works good, noticed that messages will not automatically move up in chat so user has to scroll after sending each message
> 
>  chat.mov

the same behavior on dev, i think this issue for another task

stavares843 (changes requested):
regarding the fix ''focus on input after file upload'' - is fixed on DM's but on groups still happens, do we need to add each fixed separately in groups as well going forward?

molimauro:
Should be ok once we merge #1676 ;)

stavares843:
https://github.com/Satellite-im/Core-PWA/pull/2326

PR #2313, opened 2022-03-14
feat(jest): 67% coverage for SearchIndex/index.ts

drepram:
**What this PR does** 📖

Self-explanatory. After some correspondence with Andrew, it seems that for now, we are not able to come up with a way to cover the remaining lines (consider too that it is prone to change).

No tickets IIRC.

**Additional comments** 🎤

### DIFF
--- a/libraries/SearchIndex/index.test.ts
+++ b/libraries/SearchIndex/index.test.ts
@@ -1,7 +1,8 @@
 import SearchIndex from './index'
+global.console.warn = jest.fn()
 
 describe('', () => {
-  test('', async () => {
+  test('update', async () => {
     const idx = new SearchIndex({
       ref: 'id',
       fields: ['text'],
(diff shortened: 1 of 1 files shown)


libraries/SearchIndex/index.test.ts
```diff
@@ -21,4 +22,94 @@ describe('', () => {
...
+      { id: '3', text: 'hello world' },
+    ]
+    idx.update(data)
+
+    expect(console.warn).toHaveBeenCalled()
+    expect(idx.search('$%@*&*#&@*', true)?.map((r) => r.ref)).toEqual([
+      '1',
+      '3',
+    ])
+    // expect(idx.search('world')?.map((r) => r.ref)).toEqual(['2', '3'])
```
stavares843, line 108: remove this commented out code unless this is WIP and you need this for later

if so please use /* instead of // in multiple lines

libraries/SearchIndex/index.test.ts
```diff
@@ -21,4 +22,94 @@ describe('', () => {
...
+    })
+    const data = [
+      { id: '1', text: 'hello' },
+      { id: '2', text: 'world' },
+      { id: '3', text: 'hello world' },
+    ]
+    idx.update(data)
+    const observable = jest.fn()
+    observable.subscribe = jest.fn()
+    // function callback(bool) {
```
stavares843, line 82: remove this commented out code unless this is WIP and you need this for later

if so please use /* instead of // in multiple lines

libraries/SearchIndex/index.test.ts
```diff
@@ -1,7 +1,8 @@
 import SearchIndex from './index'
+global.console.warn = jest.fn()
```
stavares843, line 2: do we need this warn?

stavares843 (changes requested):
minor comments

libraries/SearchIndex/index.test.ts
drepram: WIP because it's something that **might** be useful when work is done again on this file!

libraries/SearchIndex/index.test.ts
drepram: Hi, this is not needed. I've removed it!

libraries/SearchIndex/index.test.ts
drepram: We would, as we would later need to cover for warnings being logged. A WIP.

stavares843 (approved):

stavares843 (approved):

PR #2331, opened 2022-03-15
fix(images): heic type not always detected

molimauro:
**What this PR does** 📖
Fix HEIC images detection

AP-1077

**Additional comments** 🎤

### DIFF
--- a/components/views/files/controls/Controls.vue
+++ b/components/views/files/controls/Controls.vue
@@ -125,8 +125,8 @@ export default Vue.extend({
             return { file, nsfw: false }
           }
           // convert heic to jpg for scan. return original heic if sfw
-          const buffer = new Uint8Array(await file.arrayBuffer())
-          if (isHeic(buffer)) {
+          if (isHeic(file)) {
+            const buffer = new Uint8Array(await file.arrayBuffer())
             const outputBuffer = await convert({
               buffer,
               format: 'JPEG',
--- a/components/views/files/upload/Upload.vue
+++ b/components/views/files/upload/Upload.vue
@@ -112,9 +112,8 @@ export default Vue.extend({
         this.$data.count_error = false
         for (let i = 0; i < files.length; i++) {
           /* checking .heic file needs file array buffer because sometimes its file type return empty string */
-          const buffer = new Uint8Array(await files[i].arrayBuffer())
-          const isHeicType = isHeic(buffer)
-          if (isHeicType) {
+          if (isHeic(files[i])) {
+            const buffer = new Uint8Array(await files[i].arrayBuffer())
             /* convert .heic file to jpeg so that we can convert it for html5 style */
             const oBuffer = await converter({
               buffer, // the HEIC file buffer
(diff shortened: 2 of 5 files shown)


WanderingHogan (approved):

phillsatellite (approved):

stavares843 (approved):

stavares843:
thanks for updating the test as well 🔨

stavares843 (changes requested):
Captura de ecrã 2022-03-15, às 23 54 23

strange i can't upload a .heic file on this branch on files section, other files like jpeg works fine

works fine on dev

Captura de ecrã 2022-03-15, às 23 54 40

flow
- go to this branch
- files
- upload heic
- gets stuck on loading and never sents the image
- upload other file like jpeg
- file appears uploaded
- go to dev
- files
- upload heic
- file gets uploaded

josephmcg:
I noticed weird bugs when I was just checking the MIME type as well. I think @genie-magic setup the HEIC check based on byte values. Eric, could you explain why you did it that way?

genie-magic:
> I noticed weird bugs when I was just checking the MIME type as well. I think @genie-magic setup the HEIC check based on byte values. Eric, could you explain why you did it that way?
> 
> After alpha, I'd like to check byte value for all files rather than relying on MIME type (AP932)

Heic file is only able to detectable with byte check... not all heic file return type image/heic... some files return its type empty string.

josephmcg:
Filesystem is working fine on dev for me, but I see chat upload is a little buggy.
Probably related to this forEach in the async handleFile function. I got around this by using a `for await`, feel free to borrow that in the Upload.vue rework

image

image

molimauro:
> > I noticed weird bugs when I was just checking the MIME type as well. I think @genie-magic setup the HEIC check based on byte values. Eric, could you explain why you did it that way?
> > After alpha, I'd like to check byte value for all files rather than relying on MIME type (AP932)
> 
> Heic file is only able to detectable with byte check... not all heic file return type image/heic... some files return its type empty string.

Not always true unfortunately. Some HEIC where not properly detected only with byte value

josephmcg:
I wonder if the HEIC that aren't detected via byte value have been converted from another format. I would assume checking byte value is the most accurate way

genie-magic:
> I wonder if the HEIC that aren't detected via byte value have been converted from another format. I would assume checking byte value is the most accurate way

yes.. that's why I chose byte checking.

molimauro:
> I wonder if the HEIC that aren't detected via byte value have been converted from another format. I would assume checking byte value is the most accurate way

Its a photo i took with my iPhone and sent to my Mac via Airdrop

molimauro:
I forgot to update a function for the thumbnails but it seems that my heic image is not previewing correctly in the file section (maybe because its .HEIC instead of .heic?). This is the image that is not recognized as HEIC with the byte check: https://we.tl/t-PkP1zcbmvA

molimauro:
> Works well with genuine heic, but crashes on what I assume is a converted one. This is a file Phil sent me for testing https://we.tl/t-v6u0yMMoKx

On dev this is not recognized as valid heic right?

josephmcg:
correct, that evaluates false on dev isHeic check. Looking at the headers, it's a converted png
https://www.fileformat.info/tool/hexdump.htm

molimauro:
So i had a quick chat with Joe and we can stick to the current implementation based on bytes check. Still it is not perfect of course but i'll create a ticket for the future, we'll need to find a proper library to handle all these formats. Thanks everyone!

PR #2365, opened 2022-03-17
fix(files): responsive issue on files list view

josephmcg:
**What this PR does** 📖
- hide modified and file type columns if sidebar isn't collapsed on smaller desktops

AP-953

**Additional comments** 🎤


WanderingHogan (approved):
it would probably be better to name the class mobile-hide or something, so you know right off exactly what it does without looking at the stylesheet, but it's ok as is

josephmcg:
@WanderingHogan 
image

WanderingHogan (approved):

stavares843 (approved):

PR #2400, opened 2022-03-18
fix(nsfw): improve nsfw scan size constraints

josephmcg:
**What this PR does** 📖
- switch EnvInfo currentPlatform to use enums rather than string value
  - update all checks to be enum based
- do not scan if file exceeds canvas size limits based on [this](https://github.com/jhildenbiddle/canvas-size#test-results). app would previously hang trying to scale down an image thats too large
- if file is within canvas limits, but larger than 8mb, scale down using skaler pkg(canvas) before scan
- allow videos to be scanned regardless of size
  - tested up to ~100mb video, maybe larger ones would cause issues - please test 

AP-1066

- please try BIG pictures and videos (mp4, webm, mov, ogg)
- I updated our platform enum to match the capacitor doc. Not sure if we would ever get an 'electron' value for it, but I left those in to avoid bugs. https://capacitorjs.com/docs/v2/basics/utilities#getplatform

**Additional comments** 🎤
- Chat Upload.vue needs an update too, but that should be handled separately since it may cause unintended bugs
- I want users to be able to upload their large images in the file system, so I allow images through if theyre too large to be scanned
- I'm leaving the 8mb limit on profile picture upload so the scan doesnt mark them as safe. nsfw profile pics would be bad


stavares843 (changes requested):
Captura de ecrã 2022-03-22, às 03 49 12

i used a 2gb file video and crashed when uploading

STR:
- go to files
- upload a 2gb file

i used the 2gb one from this list

https://www.mediafire.com/folder/9971h5184vpju/Osmo_Pocket_test_files

from https://forum.dji.com/thread-183889-1-1.html

- crash

josephmcg:
Thanks, I never know where to find these massive files. I'll check it out tomorrow

josephmcg:
I wasn't able to check this for long today, but it seems to be arrayBuffer size limits during HEIC checks. I also need to apply canvas size limits to thumbnail generation

josephmcg:
@stavares843 ok! should be fixed now. It was running into memory issues taking a buffer of a 2gb file. I got around this by taking a buffer of the first 256 bytes for the heic check, rather than the full file. There may still be problems in chat file upload, but I believe those should be handled separately.

components/views/files/controls/Controls.vue
```diff
@@ -122,13 +122,13 @@ export default Vue.extend({
...
-          if (file.size > this.$Config.uploadByteLimit) {
-            return { file, nsfw: false }
-          }
           // convert heic to jpg for scan. return original heic if sfw
-          const buffer = new Uint8Array(await file.arrayBuffer())
-          if (isHeic(buffer)) {
+          if (isHeic(new Uint8Array(await file.slice(0, 256).arrayBuffer()))) {
+            // prevent crash in case of larger than 2GB heic files. could possibly be broken up into multiple buffers
+            if (file.size >= this.$Config.arrayBufferLimit) {
+              return { file, nsfw: false }
```
molimauro, line 129: so we have to accept heic nsfw files if > than the limit right?

phillsatellite:
Tested: I tried replicating Sara's crash and used a 3GB .mov File from the list she sent above. Was still able to replicate crash
Screen Shot 2022-03-24 at 2 01 42 PM

components/views/files/controls/Controls.vue
josephmcg: yes, we need the full file buffer in order to convert to heic. I was getting crashes trying to take a 2gb buffer. it's somewhat of an edge case due to the highly efficient nature of the heic format.

josephmcg:
that seems like a different error, but I'll look into it

components/views/files/controls/Controls.vue
josephmcg: https://chromium.googlesource.com/chromium/src/+/224e43ce1ca4f6ca6f2cd8d677b8684f4b7c2351/storage/browser/blob/README.md#blob-storage-limits

josephmcg:
@phillsatellite This seems to be a memory issue, handling such a large blob. Could we break this out into a separate issue?

stavares843:
yes agree @josephmcg 

or maybe we can also block upload with those sizes

tooshel:
> 

image

We need to change the front page of the site too.

stavares843:
thats not on this repo @tooshel 

but i can update that

stavares843:
what number do you want there?

WanderingHogan:
we can't just update the marketing without buy in from more people, we need to debug this to figure out what the limit is. It looks like this is a problem with nsfwjs or something else and not actually uploading

josephmcg:
my guess is chrome running out of memory since it's such a large blob. one possible solution is to use a stream instead, textile buckets support streams. however, I'm not sure if we could nsfw check a stream

josephmcg:
created some new utility functions, including one for HEIC checks using magic-bytes. I implemented it in the filesystem, but not chat Upload. I plan on changing Upload after this goes through

RetroPronghorn:
As the higher upload size is desirable enough to be one of the major things we're marketing I think it makes sense just to rip out NSFWjs. 

We shouldn't sacrifice quality for babysitting our users uploads. I think we should close this for now, and after alpha investigate comparing hashes against publicly available hash databases to detect nefarious uploads instead. (This was suggested to me by @WanderingHogan in DMs.) Ideally we really don't care to block NSFW content anyways, this was merely a crutch to get us to thinking up a better solution. Going to make the executive decision here for now and say rip out NSFW.js scanning and we'll work on a better solution over time.

josephmcg:
reopened after discussing with Matt since there's a lot of good stuff in here

RetroPronghorn (approved):

stavares843:
added ticket to
- handle file max size - [AP-1280](https://satellite-im.atlassian.net/browse/AP-1280)
- update chatbar when user tries to update corrupted files - [AP-1281](https://satellite-im.atlassian.net/browse/AP-1281)

stavares843:
thanks @josephmcg 🔨

stavares843 (approved):

PR #2401, opened 2022-03-18
fix(chat): fix unknown name issue when the user is registered with an avatar

genie-magic:
**What this PR does** 📖
fix unknown name issue when the user is registered with an avatar.

AP-920

### DIFF
--- a/store/accounts/actions.ts
+++ b/store/accounts/actions.ts
@@ -242,11 +242,12 @@ export default {
     }
 
     commit('setRegistrationStatus', RegistrationStatus.SENDING_TRANSACTION)
-
+    /* textilePubKey is generated first before setting account details if user is registered with avatar */
+    let preGeneratedTextilePubKey = null
     // only init textile if we need to push an image to bucket
     if (userData.image) {
       const { pin } = state
-      await dispatch(
+      preGeneratedTextilePubKey = await dispatch(
         'textile/initialize',
         {
           id: payerAccount?.publicKey.toBase58(),
@@ -273,7 +274,9 @@ export default {
       photoHash: imagePath,
       address: payerAccount.publicKey.toBase58(),
     })
-
+    /* reset textilePubKey after setting user detail if it is not set properly */
+    if (preGeneratedTextilePubKey && !state.details?.textilePubkey)
+      commit('updateTextilePubkey', preGeneratedTextilePubKey)
     dispatch('startup', payerAccount)
   },
   /**
--- a/components/views/friends/add/Add.vue
+++ b/components/views/friends/add/Add.vue
@@ -58,10 +58,9 @@ export default Vue.extend({
         this.error = this.$t('friends.self_add') as string
         return
       }
-
       if (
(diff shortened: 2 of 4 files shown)


phillsatellite (approved):

store/accounts/actions.ts
```diff
@@ -275,7 +276,9 @@ export default {
       photoHash: imagePath,
       address: payerAccount.publicKey.toBase58(),
     })
-
+    /* reset textilePubKey after setting user detail if it is not set properly */
+    if (preGeneratedTextilePubKey && !state.details?.textilePubkey)
```
iltumio, line 280: looks that `textile/initialize` action is already commiting the public key to the state. Can you check why it is not working instead of commiting the same key again?

iltumio (changes requested):

store/accounts/actions.ts
genie-magic: @iltumio  When no avatar, they first use setUserDetails to set user details and after that call 'textile/initialize' to set pubKey.... but when we register user info with avatar, we first set pubKey (without account details0 first and after that setUserDetails and pubKey is removed because setUserDetails reset value of account/details.

phillsatellite (approved):

store/accounts/actions.ts
iltumio: ok, let's investigate more later on

iltumio (approved):

phillsatellite:
When I went back to check something I noticed a crash on this branch. 
Browser: Safari
Flow: Select StorePin, select profile pic/Username, add a user - crash
Screen Shot 2022-03-30 at 1 39 26 PM

stavares843:
🔨

genie-magic:
> When I went back to check something I noticed a crash on this branch. Browser: Safari Flow: Select StorePin, select profile pic/Username, add a user - crash Screen Shot 2022-03-30 at 1 39 26 PM

@phillsatellite  Is it only happen for this branch?

stavares843:
yes phil tested on dev and was fine

stavares843:
are you able to reproduce the issue?

genie-magic:
> are you able to reproduce the issue?

not yet...  I want him to test again if he can.

phillsatellite:
@genie-magic Hi Eric! I just tested again and verified I could only replicate crash on this branch. I noticed some other details that I think causes a crash I setup one account with Safari and the other with Chrome
(Github won't let me attach the image because its heic so I'm gonna dm it to you)
Safari User:
1. Store Pin 
2. Add profile picture with image I attach and create username
3. Login and add the Chrome user
4. Crash

Chrome User:
1. Do not store pin
2. Do not add a profile picture

genie-magic:
@phillsatellite  I fixed that issue and pushed. can you check on your side again? thanks.

phillsatellite (approved):

WanderingHogan (approved):

stavares843 (approved):

PR #2424, opened 2022-03-20
chore(update): update swiper due to security alert

stavares843:
**What this PR does** 📖

Updates swiper lib to the latest version due to

Captura de ecrã 2022-03-20, às 19 47 52

was also needed to update the imports due to this error when building

image

now is fixed

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "simple-markdown": "^0.7.3",
     "simple-peer": "^9.11.1",
     "skaler": "^1.0.7",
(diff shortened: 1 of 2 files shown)


WanderingHogan:
I can't find the API docs or migration guides from swiper 5.x or swiper 6.x, but the most recent swiper (8) still has all the same options that we are using.

Jekrimo (approved):

phillsatellite (approved):

PR #2431, opened 2022-03-21
fix(chat): scroll on iOS

vimercati-samir:
What this PR does 📖
fix chat scroll on iOS

[AP-1008](https://satellite-im.atlassian.net/browse/AP-1008)

Additional comments 🎤

### DIFF
--- a/components/ui/Chat/Scroll/Scroll.less
+++ b/components/ui/Chat/Scroll/Scroll.less
@@ -14,6 +14,8 @@
 }
 
 .auto-scroll {
+  display: flex;
+  flex-direction: column;
   overflow-y: auto;
   overflow-x: hidden;
   height: 100%;
--- a/components/ui/Chat/Scroll/Scroll.less
+++ b/components/ui/Chat/Scroll/Scroll.less
@@ -14,8 +14,6 @@
 }
 
 .auto-scroll {
-  display: flex;
-  flex-direction: column-reverse;
   overflow-y: auto;
   overflow-x: hidden;
   height: 100%;
--- a/components/ui/Chat/Scroll/Scroll.less
+++ b/components/ui/Chat/Scroll/Scroll.less
@@ -42,6 +42,13 @@
     &:extend(.background-primary);
     &:extend(.round-corners);
   }
+
+  .chat-scroll-content {
+    display: flex;
+    flex: 1 1 auto;
+    justify-content: flex-end;
+    flex-direction: column;
+  }
 }
 
 .new-message-alert {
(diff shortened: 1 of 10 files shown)


genie-magic (approved):

stavares843 (changes requested):
image

i used the netlify URL on a real device, a iPhone 12 pro max and i can't scroll on the chat view

components/ui/Chat/Scroll/Scroll.less
```diff
@@ -14,8 +14,6 @@
 }
 
 .auto-scroll {
-  display: flex;
-  flex-direction: column-reverse;
```
molimauro, line 18: @vimercati-samir cannot remove these, otherwise the messages in the chat are not at the bottom (try with a new chat)

vimercati-samir: flex-direction column-reverse was breaking the scroll, should be fixed in commit a3787878f04bd733667ce8409c6198df4c2ad59e

vimercati-samir:
fixing scroll to bottom to be preemptive instead of waiting for the message to be sent and scroll to bottom not going all the way to the bottom

vimercati-samir:
> fixing scroll to bottom to be preemptive instead of waiting for the message to be sent and scroll to bottom not going all the way to the bottom

should be fixed in commit 2da93f4f61b3bb6dec6224a34963604ee0730379

molimauro:
If you dont store the pin you'll encounter problems in scrolling once the messages are loaded. @vimercati-samir can you fix it?

phillsatellite (approved):

molimauro:
@vimercati-samir conflicts then ok :)

molimauro (approved):

vimercati-samir:
> @vimercati-samir conflicts then ok :)

rebased in commit 54d441410be32de12ce61eb3889d73996ce70f1f

stavares843 (changes requested):
when you scroll up and add a reaction it scroll to the bottom - shouldn't happen

https://user-images.githubusercontent.com/29093946/159952622-856f9800-c88a-4b0e-b73e-d5ccffc5368c.mov

vimercati-samir:
> when you scroll up and add a reaction it scroll to the bottom - shouldn't happen
> 
>  Gravacao.do.ecra.2022-03-24.as.15.31.45.mov

Changed accordingly in commit be594758a1a5309defb462cb006ef439ae5b6961

molimauro:
Thanks @vimercati-samir, im doing some updates as well since i noticed some behaviour that needs to change. I'll let you all know once its done

josephmcg (approved):
🚀

stavares843 (approved):

PR #2444, opened 2022-03-22
fix(chat): block sending empty message on quick profile

genie-magic:
**What this PR does** 📖
block sending an empty message on the quick profile

AP-1101


josephmcg (changes requested):
try using an empty regex instead, I put one in the config.ts.
This implementation would allow a message of spaces

`if (this.$Config.regex.empty.test(this.text))`

genie-magic:
> try using an empty regex instead, I put one in the config.ts. This implementation would allow a message of spaces
> 
> `if (this.$Config.regex.empty.test(this.text))`

Thanks.

josephmcg:
We may want to show some type of error message, otherwise the user may be confused

genie-magic:
> We may want to show some type of error message, otherwise the user may be confused. maybe something like `Message must contain at least one non-space character`

I added that too. Thanks for your suggestion. we may need to add that to main chatbar too.

molimauro (approved):

phillsatellite (approved):

stavares843:
Captura de ecrã 2022-03-22, às 17 25 22

the error is not dismissed when user already entered a valid char

stavares843 (approved):

molimauro (approved):

PR #2445, opened 2022-03-22
feat(files): rename files and directories

josephmcg:
🔴 **You will need to add new files after switching to this branch. You won't be able to fetch the file on full screen view for old items**

**What this PR does** 📖
- new modal to rename items on right click. Works in both grid and list view
  - throw errors as expected if name already exists, etc
  - file system counts `a.txt` and `a.TXT` as different files. I made a new ticket for case sensitive duplicate name checks
- clean up old dummy rename emits/methods
- encourage user to keep extension in name by highlighting specific text, but they can remove if desired
  - on download, it will add the correct extension if the user removed/changed it
- removed hash property from Fil class. We were using it previously for unencrypted files.
- set file path in bucket based on item id (uuid). was based on name before which caused renaming issues
- Change renameChild to use a more typical guard clause
- update mobile half size modal css. Will only take up necessary room now, much better looking. Can be seen in glyph details (click a glyph in chat) and rename modal

AP-797

- if archive file extension is removed from name, it will only show regular icon for now. I made a new ticket for this
- Items are technically removed and added as a child, changing the order of items. This will be fixed later, with the `Recent` filter on the right

**Additional comments** 🎤


josephmcg:
there may be a bug if you take the following steps
1. upload 'text.txt'
2. rename to something else
3. upload another text.txt

Will check this tomorrow

molimauro:
Great job Joe. I found a bug when you open the preview of a file i cannot close it anymore, the buttons on top right are gone! Other than that we're good to go

josephmcg:
Thanks Mauro! I think that bug was introduced by my [PR the other day](https://github.com/Satellite-im/Core-PWA/pull/2364). Phil opened a ticket for it (AP1150)

molimauro:
Oh great then! Yeah could be related to the swiper library update, we had also other problems related to that in the chat/file view.

phillsatellite:
Tested: Works good but found a couple issues, 

1. I can save a rename a file to no name (or just spaces) and it will still create. If i file name is just ".png" and I try to rename it to that same name it will say "Already a file with this name"

https://user-images.githubusercontent.com/93608357/159762874-d1c146c5-ca55-4923-9639-deefa6f354f4.mov



2. On an actual iOS device when I click rename the screen will blur but no rename modal will appear

https://user-images.githubusercontent.com/93608357/159762880-9c0ce1be-5fb6-410b-b849-7be8dc3b0015.mov

josephmcg:
@phillsatellite The name issue should be resolved. It was initially setup to trim white space on names. I changed it to preserve whitespace (similar to how mac os handles it). However, mac will prevent you from naming something `.png`. I could try to address this separately if we don't like it.

Unsure of the modal issue, I added something that hopefully fixes it. waiting for netlify build to finish

josephmcg:
Yep, css fix worked on my iphone

phillsatellite:
Hi @josephmcg ! I tested again and iOS issue was fixed! But one little issue with it, the modal will appear at the bottom of the screen, I asked Liz and she had said "Yes it should appear in the middle. But not behind the keyboard. So like if the keyboard disrupts it or happens to sit on top of it, it should push it up" Lemme know if that clarified it enough if not feel free to DM me 😄 
ios

josephmcg:
the keyboard will more or less push it up to the middle. happy to make any additional edits if we receive clarification

stavares843:
Captura de ecrã 2022-03-25, às 20 02 20

i know this is already on https://github.com/Satellite-im/Core-PWA/blob/dev/libraries/Files/errors/Errors.ts#L9 but maybe we could edit for something more friendly

stavares843:
Captura de ecrã 2022-03-25, às 20 03 10

when user reaches the max char number doesn't return any feedback to the user

stavares843:
i can add a folder with a * but I can't rename with a *
Captura de ecrã 2022-03-25, às 20 18 43

Captura de ecrã 2022-03-25, às 20 18 37

stavares843:
rename is pretty neat 🎉 
Captura de ecrã 2022-03-25, às 20 30 25


im merging this, we can add those minor improvements I mentioned above separately

stavares843 (approved):

PR #2468, opened 2022-03-23
feat(jest): 100% coverage for libraries/SatelliteDB/SatelliteDB

drepram:
**What this PR does** 📖

Increase coverage

No tickets IIRC

**Additional comments** 🎤

### DIFF
--- a/libraries/SatelliteDB/SatelliteDB.test.ts
+++ b/libraries/SatelliteDB/SatelliteDB.test.ts
@@ -32,6 +32,10 @@ describe('SatelliteDB', () => {
     expect(db.search.conversationMessages).toBeInstanceOf(SearchIndex)
   })
 
+  test('reinitializing schema', async () => {
+    expect(await db.initializeSchema()).toEqual(db.initializeSchema())
+  })
+
   test('restoring search indexes', async () => {
     const where = { address: 'foo' }
     const data = { name: 'bar', textilePubkey: 'baz' }
@@ -45,6 +49,22 @@ describe('SatelliteDB', () => {
     expect(await db.search.friends.search('bar')?.[0]?.address).toEqual(
       where.address,
     )
+  })
+
+  test('reinitializing search indexes', async () => {
+    const where = { address: 'foo' }
+    const data = { name: 'bar', textilePubkey: 'baz' }
+    await db.initializeSearchIndexes()
+    await db.upsert('friends', where, data)
+    await db.saveSearchIndexes()
+    await db.close()
+
+    db = new SatelliteDB()
+    await db.initializeSearchIndexes()
+    expect(await db.search.friends.search('bar')?.[0]?.address).toEqual(
+      where.address,
+    )
+    expect(await db.initializeSearchIndexes()).toEqual(undefined)
   })
 
   test('upserting records', async () => {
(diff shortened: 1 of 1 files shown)


libraries/SatelliteDB/SatelliteDB.test.ts
```diff
@@ -32,6 +32,10 @@ describe('SatelliteDB', () => {
     expect(db.search.conversationMessages).toBeInstanceOf(SearchIndex)
   })
 
+  test('reinitializing schema', async () => {
+    expect(await db.initializeSchema()).toEqual(db.initializeSchema())
```
aewing, line 36: This calls the function, but doesn't actually assert that it did what it needed to

libraries/SatelliteDB/SatelliteDB.test.ts
```diff
@@ -47,6 +51,22 @@ describe('SatelliteDB', () => {
...
+    await db.upsert('friends', where, data)
+    await db.saveSearchIndexes()
+    await db.close()
+
+    db = new SatelliteDB()
+    await db.initializeSearchIndexes()
+    expect(await db.search.friends.search('bar')?.[0]?.address).toEqual(
+      where.address,
+    )
+    expect(await db.initializeSearchIndexes()).toEqual(undefined)
```
aewing, line 67: Again, it's always undefined, did it do what it needed to? Are we able to access the search index still after calling the method again?

libraries/SatelliteDB/SatelliteDB.test.ts
drepram: How could it assert what it needed to when it doesn't return anything?

aewing: Test the side effects (make sure the db tables are there after manually removing them). Same with search.

libraries/SatelliteDB/SatelliteDB.test.ts
drepram: From what I'm seeing, it seems that if it has been initialized once it would not be able to be accessed again because it would go into the `return ` block. Correct me if I am wrong!

aewing: The intent of this method isn't what it returns, but what it does. What it does is create the search indexes, so to test the full functionality we could initialize, initialize again and make sure we don't throw, manually delete indexes and initialize again, make sure they're there.

aewing: To that point the method could be altered to return some indication of whether or not the indexes were created, but short of that we would want to test for that outcome.

drepram: HI, I updated it. Please check again, thanks.

libraries/SatelliteDB/SatelliteDB.test.ts
drepram: HI, I updated it. Please check again, thanks.

aewing (approved):

libraries/SatelliteDB/SatelliteDB.test.ts
aewing: thanks for addressing that!

stavares843 (approved):

PR #2506, opened 2022-03-25
fix(user): styled as planned on figma

ThomBos:
**What this PR does** 📖
Restyles user's UI on sidebar

AP-1156

For testers, dates notification correct behaviour should be:

- "now" between 0 and 1 minute
- "hh:mm AM/PM" between 1 minutes and a day
- "yesterday" the day before
- "2 d" 2 days before 
- "MM/DD/YYYY" > 2 days before

***Updated with new format

**Additional comments** 🎤

### DIFF
--- a/components/views/user/User.vue
+++ b/components/views/user/User.vue
@@ -143,36 +143,23 @@ export default Vue.extend({
           return this.$t('friends.details.now')
         }
 
-        const sameDay =
-          this.$dayjs().format('DD') === this.$dayjs(uLastUpdate).format('DD')
+        const sameDay = this.$dayjs().isSame(uLastUpdate, 'day')
 
         if (sameDay) {
           return this.$dayjs(uLastUpdate).format('LT')
         }
 
-        if (!sameDay) {
-          const daysDif = this.$dayjs().diff(this.$dayjs(uLastUpdate), 'day')
+        const daysDif = this.$dayjs().diff(this.$dayjs(uLastUpdate), 'day')
 
-          if (daysDif <= 1) {
-            return this.$t('friends.details.yesterday')
-          }
-
-          if (daysDif > 1 && daysDif <= 2) {
-            return `${daysDif} + ${this.$t('friends.details.days_short')}`
-          }
-
-          const sameYear =
-            this.$dayjs().format('YYYY') ===
-            this.$dayjs(uLastUpdate).format('YYYY')
-
-          if (!sameYear) {
-            return this.$dayjs(uLastUpdate).format('L')
-          }
-
-          return this.$dayjs(uLastUpdate)
-            .format('L')
-            .replace(`/${this.$dayjs(uLastUpdate).format('YYYY')}`, '')
+        if (daysDif <= 1) {
+          return this.$t('friends.details.yesterday')
         }
+
+        if (daysDif > 1 && daysDif <= 2) {
+          return `${daysDif} + ${this.$t('friends.details.days_short')}`
+        }
+
+        return this.$dayjs(uLastUpdate).format('L')
       }
 
       return this.$t('friends.details.no_message')
--- a/components/views/user/User.vue
+++ b/components/views/user/User.vue
@@ -156,7 +156,7 @@ export default Vue.extend({
         }
 
         if (daysDif > 1 && daysDif <= 2) {
-          return `${daysDif} + ${this.$t('friends.details.days_short')}`
+          return `${daysDif} ${this.$t('friends.details.days_short')}`
         }
 
         return this.$dayjs(uLastUpdate).format('L')
--- a/components/views/user/User.vue
+++ b/components/views/user/User.vue
@@ -11,6 +11,11 @@ import { User } from '~/types/ui/user'
 import { Conversation } from '~/store/textile/types'
 import { Message, TextMessage } from '~/types/textile/mailbox'
 import { MessagingTypesEnum } from '~/libraries/Enums/enums'
+import { Config } from '~/config'
+import {
+  refreshTimestampInterval,
+  convertTimestampToDate,
+} from '~/utilities/Messaging'
 
 declare module 'vue/types/vue' {
   interface Vue {
@@ -54,6 +59,11 @@ export default Vue.extend({
       ],
       existConversation: false,
       isLoading: false,
+      timestamp: convertTimestampToDate(
+        this,
+        this.$store.state.textile.conversations[this.user.address]?.lastUpdate,
+      ),
+      timestampRefreshInterval: null,
     }
   },
   computed: {
@@ -76,10 +86,28 @@ export default Vue.extend({
     'textile.conversations': {
       handler(newValue) {
         this.existMessage(newValue)
+        this.$data.timestamp = convertTimestampToDate(
+          this,
+          newValue[this.user.address]?.lastUpdate,
+        )
       },
       deep: true,
       immediate: true,
     },
+  },
+  created() {
+    const setTimestamp = (timePassed: number) => {
+      this.$data.timestamp = convertTimestampToDate(this, timePassed)
+    }
+
+    this.$data.timestampRefreshInterval = refreshTimestampInterval(
+      this.$store.state.textile.conversations[this.user.address]?.lastUpdate,
+      setTimestamp,
+      Config.chat.timestampUpdateInterval,
+    )
+  },
+  beforeDestroy() {
+    clearInterval(this.$data.timestampRefreshInterval)
   },
   methods: {
     testFunc() {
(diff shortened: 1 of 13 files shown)


josephmcg:
Why aren't we formatting based on locale of the user? I set a locale value in the config.ts based on browser settings.
https://day.js.org/docs/en/i18n/i18n

RetroPronghorn:
agreed with @josephmcg, let's open a new ticket for changing to use Locales, once conflicts are resolved this should be good for QA.

ThomBos:
My bad with commits names.
@RetroPronghorn let me know what's the best way to proceed now, when you can 😄

components/views/user/User.vue
```diff
@@ -116,18 +121,60 @@ export default Vue.extend({
...
+          this.$dayjs(uLastUpdate),
+          'second',
+        )
+
+        if (secondsDif < 30) {
+          return this.$i18n.t('friends.details.now')
+        }
+
+        const sameDay =
+          this.$dayjs(today).format('DD') ===
```
molimauro, line 137: I dont think `today` is necessary, `this.$dayjs()` should be enough

components/views/user/User.vue
```diff
@@ -116,18 +121,60 @@ export default Vue.extend({
...
-          return this.$dayjs(uLastUpdate).format('HH:mm')
+        if (sameDay) {
+          return this.$dayjs(uLastUpdate).format('LT')
+        }
+
+        if (!sameDay) {
+          const daysDif = this.$dayjs().diff(this.$dayjs(uLastUpdate), 'day')
+
+          if (daysDif <= 1) {
+            return this.$i18n.t('friends.details.yesterday')
```
molimauro, line 148: `this.$i18n.t ` -> `this.$t `

components/views/user/User.vue
ThomBos: yes it's enough, i'm going to remove it

components/views/user/User.vue
```diff
@@ -116,18 +121,57 @@ export default Vue.extend({
...
+            return this.$t('friends.details.yesterday')
+          }
+
+          if (daysDif > 1 && daysDif <= 2) {
+            return `${daysDif} + ${this.$t('friends.details.days_short')}`
+          }
+
+          const sameYear =
+            this.$dayjs().format('YYYY') ===
+            this.$dayjs(uLastUpdate).format('YYYY')
```
molimauro, line 154: Same as above

components/views/user/User.vue
```diff
@@ -116,18 +121,57 @@ export default Vue.extend({
...
+        }
+
+        const sameDay =
+          this.$dayjs().format('DD') === this.$dayjs(uLastUpdate).format('DD')
+
+        if (sameDay) {
+          return this.$dayjs(uLastUpdate).format('LT')
+        }
+
+        if (!sameDay) {
```
molimauro, line 141: This if is not needed, at this point sameDay is false for sure

components/views/user/User.vue
```diff
@@ -116,18 +121,57 @@ export default Vue.extend({
...
+            this.$dayjs().format('YYYY') ===
+            this.$dayjs(uLastUpdate).format('YYYY')
+
+          if (!sameYear) {
+            return this.$dayjs(uLastUpdate).format('L')
+          }
+
+          return this.$dayjs(uLastUpdate)
+            .format('L')
+            .replace(`/${this.$dayjs(uLastUpdate).format('YYYY')}`, '')
```
molimauro, line 162: We cannot rely on this replace, this for example will break: https://github.com/iamkun/dayjs/blob/dev/src/locale/az.js
Since its not that important atm, lets keep the default L

ThomBos: > We cannot rely on this replace, this for example will break: https://github.com/iamkun/dayjs/blob/dev/src/locale/az.js Since its not that important atm, lets keep the default L


```

 const newFormatDate = this.$dayjs(uLastUpdate)
       .format('L')
       .replace(`${this.$dayjs(uLastUpdate).format('YYYY')}`, '')
 
const lastChar =  newFormatDate[newFormatDate.length - 1]

if(isNaN(Number(lastChar))){
      return newFormatDate
}

return newFormatDate.slice(0, -1)

```

Could this work in this case?  @molimauro

molimauro: https://github.com/iamkun/dayjs/blob/dev/src/locale/ug-cn.js

ThomBos: > https://github.com/iamkun/dayjs/blob/dev/src/locale/ug-cn.js

I've been checking all formats, a good solve could be clearing the string before and after YYYY.
There'd be just a few formats that'd not be displayed as we planned (the output would be like "DD. MM." because the use spaces 😮‍💨

molimauro (approved):

phillsatellite:
Tested: Should the timestamps automatically update or does the user have to navigate to a different page or send another message for it to update?

Message sent at 12:32
message 1232

Waited till 12:34 no update on time
message at 1234

stavares843:
let us know regarding phil's question above @ThomBos 🔨 

> Also is the PR supposed to have the entire UI match figma or is this mostly for the timestamps?

ThomBos:
> Tested: Should the timestamps automatically update or does the user have to navigate to a different page or send another message for it to update? Also is the PR supposed to have the entire UI match figma or is this mostly for the timestamps?
> 
> Message sent at 12:32 message 1232
> 
> Waited till 12:34 no update on time still says "now" message at 1234

it should stay "now" until the user leave the chat, but idk if it is good it stays potentially forever "now"... i'm going to talk with @molimauro to align the timing with the chat (atm aren't aligned, one says "now", the other one says "a minute ago")

josephmcg:
Sorry to drag this out, but I figured out something cool with the translations today that would help your `days_short`. you can pass params to translations. It will be helpful for number translations for languages with a different grammatical structure (Japanese comes to mind for me)

https://vue-i18n.intlify.dev/guide/essentials/syntax.html#named-interpolation

ThomBos:
> Sorry to drag this out, but I figured out something cool with the translations today that would help your `days_short`. you can pass params to translations. It will be helpful for number translations for languages with a different grammatical structure (Japanese comes to mind for me: `2 birds 1 stone` vs `一石二鳥`)
> 
> https://vue-i18n.intlify.dev/guide/essentials/syntax.html#named-interpolation
> 
> We can definitely move this to a separate ticket though and increase the scope of translations to update



> Tested: Should the timestamps automatically update or does the user have to navigate to a different page or send another message for it to update? Also is the PR supposed to have the entire UI match figma or is this mostly for the timestamps?
> 
> Message sent at 12:32 
> 
> Waited till 12:34 no update on time still says "now" 

Working on both... Anything should be done tomorrow morning, around 16h from now.

Actual status:
- aligned with chat messages (done)
- auto updates (missing a few things)
- nested translations (totally missing but should be fast)

ThomBos:
The only issue with this modification is a problem related to messages:
they get attached together when they are close in time and i can't detect when it happens, this makes, sometimes, display "now" on the sidebar and "HH:mm" in chat.

stavares843:
are you gonna add further work on this or is this ready to review and test? 🔨 @ThomBos

molimauro:
> Tested: Should the timestamps automatically update or does the user have to navigate to a different page or send another message for it to update? Also is the PR supposed to have the entire UI match figma or is this mostly for the timestamps?
> 
> Message sent at 12:32 message 1232
> 
> Waited till 12:34 no update on time still says "now" message at 1234

should match the Figma :)

utilities/Messaging.ts
```diff
@@ -267,15 +267,40 @@ export function getFullUserInfoFromState(
   return userInfo
 }
 
+export function convertTimestampToDate(context: any, timestamp: number) {
+  const secondsDif = context.$dayjs().diff(timestamp, 'second')
```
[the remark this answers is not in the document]
ThomBos: Can't remove context if we look for keeping interpolated translations such as line 291

phillsatellite:
Tested again and a new bug unfortunately appeared on iOS its kinda hard to describe so I'll try to be as detailed as possible 

User sends 1st message in chat, swipes to sidebar and waits and watches for timestamp to change (without touching anything)
When timestamp changes from "now" it will show the date the message was sent

User exits app to refresh page timestamp will now shows correct time when they are brought back into app

User sends another message in chat/goes to sidebar to watch timestamp change
When timestamp changes from "now" it will show the time that the original first message in chat was sent and not the most recent message

*feel free to DM me if you need me to explain more 

Screen Shot 2022-04-06 at 3 35 19 PM

ThomBos:
> Tested again and a new bug unfortunately appeared on iOS its kinda hard to describe so I'll try to be as detailed as possible (screenshot is from actual iOS device have to take a screenshot of photo I took from actual device cause github)
> 
> User sends 1st message in chat, swipes to sidebar and waits and watches for timestamp to change (without touching anything) When timestamp changes from "now" it will show the date the message was sent
> 
> User exits app to refresh page timestamp will now shows correct time when they are brought back into app
> 
> User sends another message in chat/goes to sidebar to watch timestamp change When timestamp changes from "now" it will show the time that the original first message in chat was sent and not the most recent message
> 
> *feel free to DM me if you need me to explain more
> 

Fixed, thanks for reporting

phillsatellite (approved):

molimauro (approved):

stavares843:
Captura de ecrã 2022-04-08, às 16 46 12

left side is dev, right side is this branch

on dev on chat says one hour ago, and on the sidebar shows the message at 15:35

then on this branch says on chat 3:27 PM and sidebar says 3:35 PM

the timestamps needs to match if we are gonna show exact timestamp in both places

stavares843:
Captura de ecrã 2022-04-08, às 16 48 11

this is the account in case that's useful for debug

ThomBos:
> left side is dev, right side is this branch
> 
> on dev on chat says one hour ago, and on the sidebar shows the message at 15:35
> 
> then on this branch says on chat 3:27 PM and sidebar says 3:35 PM
> 
> the timestamps needs to match if we are gonna show exact timestamp in both places

The only weird thing i see is that 15:27 (chat) and 15:35 (sidebar) when they should be aligned...

Was last reaction added at 15:35? 
If this is the case, probably the lastUpdate state get updated when it shouldn't; and something has to be changed on ```store/textile/mutations.ts ```

stavares843:
> Was last reaction added at 15:35?
If this is the case, probably the lastUpdate state gets updated when it shouldn't; and something has to be changed on store/textile/mutations.ts


probably, I can check if that was the case

but we should update then, if not gets confusing

ThomBos:
> > Was last reaction added at 15:35?
> > If this is the case, probably the lastUpdate state gets updated when it shouldn't; and something has to be changed on store/textile/mutations.ts
> 
> probably, I can check if that was the case
> 
> but we should update then, if not gets confusing

i don't think we should update when a reaction is applied to a message, or this "desync" will happen again

stavares843:
> i don't think we should update when a reaction is applied to a message, or this "desync" will happen again... also we don't show reactions timestamp


I agree


and yes was due to the reaction that was added on the message

stavares843:
after discussing with @molimauro we will address that in another ticket 🔨 thanks @ThomBos 🎉

ThomBos:
> after discussing with @molimauro we will address that in another ticket 🔨 thanks @ThomBos 🎉

no problem 😄

molimauro:
@ThomBos can you fix conflicts? thanks

stavares843:
no worries, thanks for fixing 🔨

stavares843 (approved):

stavares843:
missing fixing conflict 🔨

molimauro:
Done 🔨

PR #2512, opened 2022-03-25
feat(privacy): re-work how signaling servers are stored in config.ts, and allow the user to change their settings.

genie-magic:
**What this PR does** 📖
re-work how signaling servers are stored in config.ts, and allow the user to change their settings.

AP-1074

- See console log for checking signaling urls

### DIFF
--- a/components/views/settings/pages/privacy/Privacy.html
+++ b/components/views/settings/pages/privacy/Privacy.html
@@ -38,10 +38,19 @@
     <SettingsUnit
       :title="$t('pages.privacy.serverType.title')"
       :text="$t('pages.privacy.serverType.subtitle')"
+      :breakLine="false"
     >
+      <TypographyText
+        class="color-danger"
+        type="danger"
+        :size="6"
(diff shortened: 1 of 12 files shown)


WanderingHogan:
@genie-magic I don't think this is working as expected. Also after messing around with it, what about re-initializing instead of reloading?

tooshel:
> > But we should actually hold off on this change. @iltumio yesterday told me his webrtc rewrite will do away with the webtorrent signaling servers entirely using libp2p relays
> 
> You mean our all work for web torrents have no meaning @WanderingHogan ?

@genie-magic I wouldn't say "no meaning" . . . . even if something isn't merged in we still learned something.

genie-magic:
> @genie-magic I wouldn't say "no meaning" . . . . even if something isn't merged in we still learned something.

Yes.  You are right.

WanderingHogan:
@genie-magic I am going to close this but leave the branch up, and we can re-visit after we see what @iltumio and @Jekrimo get in for libp2p. Sorry, but good job on this

PR #2516, opened 2022-03-27
chore(Groups): combines groups and users, fixes several group UI bugs

RetroPronghorn:
**What this PR does** 📖
Fixes the top of groups not using the correct icon
Fixes the top of groups not having the the complete text
Fixes groups in sidebar using sidebar menu instead of correct clickable groups
Fixes groups sidebar being in separate area in favor of combining them like all other chat apps.
Fixes sorting users and groups in sidebar by most recently updated.

**Requires more work for some features outlined below:** 🎤

Needs last updated to be added to groups to properly sort. (Differing to someone with more experience here). Once added it will automatically work, the code is bootstrapped for it down to the state level.

Needs addresses tracked in groups instead of just `members: int` so that we can properly build out the user icons, 5 mins fix by changing the values in the GroupIcon component once a list of members in the groups addresses are avail. The icons will be generated by up to the first 5 members in the group, group icon can be customized to show the last person whom messaged biggest by ordering the members array.

Group status message needed to properly display the group status or subject. Added placeholder member count instead for now.


components/views/group/icon/GroupIcon.html
```diff
@@ -21,7 +21,5 @@
       <UiCircle type="random" :seed="address" :size="35" />
     </div>
   </div>
-  <span v-if="group.members.length > 3" class="label"
-    >{{group.members.length}}</span
-  >
+  <span v-if="group.membersh > 3" class="label">{{group.members}}</span>
```
molimauro, line 24: just a typo `group.membersh` -> `group.members`

molimauro:
Great! Thanks a lot Matt. Only one minor typo then we're ready for QA

stavares843:
Captura de ecrã 2022-03-27, às 19 38 13

is this on purpose? 

left side is this branch, right side is dev

theres no messages/groups tab or the option to create a group as well

RetroPronghorn:
@stavares843 we combined the two to make the UI cleaner. I can explain more if you'd like later. but the TLDR is attached to the top 

It moves creation of groups to the DMs with a user like in most other chat apps, it also combines the list of active chats and active group chats, and sorts them by the most recent messages. Some of this functionality for sorting isnt 100% yet but it just depends on us doing work on some other things first.

stavares843:
ohh got it

thanks for explaining @RetroPronghorn 🔨

molimauro (approved):

PR #2520, opened 2022-03-28
feat(chat): copy image

josephmcg:
**What this PR does** 📖
- copy image via clipboard API which only accepts png
- if png, copy. if other embeddable image format, convert to png with canvas
- change heic conversion in chat to png rather than jpeg. this skips a conversion step on copy

AP-1080

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/message/Message.vue
+++ b/components/views/chat/message/Message.vue
@@ -4,6 +4,7 @@ import Vue, { PropType } from 'vue'
 import { mapState } from 'vuex'
 
 import { ArchiveIcon } from 'satellite-lucide-icons'
+import { filetypemime } from 'magic-bytes.js'
 import ContextMenu from '~/components/mixins/UI/ContextMenu'
 import { Config } from '~/config'
 import { UIMessage, Group } from '~/types/messaging'
@@ -17,6 +18,8 @@ declare module 'vue/types/vue' {
     editMessage: () => void
     emojiReaction: (e: MouseEvent) => void
     copyMessage: () => void
+    copyImage: () => void
+    toPng: (blob: Blob) => Blob
   }
 }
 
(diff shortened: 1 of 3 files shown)


stavares843:
''if png, copy. if other embeddable image format, convert to png with canvas''

should we do this also for .gif's?

josephmcg:
@stavares843 I don't think we're able to. Clipboard API only supports png, so a gif can't be converted. Maybe in case of png, we should offer the 'copy link' option which hasn't been implemented yet

stavares843:
atm they don't crash they just copy into a .png instead of copying into .gif

josephmcg:
ah ok good. In that case.. Could we implement gifs in a separate PR along with the copy link feature?

stavares843:
Captura de ecrã 2022-03-28, às 09 53 20

maybe we should hide the ''copy image'' in case the image that the user sent is broken - for example sending .tiff (there's a ticket for fixing the .tiff)

stavares843:
> ah ok good. In that case.. Could we implement gifs in a separate PR along with the copy link feature?

agree 🔨

josephmcg:
tiffs aren't embeddable. we could possibly install a converter, but it would increase our bundle size. I don't think the payoff is worth it. I will switch tiff to a regular file embed rather than image in that ticket. Will add this info to ticket as well

stavares843:
> tiffs aren't embeddable. we could possibly install a converter, but it would increase our bundle size. I don't think the payoff is worth it. I will switch tiff to a regular file embed rather than image in that ticket. Will add this info to ticket as well

thanks 🔨

stavares843:
Captura de ecrã 2022-03-28, às 10 00 31

when sending an SVG, the left side is copy-pasted using copy image, right side original - less resolution


Captura de ecrã 2022-03-28, às 10 01 15

when sending an SVG, the bottom one is copy-pasted using copy image, the top one is original - small and less resolution

josephmcg:
it's copying the intrinsic dimensions of the file. Do we want to manually scale it up if it's below the rendered value of 300px wide?
image

stavares843:
sounds good 🔨

josephmcg:
Experimented with it, but couldn't figure out a clean implementation. I think copying the intrinsic dimensions is good functionality. most svg I come across have a defined width and height which get overriden with css

josephmcg:
closing this because its slow. I'm going to refactor and do a magic numbers check on upload rather than the copy step

PR #2525, opened 2022-03-28
feat(chat): implement vue-virtual-scroll-list for conversation groups

AlexZakablukov:
**What this PR does** 📖
implement vue-virtual-scroll-list for conversation groups

[AP-17](https://satellite-im.atlassian.net/browse/AP-17)

**Additional comments** 🎤


phillsatellite:
Tested: Works awesome on Desktop, I noticed when I tested on actual iOS device that it will cut off the text box for chat, and also it seems to take a lot longer to send messages compared to dev

https://user-images.githubusercontent.com/93608357/160485551-6707a5db-0988-403f-bf5b-0a4af4338191.mp4

phillsatellite (approved):

WanderingHogan:
This exposes an issue with the way our chat is structured.

I think the virtual scroll is too high up in the dom, and the component we have created for group might need to be re-thought. By having it in `components/views/chat/conversation/Conversation.html`, the `estimate-size` param on virtual scroll is saying that the group component could be up to that many groups. Even though we will only ever have the one 'group' class.

Instead of just adding the messages we need on screen, we still add all the messages to the dom: 
image

The problem will make more sense if you use the account seed in the ticket, and drill down into the messages:
image

WanderingHogan:
we might need to brainstorm what we should do here with @molimauro - re-working the chat view would be a big effort I think but might be worth the investment to get this to try to add way fewer things to the dom and feel faster

molimauro:
Yes @WanderingHogan thats why i put the "temporary blocked" label. As already said to Matt im planning a refactor on the chat that will involve many modifications (like pagination).

Just to clarify, here the virtualization is working only on groups of messages because, as you know, our chat is composed by putting these groups of messages one after the other. To me it would be much better to have a chat composed simply by a list of messages and this require a lot of updates of course.

tl;dr: better do some brainstorming

molimauro:
I'll close this one atm, thanks @AlexZakablukov i think it was worth the experiment.

PR #2539, opened 2022-03-29
fix(chat): check magic number to determine embed type

josephmcg:
**What this PR does** 📖
- create utility function to determine mime type via magic numbers
- fix tiff embed issue
- adjust mobile styles for file embed
- remove unused conditional render. all image messages are type 'file'
- minor unrelated issue of archive file checking. wanted to place utility function in this new file

AP-1181, AP-1155

- Going to refactor this after nsfw goes through. Right now it's byte checking on the receiving end. I want to byte check on the upload end instead

image

**Additional comments** 🎤


josephmcg:
going to revise this to read mime type in the message data, future PR should byte check outgoing messages

PR #2643, opened 2022-03-30
feat(chat): copy image

josephmcg:
**What this PR does** 📖
- reopened PR after some minor refactoring https://github.com/Satellite-im/Core-PWA/pull/2520
- copy image via clipboard API which only accepts png
- if png, copy. if other embeddable image format, convert to png with canvas
- change heic conversion in chat to png rather than jpeg. this skips a conversion step on copy
- switch ContextMenuItem definition so we don't need to do `as string` every time we use it

AP-1080

https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem#writing_to_clipboard
we need the blob to write to the clipboard. Initially I was fetching it during the copy method, but it felt very slow. Decided to fetch the blob on mounted instead. It will only convert to png one time as needed. We could also convert every image on mount instead, what do you think?

**Additional comments** 🎤

### DIFF
--- a/store/ui/types.ts
+++ b/store/ui/types.ts
@@ -153,7 +153,7 @@ export enum SettingsRoutes {
 // | 'realms'
 
 export interface ContextMenuItem {
-  text: string
+  text: string | TranslateResult
   func: Function
 }
 
--- a/components/views/chat/message/Message.vue
+++ b/components/views/chat/message/Message.vue
@@ -4,24 +4,12 @@ import Vue, { PropType } from 'vue'
 import { mapState } from 'vuex'
 
 import { ArchiveIcon } from 'satellite-lucide-icons'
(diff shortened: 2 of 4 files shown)


phillsatellite (approved):

store/ui/types.ts
```diff
@@ -153,7 +153,7 @@ export enum SettingsRoutes {
 // | 'realms'
 
 export interface ContextMenuItem {
-  text: string
+  text: string | TranslateResult
```
WanderingHogan, line 156: neat

WanderingHogan (approved):

stavares843:
added ticket related to SVG [here](https://satellite-im.atlassian.net/browse/AP-1289)
added ticket related to .GIF [here](https://satellite-im.atlassian.net/browse/AP-1290)

PR #2654, opened 2022-03-31
fix(files): rename quality of life update

josephmcg:
**What this PR does** 📖
- enforce invalid name rules on file/directory creation and rename
- refactor name validation to a separate method which can be called from the constructor
- move Directory name validation inside constructor rather than inside of Controls.vue
- allow files of the same name to be located in different directories. Rename update switch to uuid for the bucket identifier rather than name, so now it's allowed. Still preventing same name inside a directory

AP-1179

- I want to improve file Upload validation in a separate ticket. Right now I'm checking on the front end before attempting to upload. I would much rather move it to the constructor, but it would be fairly involved

**Additional comments** 🎤
- ticket mentioned feedback for max char limit in folder name input. That should probably be a global change for all Input/InputGroup. separate ticket

### DIFF
--- a/libraries/Files/abstracts/Item.abstract.ts
+++ b/libraries/Files/abstracts/Item.abstract.ts
@@ -4,6 +4,7 @@ import { FileSystemErrors } from '../errors/Errors'
 import { ItemInterface } from '../interface/Item.interface'
 import { DIRECTORY_TYPE } from '../types/directory'
 import { FILE_TYPE } from '../types/file'
+import { Config } from '~/config'
 
 export abstract class Item implements ItemInterface {
   private _id: string = ''
@@ -12,7 +13,7 @@ export abstract class Item implements ItemInterface {
   private _liked: boolean = false
   private _shared: boolean = false
   private _modified: number
-  abstract type: DIRECTORY_TYPE | FILE_TYPE
+  private _type: FILE_TYPE | DIRECTORY_TYPE
   abstract modified: number
   abstract size: number
 
@@ -29,6 +30,7 @@ export abstract class Item implements ItemInterface {
     shared,
     parent,
     modified,
+    type,
   }: {
     id?: string
     name: string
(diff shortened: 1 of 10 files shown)


libraries/Files/abstracts/Item.abstract.ts
```diff
@@ -29,23 +30,32 @@ export abstract class Item implements ItemInterface {
...
+
     this._id = id || uuidv4()
     this._name = name
     this._shared = shared || false
     this._liked = liked || false
     this._parent = parent || null
     this._modified = modified || Date.now()
+    this._type =
+      type ||
+      (this.constructor.name === 'Fil'
```
Jekrimo, line 56: Is this a typo for 'File'?

Jekrimo (changes requested):

libraries/Files/abstracts/Item.abstract.ts
stavares843: i believe this is related to https://github.com/Satellite-im/Core-PWA/blob/dev/libraries/Files/Fil.ts and we use ''Fil'' because ''File'' is already in use for typescript itself or similar

stavares843: but joe can better say

Jekrimo: You're totes right, I'mma close this, thank you!

Jekrimo (approved):

phillsatellite:
Tested: So when I create a new folder and go into it there will already be another folder named "un-named directory" in new folder. Is that intended
Screen Shot 2022-03-31 at 8 07 03 PM
?

josephmcg:
I'm not encountering that issue. can you try this with a new account? There may be some issues importing current filesystem structure with these changes

molimauro (approved):

josephmcg:
Thanks, that was something from the original filesystem setup that I missed on this round of updates. should be fixed now

phillsatellite (approved):

stavares843 (approved):

PR #2661, opened 2022-03-31
fix(toolbar): remove glitch effect on hover in safari

AlexZakablukov:
**What this PR does** 📖
remove glitch effect on hover in safari

[AP-1210](https://satellite-im.atlassian.net/browse/AP-1210)

**Additional comments** 🎤

### DIFF
--- a/components/ui/ComingSoon/ComingSoon.less
+++ b/components/ui/ComingSoon/ComingSoon.less
@@ -14,7 +14,6 @@
   }
   &.grayscaled {
     .circle{
-      -webkit-filter: grayscale(100%); /* Safari 6.0 - 9.0 */
       filter: grayscale(100%);
     }
   }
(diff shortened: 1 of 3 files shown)


components/ui/ComingSoon/ComingSoon.less
```diff
@@ -11,9 +12,11 @@
       justify-content: center;
     }
   }
-  .grayscaled {
-    -webkit-filter: grayscale(100%); /* Safari 6.0 - 9.0 */
-    filter: grayscale(100%);
+  &.grayscaled {
+    .circle{
+      -webkit-filter: grayscale(100%); /* Safari 6.0 - 9.0 */
```
stavares843, line 17: I know this was already there but do we need this? The last safari version is 15.0

I don't think we need to add support for such old versions

molimauro: Sara is right, we can remove it

AlexZakablukov: ok, i agree with you)

AlexZakablukov: fixed

stavares843 (approved):

PR #2664, opened 2022-04-01
feat(chat): save image

josephmcg:
**What this PR does** 📖
- save image functionality

- I tried to minimize the # of blobs we're storing in memory while keeping copy/save image fast

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/message/Message.vue
+++ b/components/views/chat/message/Message.vue
@@ -48,7 +48,9 @@ export default Vue.extend({
       disData: 'DataFromTheProperty',
       timestampRefreshInterval: null,
       timestamp: this.$dayjs(this.$props.message.at).fromNow(),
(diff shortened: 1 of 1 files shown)


molimauro (approved):

stavares843:
Captura de ecrã 2022-04-01, às 15 17 47

should we have the save image on broken images? example sending a corrupted file

also, if you send a corrupted image, that was a .txt originally, when saving, saves the .txt and not the .png - is that expected on chat due to the last changes on the files?

stavares843:
Captura de ecrã 2022-04-01, às 15 30 46

if you send a .psd file, send as a file, not an image preview

but we already have the download button there, so the save image seems redundant

stavares843:
minor questions above, merging 🔨

PR #2670, opened 2022-04-01
fix(register): throw error on unscannable file type and block upload

AlexZakablukov:
**What this PR does** 📖
throw error on unscannable file type and block upload

[AP-959](https://satellite-im.atlassian.net/browse/AP-959)

**Additional comments** 🎤

### DIFF
--- a/components/views/user/create/Create.vue
+++ b/components/views/user/create/Create.vue
@@ -3,6 +3,7 @@
 <script lang="ts">
 import Vue, { PropType } from 'vue'
 import { UserRegistrationData } from '~/types/ui/user'
+import { isEmbeddableImage } from '~/utilities/FileType'
 
 export default Vue.extend({
   name: 'CreateUser',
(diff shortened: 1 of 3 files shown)


josephmcg (changes requested):
please dont make the nsfw function throw an error. we want to allow uploads if the filesystem if theyre too large to be scanned or an un-scannable type.

could you do something like this instead? including the valid types.
`<input type="file" name="myImage" accept="image/png, image/gif, image/jpeg" />`

AlexZakablukov:
> please dont make the nsfw function throw an error. we want to allow uploads if the filesystem if theyre too large to be scanned or an un-scannable type.
> 
> could you do something like this instead? including the valid types. `<input type="file" name="myImage" accept="image/png, image/gif, image/jpeg" />`

Including valid types to input tag does not solve this issue, I can make mime type check before nsfw check

const mime = await mimeType(file) 
if(!mime){
  return 
}

something like this in Create.vue

josephmcg:
I see, we may want to add it anyways just so the user doesn't try to upload an invalid type.

I made a utility function that could help with the file byte check
`isEmbeddableImage`
utilities/FileType.ts

AlexZakablukov:
fixed

phillsatellite (approved):

josephmcg (approved):
thanks Alex, I appreciate your flexibility on this

PR #2689, opened 2022-04-01
fix(reply): fix 'if user sends reply chat will autoscroll to bottom'

genie-magic:
**What this PR does** 📖
fix 'if a user sends reply chat will auto-scroll to bottom'

AP-1182

### DIFF
--- a/store/textile/actions.ts
+++ b/store/textile/actions.ts
@@ -397,7 +397,6 @@ export default {
       throw new Error(TextileError.FRIEND_NOT_FOUND)
     }
 
-    commit('setMessageLoading', { loading: true })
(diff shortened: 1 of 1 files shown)


phillsatellite (approved):

ThomBos:
Should this scroll to the last reply?

genie-magic:
> Should this scroll to the last reply?

It doesn't need to scroll to the last reply. because once you reply, a reply message appeared.  Just check if it is not scrolling to bottom when reply.

molimauro (approved):

stavares843 (approved):

PR #2714, opened 2022-04-04
fix(files): remove text select from files table

josephmcg:
**What this PR does** 📖
- remove highlight issue Phil found on mobile

AP-1067

**Additional comments** 🎤


phillsatellite:
Tested: Is there anyway we can have it not highlight the entire list? 
https://user-images.githubusercontent.com/93608357/161579213-73708c2d-7c7b-42d9-a3e4-97be29af562f.mp4

josephmcg:
hoping this works, I can't test it

components/views/files/list/List.less
```diff
@@ -3,6 +3,8 @@
   &:extend(.full-width);
   font-size: @text-size;
   background: transparent;
+  user-select: none;
```
molimauro, line 6: `-webkit-user-select` is not needed here?

josephmcg: Would that work better?

molimauro: Yes it is but if you hover on Safari it say "only with -webkit prefix". Thanks Safari..

josephmcg: lol classic, thanks

molimauro (approved):

stavares843 (approved):

PR #2758, opened 2022-04-06
feat(libp2p): implement libp2p instead of p2pt wire and peer

tooshel:
**What this PR does** 📖

Implements libp2p in place of p2pt, wire, and peer

[AP-1365](https://satellite-im.atlassian.net/browse/AP-1365)

This will likely cause other bugs we didn't know about before. We can track those seperatly when you think this is good enough. 

**Additional comments** 🎤
a burrito is like a sleeping bag for beef


aewing:
> ⚠️ The title of this PR is invalid. Please make the title match this format `<type>(<scope>): <description>`. e.g.) `add(cli): enable --verbose flag` `fix(api): avoid unexpected error in handler`
> 
> allowed types - add, update, task, chore, feat, test, fix

what do you want from me

WanderingHogan (approved):

josephmcg (approved):

phillsatellite:
Tested: This works sooo much better I wanted to note down a few bugs that me and Hogan experienced but otherwise this is so cool. (For me all of these bugs were on Chrome)
1. I would attempt to call Hogan and call would never go through to his side
2. When Hogan would call me I would click the phone or video button to answer and modal would just keep appearing and I could never answer call
3. There was a moment when Hogan and I were video chatting and he received a call from Sheldon, at the same time I received a call from a profile I had logged into on Safari at the time.

Both of those issues seemed to happen after the first time we would successfully phone/video chat with each other and would sometimes be fixed by just refreshing browser page

phillsatellite (approved):

stavares843:
I have been testing this, from desktop and mobile browser and I also had the same first and second issues that Phil mentioned above

also, if you call from user A to user B and you hang up, doesn't update on user B side, still has the call or the call modal showing up

PR #2771, opened 2022-04-07
feat(jest): 100% coverage for FileType.ts

drepram:
**What this PR does** 📖

add coverage for FileType.ts, and making changes so that arrayBuffer can be used all across the codebase.

AP-1319

**Additional comments** 🎤


drepram:
Do not merge this yet, I think I can improve the coverage more. This is a WIP

WanderingHogan:
Do we need to keep a copy of a font we dont use for this?

drepram:
We need a file type that is not in the valid array:

image

An actual file, hence what came off my head was an OTF file. Is there something better that I could use on the unit test? And to answer your question, yes; to my knowledge, we need the file to be there to be tested, much like the SVG and the HEIC file being present.

phillsatellite (approved):

WanderingHogan:
@drepram There is already an svg file and heic file in that utilities/assets folder, can you use those? Also can you change it from assets to test_assets or something? I just looked into the utilities folder and was pretty confused why we had an assets folder with random files in there.

phillsatellite (approved):

stavares843 (approved):

stavares843:
thanks for addressing @WanderingHogan 's review 🔨

stavares843:
Captura de ecrã 2022-04-08, às 20 13 27

PR #2796, opened 2022-04-11
feat(files): dropzone for file uploads

josephmcg:
**What this PR does** 📖
- refactor files upload to accommodate file drag & drop
- improve tsconfig to support drag & drop event files
- prevent upload if currently in progress
- minor readability/type improvements along the way

AP-925

- I feel like the way I accessed the ref of a child component is wonky, any better ways?

**Additional comments** 🎤
- discussed with Phill. We eventually want to drag and drop files to move them around / upload to a specific folder on drop. This is a good first step

### DIFF
--- a/layouts/files.vue
+++ b/layouts/files.vue
@@ -140,7 +140,7 @@ export default Vue.extend({
      * @description Allows the drag and drop of files into the filesystem
      * @param e Drop event data object
      */
-    async handleDrop(e: DragEvent) {
+    handleDrop(e: DragEvent) {
       e.preventDefault()
 
       // if already uploading, return to prevent bucket fast-forward crash
--- a/layouts/files.vue
+++ b/layouts/files.vue
@@ -38,14 +38,16 @@
           />
         </swiper-slide>
         <swiper-slide class="dynamic-content">
-          <menu-icon
-            v-if="!showSidebar || $device.isMobile"
-            class="toggle--sidebar"
-            size="1.2x"
-            full-width
-            @click="toggleMenu"
-          />
-          <Nuxt id="files" ref="files" />
+          <DroppableWrapper @handle-drop-prop="handleDrop">
+            <menu-icon
+              v-if="!showSidebar || $device.isMobile"
+              class="toggle--sidebar"
+              size="1.2x"
+              full-width
+              @click="toggleMenu"
+            />
+            <Nuxt id="files" ref="files" />
+          </DroppableWrapper>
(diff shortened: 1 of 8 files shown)


store/ui/getters.ts
```diff
@@ -13,6 +13,9 @@ const getters = {
   swiperSlideIndex: (state: UIState) => {
     return state.swiperSlideIndex
   },
+  getFilesIndexLoading: (state: UIState) => {
+    return state.isLoadingFileIndex
+  },
```
molimauro, line 18: Its just my idea but im not too much fan of this getters where just a piece of state is returned w/o any sort of transformation. I prefer this way directly in the component:
```
mapState({
    isLoadingFileIndex: state => state.isLoadingFileIndex,
})
```
However its just my opinion we have other several placed where we follow this pattern :)

layouts/files.vue
```diff
@@ -125,10 +129,31 @@ export default Vue.extend({
...
+        return
       }
       this.$store.commit('ui/showSidebar', true)
     },
+    /**
+     * @method handleDrop
+     * @description Allows the drag and drop of files into the filesystem
+     * @param e Drop event data object
+     */
+    async handleDrop(e: DragEvent) {
```
molimauro, line 142: the async can be removed

store/ui/getters.ts
josephmcg: Good to know that's an option, thanks! I'm checking this state in many other components too, and was planning on refactoring those to use the getter after this goes through

layouts/files.vue
josephmcg: good call, leftover from copy paste lol

store/ui/getters.ts
molimauro: That sounds awesome! Thanks 🙏

molimauro (approved):

stavares843:
minor comments


- when creating a folder, if i drag a file while folder is being created appears warning saying ''upload in progress, please try again later'' - but that error is incorrect, because is not a upload thats in progress, is a folder thats being created

- i noticed on one of the uploads inside a folder, displayed this message, do we need to show this to the user?

Captura de ecrã 2022-04-13, às 18 48 22

stavares843 (approved):

stavares843:
[1402](https://satellite-im.atlassian.net/browse/AP-1402)
[1403](https://satellite-im.atlassian.net/browse/AP-1403)

PR #2818, opened 2022-04-12
fix(chat): show placeholder on image load error

AlexZakablukov:
**What this PR does** 📖
show placeholder on image load error

[AP-1378](https://satellite-im.atlassian.net/browse/AP-1378)

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/message/Message.vue
+++ b/components/views/chat/message/Message.vue
@@ -336,7 +336,7 @@ export default Vue.extend({
       const response = await fetch(imageSrc)
 
       if (!response.ok) {
-        throw new Error(this.$t('errors.chat.failed_upload') as string)
+        throw new Error(this.$t('errors.chat.failed_load') as string)
       }
 
       return response.blob()
--- a/components/views/chat/message/Message.vue
+++ b/components/views/chat/message/Message.vue
@@ -133,8 +133,14 @@ export default Vue.extend({
     this.cancelMessage()
   },
   async mounted() {
-    const data = await fetch(this.message.payload.url)
-    this.blob = await data.blob()
+    if (!this.message.payload?.url) {
+      return
+    }
+    try {
+      this.blob = await this.getImageBlob(this.message.payload?.url)
+    } catch (error: any) {
+      this.$Logger.log('error', error.message)
+    }
   },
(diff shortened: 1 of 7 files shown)


AlexZakablukov:
I need correct image for placeholder

stavares843:
added draft label due to be waiting for the image

components/views/chat/message/Message.vue
```diff
@@ -311,6 +317,19 @@ export default Vue.extend({
...
+        return
+      }
+
+      const response = await fetch(imageSrc)
+
+      if (!response.ok) {
+        throw new Error('Could not load image')
+      }
+
+      return await response.blob()
```
molimauro, line 331: `return await` is reduntant, simply `return`

components/views/chat/message/Message.vue
```diff
@@ -124,8 +124,14 @@ export default Vue.extend({
...
   async mounted() {
-    const data = await fetch(this.message.payload.url)
-    this.blob = await data.blob()
+    if (!this.message.payload?.url) {
+      return
+    }
+    try {
+      this.blob = await this.getImageBlob(this.message.payload?.url)
+    } catch (error: any) {
+      this.$Logger.log('error', error.message)
```
molimauro, line 133: can we put the image placeholder on blob error?

molimauro: nope just `this.blob = immage placeholder` here in catch

molimauro:
this is the placeholder image and the text that should be used:
Schermata 2022-04-13 alle 14 19 01

Can we try it? thanks ;)

molimauro:
Minor changes: "I would make the text color the lightest text color instead of white" and change text to "Image failed to load". Thanks!

AlexZakablukov:
done

molimauro (approved):

molimauro:
To test it, just run it locally and on `components/ui/Chat/Image/Image.html` change the image url with an invalid url 🔨

stavares843:
thanks @molimauro 🔨

stavares843 (approved):

stavares843:
image

neat 🎉

PR #2821, opened 2022-04-12
fix(mobile): user jump between different photo

baymax-codecare:
**What this PR does** 📖
while mobile users preview image in chat, the user can click image through mask overlay, this PR fix this
AP-1212

used setTimeOut function, not sure if there's any better solution for this. thanks.
**Additional comments** 🎤

### DIFF
--- a/components/ui/Chat/Image/Overlay/Overlay.html
+++ b/components/ui/Chat/Image/Overlay/Overlay.html
@@ -1,10 +1,10 @@
-<div class="image-mask" v-if="ui.chatImageOverlay">
-  <div class="image-container">
-    <img
-      class="is-image-in-mask"
-      :src="ui.chatImageOverlay.url"
-      v-click-outside="closeChatImageOverlay"
-    />
+<div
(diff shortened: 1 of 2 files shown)


phillsatellite (approved):

molimauro:
Honestly it feels strange to me to able to select photo in the background while having a preview opened. Shouldnt we just block the background selection?

baymax-codecare:
> Honestly it feels strange to me to able to select photo in the background while having a preview opened. Shouldnt we just block the background selection?

That's what this PR is fixing, no?

molimauro:
> > Honestly it feels strange to me to able to select photo in the background while having a preview opened. Shouldnt we just block the background selection?
> 
> That's what this PR is fixing, no?

Yes of course! What i meant is that is more natural to me to close when you click on the mask instead of "click outside the element" because the latter does not prevent you to click somewhere else.
I pushed some code maybe its more self explanatory :)

stavares843 (approved):

PR #2846, opened 2022-04-13
feat(files): large file downloads, improved qol, and improve extension fixing

josephmcg:
**What this PR does** 📖
- quality of life file download, only pulls file when the user clicks download button, rather than on fullscreen view
- no longer saves file blob, direct download using streamsaver (streams api). saves RAM
- on upload, record original file extension. if user has removed it via rename, apply again on file download name. this will save time, rather than byte checking on download. also less prone to bugs
- prevent concurrent downloads on same file with store
  - went with indexof and slice for performance over filter https://www.measurethat.net/Benchmarks/Show/10433/0/remove-an-item-from-array-indexof-splice-vs-filter#latest_results_block

AP-1397, AP-1360, AP-1400

- if user cancels download, console will go nuts with errors. still thinking about how to handle that
- it takes a second for the download to start on big files, but spinner indicates that it's working

**Additional comments** 🎤


molimauro (approved):

phillsatellite:
Tested: I tested the 2gb file as usual and got two errors this time, got this error mid download then I created new account and tried again and got 2nd error

Screen_Shot_2022-04-20_at_12 46 14_PM
Screen Shot 2022-04-20 at 12 59 22 PM

stavares843:
updated the snap 🔨 

Captura de ecrã 2022-04-21, às 00 32 11

but Jest is failing with the above, can you take a look? I can also ask Andre if you are too busy 🔨

josephmcg:
I would guess those are connection related. I would suggest testing this when solana/textile is more stable.
When I initially tested, I was able to push and pull a 4gb file

josephmcg:
@stavares843 you need to yarn install first

stavares843:
all clear 😂 both locally and CI

phillsatellite (approved):

stavares843 (changes requested):
https://user-images.githubusercontent.com/29093946/164540950-126c6522-7a9e-472f-b004-1cd538dc5d4f.mov

download works on test.satellite.one but doesn't work on this branch

same file type on both as well

no relevant info on the console log

josephmcg:
if you sent me the file, I'd be happy to test it. I always restart vue when switching to a relatively big feature branch. sometimes hot reload doesn't fully work for pages

stavares843 (approved):

PR #2878, opened 2022-04-14
fix(replies): change avatar color to correct one

AlexZakablukov:
**What this PR does** 📖
change avatar color to correct one

[AP-1411](https://satellite-im.atlassian.net/browse/AP-1411)

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/message/reply/Item/Item.vue
+++ b/components/views/chat/message/reply/Item/Item.vue
@@ -1,0 +1,76 @@
+<template src="./Item.html"></template>
+<script lang="ts">
+import Vue, { PropType } from 'vue'
+import { mapState } from 'vuex'
+import { UIMessage, Group } from '~/types/messaging'
+import {
+  getUsernameFromState,
+  convertTimestampToDate,
+  getAddressFromState,
+} from '~/utilities/Messaging'
+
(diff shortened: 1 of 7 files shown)


phillsatellite:
Tested: Noticed on Safari if user has a  profile picture, the profile picture will not appear in they're reply 
Screen Shot 2022-04-14 at 1 13 55 PM

AlexZakablukov:
fixed

phillsatellite (approved):

Jekrimo (approved):

components/views/chat/message/reply/Item/Item.vue
```diff
@@ -0,0 +1,84 @@
...
+      type: Function,
+      default: () => {},
+    },
+    emojiReaction: {
+      type: Function,
+      default: (): void => {},
+    },
+    showQuickProfile: {
+      type: Function,
+      default: (): void => {},
```
molimauro, line 41: @AlexZakablukov do you think we can emit here instead?

AlexZakablukov: done

molimauro (approved):

stavares843 (approved):

PR #2889, opened 2022-04-15
fix(files): persistent files upload status

josephmcg:
**What this PR does** 📖
- remove file upload progress bar in favor of percentage in status message. more consistent ui, less jarring than seeing the progress bar appear & disappear. Liz also said we don't use progress bars, so this is more cohesive with the rest of the app
- upload status will remain if you navigate away from this page and come back

AP-1395

**Additional comments** 🎤
I made a ticket for these
- since we're using store, I'd like to add status updates for other file operations (rename, remove, favorite) after this
- could probably also clean up store, switching to this rather than the boolean for keeping track of active files operations

### DIFF
--- a/components/views/files/controls/Controls.html
+++ b/components/views/files/controls/Controls.html
@@ -48,8 +48,12 @@
       multiple
     />
   </div>
-  <TypographyText v-if="status" class="status" :text="status" :size="6" />
-  <UiProgress v-if="progress < 100" :progress="progress" />
+  <TypographyText
+    v-if="ui.filesUploadStatus"
+    class="status"
+    :text="ui.filesUploadStatus"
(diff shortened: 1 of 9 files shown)


phillsatellite (approved):

Jekrimo (approved):

molimauro (approved):

stavares843 (changes requested):
Captura de ecrã 2022-04-20, às 00 43 49


i was uploading a file on files and showed 105%

josephmcg:
weird lol, I added something to display the smaller number given the percentage and 100.

stavares843:
thanks 😂

stavares843 (approved):

PR #2904, opened 2022-04-18
chore(vuetify): shrink bundle, remove vuetify

josephmcg:
**What this PR does** 📖
title

we were using v-datepicker, but its commented out at the moment. Seems like we can just use v-calendar though when we add it back.

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/search/result/Result.html
+++ b/components/views/chat/search/result/Result.html
@@ -45,6 +45,7 @@
       </multiselect>
     </div>
     <div class="container-filter date">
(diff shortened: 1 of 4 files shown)


Jekrimo (approved):

stavares843 (commented):
considering your comment, should we add a note on the v-datepicker that is commented out somewhere? regarding this change in case anyone removes that commented out in the future

josephmcg:
yep good idea, comment added

WanderingHogan (approved):

stavares843 (approved):

PR #2908, opened 2022-04-18
feat(notifications): initial set up

Jekrimo:
**What this PR does** 📖
Initial set up for textile and state for notifications.

AP-1371

 Written to be flexible with whatever needs we may have for notifications. I'm sure the data needs will evolve depending on the types of notifications we incorporate. One thing we will probably need is a listener to look out for unconventional notifications. Notifications from dm's, group messages, etc, should just shoot off from the successful send, like a child component. We need to decide what other types of notifications we will have, app-updates, satellite co notifications, missed calls, mentions, etc.


libraries/Textile/MailboxManager.ts
```diff
@@ -231,6 +231,7 @@ export class MailboxManager {
     const result = await this.textile.users.sendMessage(
       this.textile.identity,
       recipient,
+      recipient,
```
WanderingHogan, line 234: what's going on here @Jekrimo

Jekrimo: Typo! Good catch thank you

store/ui/actions.ts
```diff
@@ -8,6 +8,9 @@ import { Friend } from '~/types/ui/friends'
 import { Channel } from '~/types/ui/server'
 import { getFullUserInfoFromState } from '~/utilities/Messaging'
 import { getCorrectKeybind } from '~/utilities/Keybinds'
+import { Notification, Notifications } from '~/types/ui/notifications'
```
stavares843, line 11: seems this import is not used 🔨

Captura de ecrã 2022-04-18, às 23 57 35

libraries/Enums/types/notification-types.ts
```diff
@@ -0,0 +1,10 @@
+export enum NotificationTypes {
+  DIRECT_MESSAGE = 'directMessage',
+  GROUP_MESSAGE = 'groupMessage',
+  MEANTIONS_NOTIFICATION = 'meantionNotification',
```
stavares843, line 4: is this expected to be mentions instead? 🔨

store/ui/actions.ts
Jekrimo: Good looking out, thank you!!!

libraries/Enums/types/notification-types.ts
Jekrimo: A++ spelling from me haha, thanks!

store/ui/mutations.ts
```diff
@@ -347,6 +352,25 @@ export default {
...
+    })
+  },
+  notificationSeen(state: UIState, notificationId: String) {
+    state.notifications.allNotifications.map((item) => {
+      if (item.id === notificationId) {
+        item.seen = true
+      }
+      return item
+    })
+  },
```
josephmcg, line 373: map should be used when you're doing a specific operation on each array item. Could you use array.find() here instead?

components/views/navigation/toolbar/alerts/Alerts.html
```diff
@@ -6,8 +6,14 @@
     :key="alert.id"
     @sync="syncAlerts"
   />
+  <ToolbarAlertsAlert
+    v-if="notifications.length !== 0"
```
josephmcg, line 10: this could be shortened to `notifications.length` since 0 is falsy

components/views/navigation/toolbar/alerts/Alerts.vue
```diff
@@ -2,11 +2,26 @@
...
   data() {
     return {
+      allNotifications: [],
       alerts: this.$Alerts.all,
+      allSeen: true,
+    }
+  },
+  mounted() {
+    if (this.$props.notifications?.includes(false, ['seen'])) {
+      this.allSeen = false
```
josephmcg, line 24: same here, is there a reason we do it on mount? could definitely make a computed property out of this

store/ui/mutations.ts
Jekrimo: Thanks for all your help, Joe!

stavares843:
Captura de ecrã 2022-04-29, às 00 00 55

we need to ditch the mock notifications we have there

stavares843:
Captura de ecrã 2022-04-29, às 00 00 26

im getting messages on the same chat that has the alerts, if the user is already on the chat and getting those messages, getting alerts of those same messages seems spam

stavares843:
Captura de ecrã 2022-04-29, às 00 00 26

going to the alerts doesn't clear out all the alerts, and we don't have a clear all button so user needs to clear them one by one

stavares843:
Captura de ecrã 2022-04-29, às 00 00 26

we no longer use the ''few seconds ago'' logic, that was changed [here](https://github.com/Satellite-im/Core-PWA/pull/2506) so we need to change that logic on the notifications as well to be coherent on what we are using on the chat

stavares843:
Captura de ecrã 2022-04-29, às 00 00 26

the closing icon on each alert, just closes the modal and doesn't clear the alert, is this gonna be handled on a separate PR?

stavares843:
Captura de ecrã 2022-04-29, às 00 00 26

we are not using the user's avatar on the notification

stavares843:
Captura de ecrã 2022-04-29, às 00 08 09

the notifications only say ''user ''x'' sent a message'', are we gonna use the same `'user sent you a {msgType}'` that we added [here](https://github.com/Satellite-im/Core-PWA/commit/cffc33556e3f77084c47e541ea5ef735e9c23bf8#diff-93b6f569059fddd160ddb58774259cd699eba83c03c74903cbe2097a7e146c14R75)?

stavares843 (changes requested):
couple comments

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -116,12 +117,18 @@ export default Vue.extend({
         state: { isOpen: true, group },
       })
     },
+    getNotifications() {
+      this.$data.notifications = this.ui.notifications
+    },
```
josephmcg, line 122: could you remove this if you're not using it?

components/views/navigation/toolbar/Toolbar.html
```diff
@@ -28,23 +28,27 @@
...
-        @click="$data.showAlerts = !$data.showAlerts"
       >
+
+        <UiCircle
+          v-if='ui.unseenNotifications'
+          v-on:click='toggleAlerts'
+          class='notification-alert'
+          :size='15'
+          color='red'
+          :type='"icon"'
```
josephmcg, line 41: try this instead. use v-bind (shorthand `:`) when you need a dynamic js value. this is just a string
`type="icon"`

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -133,7 +140,7 @@ export default Vue.extend({
      */
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     handleChange(value: string, item: SearchQueryItem) {
-      this.searchQuery = ''
+      this.$data.searchQuery = ''
```
josephmcg, line 143: remove after fixing `this` context

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -144,15 +151,15 @@ export default Vue.extend({
      */
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     handleSearch(value: string, items: SearchQueryItem[]) {
-      this.searchQuery = value
+      this.$data.searchQuery = value
```
josephmcg, line 154: remove after fixing `this` context

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -144,15 +151,15 @@ export default Vue.extend({
...
     },
     /**
      * @method toggleSearchResult DocsTODO
      * @description
      * @example
      */
     toggleSearchResult() {
-      this.searchQuery = ''
+      this.$data.searchQuery = ''
```
josephmcg, line 162: remove after fixing `this` context

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -59,6 +60,7 @@ export default Vue.extend({
       searchRecommend,
       showAlerts: false,
       searchQuery: '' as string,
+      notifications: [],
```
josephmcg, line 63: remove if not using

josephmcg (changes requested):

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -116,12 +117,18 @@ export default Vue.extend({
         state: { isOpen: true, group },
       })
     },
+    getNotifications() {
+      this.$data.notifications = this.ui.notifications
+    },
+    toggleAlerts() {
+      this.$data.showAlerts = !this.$data.showAlerts
+    },
```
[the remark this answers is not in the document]
Jekrimo: These two are under the methods: {}, Do we need returns on all functions within methods?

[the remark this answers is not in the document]
Jekrimo: Got rid of the getNotifications function, left over. but, does the toggle need to have a return on it then?

components/views/navigation/toolbar/Toolbar.html
Jekrimo: the :type prop of the <UiCircle> expects certain strings of, type CircleType = 'image' | 'icon' | 'random' . We should just be passing a string right?

josephmcg: you're still using the type prop. : is just shorthand for v-bind

stavares843 (changes requested):
Captura de ecrã 2022-05-03, às 19 41 23

the clear button is neat 💥 maybe we can remove the glowing as well because with the glowing seems it lacks padding

stavares843 (changes requested):
Captura de ecrã 2022-05-03, às 19 41 34


when trying to clear out a message, other icons are overlapping

stavares843 (changes requested):
Captura de ecrã 2022-05-03, às 19 42 20

after clearing up all messages using the clear button, going again to alerts only says ''alerts'' 

before having any alerts there was saying ''you are all caught up'', but then I got 2 alerts, cleared both with the clear button and now going again to alerts only says ''alerts''

stavares843:
I can add a ticket to handle the avatar + timestamp logic if we wanna avoid getting the PR even bigger, should we also handle the fact of user getting messages while in chat in other PR?

Jekrimo:
> 

Got everything fixed up, thank you for catching those! For the button glow I'm not going to bother, the glow is just coming from our pre-set 'primary' class for theming. But I think we will totally trash that button sooner than later, either replace with just an icon that does the same thing, or have an infinity scroll for notifications.

Jekrimo:
> 2 very minor things I found but other than that this is working pretty good! But one I found was I feel like it should list the message alerts newest at the top older message alerts at the bottom Screen Shot 2022-05-04 at 11 59 46 AM
> 
> Another thing I found was hovering over the scrollbar in Alerts will cover up half of the X X

All fixed! That sort gave me some seriously weird issues, sorry for the delay!

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -58,13 +58,13 @@ export default Vue.extend({
...
   },
   computed: {
     ...mapState(['ui', 'audio', 'video', 'webrtc']),
-    ...mapGetters('ui', ['showSidebar']),
-    selectedGroup() {
+    ...mapGetters('ui', ['showSidebar', 'allUnseenNotifications']),
+    selectedGroup(): string {
       return this.$route.params.id // TODO: change with groupid - AP-400
     },
     recipient() {
```
josephmcg, line 70: please add return types to this computed property and any others that are missing it. if unclear, intellisense should tell you what to put on hover (image below)
- recipient `Friend | { textilePubkey: string; type: string } | null | undefined`
- showSearchResult get `boolean`

image

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -58,13 +58,13 @@ export default Vue.extend({
     return {
       searchRecommend,
       showAlerts: false,
-      searchQuery: '' as string,
```
josephmcg, line 61: these castings help other devs understand the variable type of data attributes, please don't remove

components/views/navigation/toolbar/alerts/Alerts.html
```diff
@@ -1,13 +1,27 @@
 <div id="alerts">
+  <InteractablesButton
+  v-if='ui.unseenNotifications'
```
josephmcg, line 3: can you switch this to the new getter? you can also make it a self closing element since there's no nested tags

store/ui/mutations.ts
```diff
@@ -338,6 +338,36 @@ export default {
       count: 1,
     })
   },
+  sendNotification(state: UIState, notification: Alert) {
+    state.unseenNotifications++
```
josephmcg, line 342: can you remove all mentions of this unused state after switching the last place it's used to the getter?

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -116,6 +115,12 @@ export default Vue.extend({
         state: { isOpen: true, group },
       })
     },
+    getNotifications() {
+      this.$data.notifications = this.ui.notifications
```
josephmcg, line 119: this appears to be unused. could you please remove it? I also mentioned this in my previous review, but it was marked as resolved.

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -116,6 +115,12 @@ export default Vue.extend({
...
+    },
+    toggleAlerts() {
+      this.$data.showAlerts = !this.$data.showAlerts
```
josephmcg, line 122: please remove the $data after adding computed return types, it will fix the `this` context

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -132,7 +137,7 @@ export default Vue.extend({
      * @example
      */
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    handleChange(value: string, item: SearchQueryItem) {
```
josephmcg, line 135: since these are not returning anything, remove the :string

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -143,15 +148,15 @@ export default Vue.extend({
      * @example
      */
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    handleSearch(value: string, items: SearchQueryItem[]) {
+    handleSearch(value: string, items: SearchQueryItem[]): string {
```
josephmcg, line 151: since these are not returning anything, remove the :string

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -143,15 +148,15 @@ export default Vue.extend({
...
       this.searchQuery = value
     },
     /**
      * @method toggleSearchResult DocsTODO
      * @description
      * @example
      */
-    toggleSearchResult() {
+    toggleSearchResult(): string {
```
josephmcg, line 159: since these are not returning anything, remove the :string

components/views/navigation/toolbar/alerts/Alerts.html
```diff
@@ -1,13 +1,27 @@
...
+  :size='"small"'
+  :action='clearNotifications'
+  :style='"margin-bottom:5px"'
+  :type='"primary"'
```
josephmcg, line 7: we talked about this before, no need to use v-bind if it's just a static string
` size='small'
   style='margin-bottom:5px'
   type='primary'`

components/views/navigation/toolbar/alerts/alert/Alert.html
```diff
@@ -15,7 +20,7 @@
     </div>
     <!--Alert description-->
     <div class="description">
-      <TypographyText :text="localAlert.content.description" />
+      <TypographyText v-if="localAlert.type === AlertType.DIRECT_MESSAGE" :text="localAlert.from + ' ' + $t('messaging.user_sent.user', {msgType: localAlert.type})" />
```
josephmcg, line 23: when markup gets complicated like this, its often better to break this out into a computed property. Could you also update the translation and pass it the username? this is not scalable due to languages having different grammatical structure
https://vue-i18n.intlify.dev/guide/essentials/syntax.html#literal-interpolation

components/views/navigation/toolbar/alerts/alert/Alert.vue
```diff
@@ -3,31 +3,35 @@
...
+    },
+    notification: {
+      type: Object as PropType<Alert>,
+      default: () => {},
+      required: false,
     },
   },
   data() {
     return {
       localAlert: this.$props.alert,
```
josephmcg, line 24: I mentioned this before as well.. but this is a weird way to use props. why not remove the data attribute and just use the prop directly? you marked as resolved without changing or answering why

image

components/views/navigation/toolbar/alerts/Alerts.vue
```diff
@@ -2,17 +2,33 @@
...
+    notifications: {
+      type: Array as PropType<Array<Alert>>,
+      default: () => [],
+      required: false,
+    },
+  },
   data() {
     return {
       alerts: this.$Alerts.all,
+      allSeen: true,
```
josephmcg, line 19: this is not used, remove

josephmcg (changes requested):
i mentioned a lot of these in my last review and they were not resolved. reach out if you have questions.

--edit - chatted with Jeff, noticed we have both an Alerts global and store. Does this require both or should we only use one? We thought it seemed simple enough to only warrant one

components/views/navigation/toolbar/Toolbar.vue
Jekrimo: This tripped me up really hard. The get() and set() needing returns, messed me up pretty good. I finally get it as after returning on those it fixed it! Makes sense.... kinda lol I'll just let Vue be weird

components/views/navigation/toolbar/alerts/alert/Alert.html
Jekrimo: image

I ended up creating a switch for the translate, thinking ahead when we can have all sorts of types coming into the alerts. Is this ok?

phillsatellite:
One more tiny thing I noticed when re testing that Sara said she hadn't seen before was if the user has 0 notifications the first one they do receive will open automatically 

https://user-images.githubusercontent.com/93608357/167736975-6d65327e-9138-4530-9c0e-c34fa8da7d84.mov

stavares843:
yes when I tested previously wasn't open automatically, at least here 💥

Jekrimo:
My bad! It wasn't doing it for a bit and then started up again now when I just looked, got a fix for it!

phillsatellite:
I found one other super minor thing I wasnt sure if you wanted me to create another ticket for or to fix on this pr but I think the icon could use a little more click area and have it so the user cant highlight the number of alerts 


https://user-images.githubusercontent.com/93608357/167884348-0b5d489e-311e-420c-86df-c4685ad150e7.mov

Jekrimo:
Just pushed fix for this one!

libraries/Textile/TextileManager.ts
```diff
@@ -95,6 +97,12 @@ export default class TextileManager {
     // UserInfoManager initializes itself during the creation
     this.userInfoManager = new UserInfoManager(textile)
     await this.userInfoManager.init()
+
+    this.notificationManager = new NotificationManager(
+      textile,
+      textile.identity,
+    )
```
josephmcg, line 104: since you're passing textile anyway, why not derive identity from textile inside the constructor?

josephmcg, line 104: couldn't you derive identify inside the constructor from textile, since you're passing it anyway?

josephmcg (changes requested):
thanks Jeff, a couple changes. I still stand by the decision to rip out the Alerts global since we're not using any of the methods. It's essentially functioning as a redundant, second store.

In the interest of time, we can remove it later in a separate branch

components/views/navigation/toolbar/Toolbar.vue
```diff
@@ -63,11 +64,15 @@ export default Vue.extend({
   },
   computed: {
     ...mapState(['ui', 'audio', 'video', 'webrtc']),
```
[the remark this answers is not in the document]
Jekrimo: Rad, this actually makes way more sense than what we were doin.

libraries/Textile/TextileManager.ts
Jekrimo: Not even using identity, so just trashed it.

stavares843:
Captura de ecrã 2022-05-12, às 20 34 33

@drepram this test is failing on this PR, could you take a look and fix it here pls? thanks 🔨

Jekrimo:
> Captura de ecrã 2022-05-12, às 20 34 33
> 
> @drepram this test is failing on this PR, could you take a look and fix it here pls? thanks 🔨

My bad, I forgot to mention this! @drepram - I removed our Alert Class, we will be only using state for alerts and notifications now. I'm not sure how that effects tests, but that's probably why it's failing I would suspect.

phillsatellite (approved):

josephmcg:
i'll check again after conflict fixes

josephmcg (changes requested):
thanks Jeff, appreciate your patience here. one more minor change and it looks good

stavares843 (approved):

PR #2917, opened 2022-04-19
feat(settings): nsfw filter user setting

josephmcg:
**What this PR does** 📖
- user setting to obscure nsfw content, stored in textile. found in front end user privacy settings
- should be on by default
- minor language changes on settings page

AP-1430

- does not affect chat or files yet, that will be a separate PR. this is only the setting
- make sure it stays the same across different devices/browser sessions. 

**Additional comments** 🎤


Jekrimo (approved):

WanderingHogan (approved):

stavares843 (changes requested):
Captura de ecrã 2022-04-20, às 00 59 32

you mentioned on the PR description that should be on by default but when I changed branches with an already logged in account, going to settings that toggle was off


but when creating a new account was already on

Captura de ecrã 2022-04-20, às 01 09 44

josephmcg:
it may have gotten set to undefined on textile initialization, just added something to prevent that. not sure if it will work on pre-existing accounts. I think this is more focused on new accounts

josephmcg:
it may have gotten set to undefined on textile initialization, just added something to prevent that. not sure if it will work on pre-existing accounts. I think this is more focused on new accounts

phillsatellite (approved):

phillsatellite:
Verified branch is fixed and worked across all devices 👍

stavares843 (changes requested):
Captura de ecrã 2022-04-20, às 14 04 43

when creating a new account, the toggle is locked as activated, the user can't edit

josephmcg:
I realize what the issue is, the place to store that information in textile isn't available yet, so it's throwing an error before it can mutate the state. the file scanning button appears to work, but it's only mutating the state. I'll work on fixing these

stavares843:
let me know if this is ready 🎉

josephmcg:
@stavares843 ready! After discussing with Hogan, we decided to remove settings page from account creation. the other option was initializing textile earlier which we do not want.

After first login, I would like to add some kind of new user 'tour' highlighting various functionalities and where to find settings

stavares843:
thanks @josephmcg 🔨

stavares843 (approved):

PR #2924, opened 2022-04-19
fix(emoji-scroll): emoji persist on scroll

ThomBos:
**What this PR does** 📖
Fixes issue with emoji which happen to disappear when user clicks outside emojis container (but inside modal)

AP-1432

**Additional comments** 🎤


molimauro:
Thanks @ThomBos, need only to check the library since we're introducing workarounds every time we have problems with this. I'll let you know

molimauro (approved):

molimauro:
@ThomBos i added the stopPropagation on the enhancer div so that we prevent the lib outside click of being fired when clicking on it.

stavares843 (approved):

PR #2940, opened 2022-04-20
test(cypress): unskip tests and small improvements

luisecm:
**What this PR does** 📖
- Unskip cypress tests that should be working now
- Added data-cy attributes for sidebar friends used in cypress commands
- Refactored chat elements tests into one single spec in order to avoid importing account for each spec. Now these are separated by describe blocks but account import process is only done once
- Refactored mobiles-responsiveness tests to avoid switching viewport for each it block
- Removed chat elements file specs that are now part of chat-elements-validations.js
- Improvement of goToConversation cypress command to work correctly for mobiles and standard viewports 
- Update of the validation of glyph modal cypress command to match with the current DOM elements

AP-1441

- Sending this one as draft for now, since there are still issues on the Linking Satellites page taking longer than the timeouts already setup and for the recovery seed screen showing random 503 errors. So we might need to keep skipping some of these tests

**Additional comments** 🎤


stavares843:
awesome 🎉

PR #2949, opened 2022-04-20
chore(glyphs): add thumbnail and small copies of glpyhs to boost performance

WanderingHogan:
**What this PR does** 📖
This points the glyph picker to thumbnail versions of the images that are 64x64 and the chat to 200x200 small versions to load faster

https://satellite-im.atlassian.net/browse/AP-1446

**Additional comments** 🎤


components/views/chat/message/glyph/Glyph.vue
```diff
@@ -18,6 +18,12 @@ export default Vue.extend({
     ...mapState(['ui']),
   },
   methods: {
+    getSource() {
+      if (this.source.includes('/$1/')) {
+        return this.source.replace('$1', 'small')
+      }
+      return this.source
+    },
```
josephmcg, line 26: could we switch this to computed? other than that, looks good

josephmcg, line 26: could we switch this to computed? other than that, looks good

WanderingHogan: great idea, thank you

josephmcg (approved):

WanderingHogan:
@phillsatellite @stavares843 to really see the value in this, load the page for localhost and dev in incognito mode.

In both, import an account with friends you can message, and before you open the glyphs menu:
 open devtools > go to network tab > disable cache > set throttling to fast 3g
 
 then compare the loading times between local and dev. The astrobunnys are slow but you see the others load in much faster

stavares843 (commented):
can you double the resolution so we can see how that looks? 🔨

stavares843 (approved):

PR #2980, opened 2022-04-26
fix(chat): fix quick reaction context menu

josephmcg:
**What this PR does** 📖
- slight store cleanup, remove unused state, remove unused action
- fix quick reaction in message context menu

AP-1085

**Additional comments** 🎤


Jekrimo (approved):

store/ui/actions.ts
```diff
@@ -30,10 +30,6 @@ export default {
   setActiveChannel({ commit }: ActionsArguments<UIState>, channel: Channel) {
     commit('setActiveChannel', channel)
   },
-  addReaction({ commit }: ActionsArguments<UIState>, reaction: any) {
-    commit('addReaction', reaction)
-    commit('updateRecentReactions', reaction.emoji)
```
molimauro, line 35: wo this recent reactions arent updated anymore right?

josephmcg: yes, but I believe that was unused state from when the app was being mocked. the actual data was being tracked in `mostEmojiUsed`, which could probably be renamed to something more natural

molimauro: ok ty!! :)

molimauro (approved):

store/ui/actions.ts
josephmcg: I guess those associated mutations could be removed too, didn't notice that before

josephmcg:
accidentally added some unrelated enum file lol.. fixing

molimauro:
thanks!!

stavares843 (approved):

PR #2997, opened 2022-04-27
feat(files): recent sidebar functionality

josephmcg:
**What this PR does** 📖
- files right sidebar integrated w/ vue router. will append query string based on selected route and work w/ browser forward back buttons
- added default option since other options will not show the user the regular file system anymore
- on Recent selection, show the user the 15 most recently edited files
- if user manually enters a route that isn't valid, reset to default view
  - this effectively breaks the back button. It will be easier to maintain from our perspective though, especially if we integrate the folders with router too
- under the hood changes to filesystem flat getter, should be more useful going forward
- move UiSimpleList to Files component folder since that's the only place we're using it. it quickly became specialized. I could repurpose it to make it more generic if we plan on using it anywhere else too
- better dynamic icon for UiSimpleList. this could be implemented in other places as well
- move a couple files to be more properly nested, easier to navigate

AP-1468

- only Recent route shows you anything special. the others will be highlighted, but just show you the Default view

**Additional comments** 🎤


Jekrimo (approved):

josephmcg:
reworking something rn, will remove draft when ready

libraries/Files/test/FilSystem.test.ts
```diff
@@ -781,4 +783,189 @@ describe('FilSystem.findItem', () => {
...
+      const result: Item[] = filesystem.sortContent(
+        {
+          category: FileSortEnum.TYPE,
+          asc: false,
+        },
+        directory?.content,
+      )
+      expect(result).toMatchSnapshot()
+    })
+  })
```
stavares843, line 970: thanks for adding 🎉

molimauro (approved):

stavares843:
Captura de ecrã 2022-05-02, às 18 48 47

the PR description says 
> on Recent selection, show the user the 15 most recently edited files

but its showing files edited 19 or 20 days ago, is this intended?

stavares843:
also, if you are inside a folder, and click on recent, it just shows you all other recent ones even the ones outside the folder, should that filter the recent files only related to the folder the user selected before?

stavares843:
also, something else I noticed was the following

Captura de ecrã 2022-05-02, às 19 11 46

default filter

2 files zip

Captura de ecrã 2022-05-02, às 19 11 37

recent filter

4 files zip

the recent couldn't have more than the default one right?

then clicking on the second zip on the recent view

Captura de ecrã 2022-05-02, às 19 11 37


crash's

Captura de ecrã 2022-05-02, às 18 53 37


I couldn't reproduce that crash on dev, because when I change to dev branch, using the same exact account on the same browser I don't have that file

only 2 zip appear (same as screenshot 1 when using the default filter on this branch)

Captura de ecrã 2022-05-02, às 19 19 44



seed

high silk kite trend leopard fuel dynamic reform actual replace impulse floor

let me know if this is confusing 💥

stavares843:
i usually add as review but in this case most were only minor questions except the crash 😂 

thanks for clarifying all the above points 🔨 the crash is always reproducible on my side, using that flow with that specific account on that specific file

i also did a new server on both dev and this branch and the crash was always reproducible

the thing is, i don't have that file at all on dev, and on this branch he only appears using the recent filter - so the question is, where is this file from?

josephmcg:
oh, that's peculiar. If the file is nested deep inside folders somewhere, recent would bring it to the top. I just fixed an logic error. my bad, it wasn't showing the most recently edited files.

If the file is somewhat old, it may be stored inside the bucket with name rather than uuid. This was a breaking change from a few weeks ago. Could you confirm that everything works ok with new files?

stavares843:
is this ready to test? 🔨

josephmcg:
ready now, you should be able to delete/rename/favorite any file in recent view

stavares843:
thanks 🔨

phillsatellite:
Hi Joe! I tested this again and I wasn't able to replicate the crash that Sara had but I noticed if I renamed a file it wouldn't appear at the top of the recent list and the modified date wouldn't change either. Is that behavior intended?

https://user-images.githubusercontent.com/93608357/167165139-daa5f184-6492-4c08-a2d7-2b3f2804d9c0.mov

josephmcg:
yes, i mimicked how macos file system handles it. changing name does not change the contents of the file, so it doesn't update the modified date. if we want it to behave differently, that should be a separate issue

phillsatellite (approved):

stavares843 (approved):

PR #3076, opened 2022-05-06
refactor(style): code refactor and clean up, added flair gradients

ThomBos:
**What this PR does** 📖
Moves all color/theme code into a single file in order to make it easier to change it
Added gradients for all flair colors

AP-1507

**Additional comments** 🎤


phillsatellite:
Noticed something when first logged in, anything that is supposed to be default colored "#286CFE" it appears as black until the user sets moonlight theme and then back to default again

https://user-images.githubusercontent.com/93608357/167892247-08c7d55a-1aeb-4f7b-be7c-15efe31f960e.mov

ThomBos:
> Noticed something when first logged in, anything that is supposed to be default colored "#286CFE" it appears as black until the user sets moonlight theme and then back to default again
> 
>  resetting.theme.mov

If the account used isn't a new one, it's normal you get this issue: flairs are now passed as an array, but you have it just as a string.
Let me know if this is the case please 👍

phillsatellite:
@ThomBos Awesome I verified that I was using an old account, made a new one and verified that issue didnt happen 👍  I did see something a little shady I wanted to ask you about, should the notification message that appears when copying account id or sending request be themed as well? Right now it just appears as black with grey text

https://user-images.githubusercontent.com/93608357/168159533-639c05d9-e854-4f42-a881-b7d6832f0d17.mov

ThomBos:
> @ThomBos Awesome I verified that I was using an old account, made a new one and verified that issue didnt happen 👍 I did see something a little shady I wanted to ask you about, should the notification message that appears when copying account id or sending request be themed as well? Right now it just appears as black with grey text
> 
ah... yeah this isn't the correct behaviour. it should be colored with flair colors, thanks for reporting it! 👍

assets/styles/base.less
```diff
@@ -1,10 +1,12 @@
 :root {
   --app-height: 100vh;
-  --flair-color: @primary-color;
-  --flair-color-rgb: 39, 97, 253;
+  --flair-color: #2761fd;
+  --flair-color-secondary: #286cfe;
```
molimauro, line 4: Can we use variables here instead of color value? I'd like to have all color values in one place like `colors.less`.

ThomBos: it might look like a snake biting his tail, but it's possible doing it 👍 

anyway, that value gets overwritten when the app starts, so it's just like a placeholder

assets/styles/base.less
```diff
@@ -188,133 +256,171 @@ body {
...
 
-button {
-  outline: none !important;
-  border: none !important;
+//Droppable wrapper
+.droppableWrapper {
+  &.droppableWrapper--active {
+    &:after {
+      border-color: @flair-color;
+      background-color: rgba(var(--flair-color), 0.25);
```
molimauro, line 274: here and in some other files we have to change this to @flair-color, can you check?

ThomBos: seems like something went wrong with the rebase, i go check it out

ThomBos: ok those variables are correct! i used css variables because rgba function doesn't work with less variables. (there are just 2 functions rgba)

Jekrimo (approved):

assets/styles/themes/moonless_night.less
```diff
@@ -66,14 +44,10 @@
...
       svg {
-        color: @white !important;
+        &:extend(.font-primary);
       }
       &.active {
         border-radius: @half;
-        color: var(--flair-color) !important;
-        svg {
-          color: var(--flair-color) !important;
-        }
```
molimauro, line 76: These colors are lost when active

dev

image

here

image

ThomBos: didn't notice both, thanks for reporting 🔨

components/interactables/Button/Button.less
```diff
@@ -3,6 +3,7 @@
   min-width: 35px;
   overflow: hidden;
   line-height: unset;
+  transition: 250ms all ease-in;
```
molimauro, line 6: why this?

layouts/server.vue
```diff
@@ -123,9 +122,6 @@ export default Vue.extend({
     flairColor() {
       return this.ui.theme.flair.value
     },
-    flairColorRGB() {
-      return hexToRGB(this.ui.theme.flair.value)
-    },
```
molimauro, line 128: This was fixing a bug on safari [AP-1207](https://satellite-im.atlassian.net/browse/AP-1207)

layouts/friends.vue
```diff
@@ -120,9 +119,6 @@ export default Vue.extend({
     flairColor() {
       return this.$store.state.ui.theme.flair.value
     },
-    flairColorRGB() {
-      return hexToRGB(this.$store.state.ui.theme.flair.value)
-    },
```
molimauro, line 125: same bug

layouts/files.vue
```diff
@@ -110,9 +109,6 @@ export default Vue.extend({
     flairColor(): string {
       return this.$store.state.ui.theme.flair.value
     },
-    flairColorRGB(): string {
-      return hexToRGB(this.$store.state.ui.theme.flair.value)
-    },
```
molimauro, line 115: same bug

ThomBos:
i preferred to fix it in this way, still can go back though.
let me know what's the best way!

components/interactables/Button/Button.less
ThomBos: i was testing why buttons weren't working (they were keeping black/dark grey background at first),
but i forgot to remove it... do i have to change it? 😬

molimauro:
> i preferred to fix it in this way, still can go back though. let me know what's the best way!

Its ok as long as we dont reintroduce the bug resolved in the past. Did you check it? thanks :)

components/interactables/Button/Button.less
molimauro: I'd ask Liz if its ok to introduce a transition here ;)

ThomBos:
> > i preferred to fix it in this way, still can go back though. let me know what's the best way!
> 
> Its ok as long as we dont reintroduce the bug resolved in the past. Did you check it? thanks :)

yeah it's working right now! 👍 

the only issue you might have is the one phill reported, solved just changing theme once

molimauro (approved):

stavares843 (changes requested):
when using the default blue theme, these warnings should be the same color as the theme, in this case, blue instead of red

this branch
Captura de ecrã 2022-05-19, às 14 57 07

dev


Captura de ecrã 2022-05-19, às 14 57 16

ThomBos:
> when using the default blue theme, these warnings should be the same color as the theme, in this case, blue instead of red
> 
> this branch Captura de ecrã 2022-05-19, às 14 57 07
> 
> dev
> 
> Captura de ecrã 2022-05-19, às 14 57 16

did you create that account right now? if not, you have to chage theme once to update and use the new flairs

stavares843 (changes requested):
when you are using a theme example peach and hover on a already selected item

Captura de ecrã 2022-05-19, às 15 02 44
Captura de ecrã 2022-05-19, às 15 02 48

doesn't happen on dev

stavares843 (changes requested):
the upgrade button on files should be grey

Captura de ecrã 2022-05-19, às 15 09 24

dev


Captura de ecrã 2022-05-19, às 15 09 35

stavares843:
> > when using the default blue theme, these warnings should be the same color as the theme, in this case, blue instead of red
> 
> did you create that account right now? if not, you have to chage theme once to update it and use the new flairs. ps. let me know if the issue persist 🐛
> 
> Edit. tested and got the same bug, going to fix it now



yes all the above was all with new accounts 💥

stavares843 (changes requested):
https://user-images.githubusercontent.com/29093946/169315997-a300d61d-b4ea-4f80-933a-881bcb4e33a2.mov

on this branch vs dev, its adding a shadow of the theme color on the call buttons on the corner, video above

ThomBos:
i've almost fixed all bugs above, but i think we should change something for the buttons: we have `:disabled` and `:active` html props, then ```.is-active``` and `.active` as css classes, and this is kinda messy when it's about styling and changing props... 
@molimauro are we going to make any refactor for that?

stavares843:
thanks @ThomBos 🔨

molimauro:
> i've almost fixed all bugs above, but i think we should change something for the buttons: we have `:disabled` and `:active` html props, then `.is-active` and `.active` as css classes, and this is kinda messy when it's about styling and changing props... @molimauro are we going to make any refactor for that?

We're using buefy for the UI design so we have some predefined classes like `is-active` etc. I'm already considering ditching it in the future, we'll see when. But if there is one thing sure is that it'll be your job @ThomBos, you're our refactor man

stavares843:
😂

molimauro (approved):

phillsatellite:
Hi Thomas! Everything above is fixed on my end! I only found one small issue when retesting, looks like the usernames in call are missing the grey box behind them. Other than that looked perfect thanks for everything! 🔥 
This branch 
flare branch

Dev branch 
dev branch

stavares843:
thanks @phillsatellite 🔨 

i'm gonna merge this and add that as a ticket

stavares843:
[AP-1603](https://satellite-im.atlassian.net/browse/AP-1603)

stavares843 (approved):

PR #3080, opened 2022-05-09
fix(search): refine search match criterion

josephmcg:
**What this PR does** 📖
- only show search match if term matches payload. setup in a way that can be refined later when we add different search criteria (sender, etc..)
- previously, any message would hit a `conversation` match if the conversation ID was passed as the search query
- add return types to some computed props

AP-1333

**Additional comments** 🎤

### DIFF
--- a/types/search/search.ts
+++ b/types/search/search.ts
@@ -164,3 +164,14 @@ export type QueryOptions = {
   dateRange: DateOptions | null
   perPage: number
 }
+
+export enum MatchTypesEnum {
+  ID = 'id',
+  CONVERSATION = 'conversation',
+  FROM = 'from',
+  TO = 'to',
+  AT = 'at',
+  READAT = 'readAt',
+  TYPE = 'type',
+  PAYLOAD = 'payload',
+}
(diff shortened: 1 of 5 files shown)


molimauro (approved):

phillsatellite:
Tested: When I searched for the link I sent in chat, 2 other messages appeared that had the word chat in them. Should they still appear or should it only be the link? Another small thing I wasn't sure about is if I search "he" both Hey and Hi appear but not Hello. Should anything appear for just "he"?
Screen Shot 2022-05-09 at 1 13 15 PM
Screen Shot 2022-05-09 at 1 15 03 PM

josephmcg:
Thats determined under the hood by minisearch. We can adjust the specificity level, but i think the current low level is good.

stavares843:
thanks for clarifying 🔨

types/search/search.ts
```diff
@@ -164,3 +164,14 @@ export type QueryOptions = {
...
   perPage: number
 }
+
+export enum MatchTypesEnum {
+  ID = 'id',
+  CONVERSATION = 'conversation',
+  FROM = 'from',
+  TO = 'to',
+  AT = 'at',
+  READAT = 'readAt',
```
stavares843, line 174: maybe READ_AT would be prettier

josephmcg: fixed

stavares843: 🎉

stavares843 (approved):

PR #3081, opened 2022-05-09
fix(search): fix z index issue for search and call window

josephmcg:
**What this PR does** 📖
- z index issue, place chat search results on top of call info

AP-1487

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/search/result/Result.less
+++ b/components/views/chat/search/result/Result.less
@@ -11,7 +11,7 @@
   width: 80%;
   top: @toolbar-height - 9rem;
   border: @light-border !important;
-  z-index: @base-z-index + 3;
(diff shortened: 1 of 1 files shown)


components/views/chat/search/result/Result.less
```diff
@@ -11,7 +11,7 @@
   width: 80%;
   top: @toolbar-height - 9rem;
   border: @light-border !important;
-  z-index: @base-z-index + 3;
+  z-index: @base-z-index + 13;
```
molimauro, line 14: Hate those magic numbers, we should revisit z-indexes in the app

molimauro (approved):

stavares843 (approved):

PR #3117, opened 2022-05-10
test(cypress): new accounts and unskipping tests

luisecm:
**What this PR does** 📖
- Update cypress tests accounts for all tests due to textile api key change
- Updated all cypress tests to retrieve the recovery seeds from a fixtures file, to remove this data from test specs
- Unskip tests that should be working right now after updates to the application
- Added a new cypress command to get elements and avoid the cypress issue for non-attached to DOM elements. Implemented the usage of this command into the goToConversation cypress command
- Updated expected css values on reaction to chat cypress test
- Removed lines of code no longer used in privacy-page-toggles.js cypress test

None

- Sending this as draft. In case that any test case fails due to slowness, I can add a retry to that test case
- On a next PR, I will create a cypress command to retrieve the recovery seed based on the name of the user, so we don't need to use the same code lines again and again

**Additional comments** 🎤
Too many cypress videos to upload!

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -495,9 +495,15 @@ Cypress.Commands.add('goToConversation', (user) => {
   })
 
   //Find the friend and click on the message button associated
-  cy.get('[data-cy=sidebar-user-name]', { timeout: 30000 })
+  cy.get('[data-cy=sidebar-user-name]', { timeout: 60000 })
     .contains(user)
-    .click()
+    .then(($el) => {
+      cy.getAttached($el).click()
+    })
+
+  /*cy.get('[data-cy=sidebar-user-name]', { timeout: 30000 })
+    .contains(user)
+    .click()*/
 
   // Hide sidebar
   cy.get('[data-cy=hamburger-button]').click()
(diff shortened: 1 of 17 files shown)


cypress/support/commands.js
```diff
@@ -495,9 +495,15 @@ Cypress.Commands.add('goToConversation', (user) => {
...
+  cy.get('[data-cy=sidebar-user-name]', { timeout: 60000 })
     .contains(user)
-    .click()
+    .then(($el) => {
+      cy.getAttached($el).click()
+    })
+
+  /*cy.get('[data-cy=sidebar-user-name]', { timeout: 30000 })
+    .contains(user)
+    .click()*/
```
stavares843, line 506: do we need this commented out? 🔨

stavares843 (approved):
🎉

cypress/support/commands.js
luisecm: It is no longer needed but I forgot to remove it. Doing it now!

stavares843: thanks 💥

Jekrimo (approved):

stavares843 (approved):

PR #3133, opened 2022-05-12
fix(user): prevent profile picture drag

josephmcg:
**What this PR does** 📖
title

AP-1538, AP-1554

**Additional comments** 🎤

### DIFF
--- a/assets/styles/framework/typography.less
+++ b/assets/styles/framework/typography.less
@@ -11,10 +11,3 @@
   white-space: nowrap;
   display: block;
 }
-
-.no-select {
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
-  user-select: none;
-}
--- a/assets/styles/framework/typography.less
+++ b/assets/styles/framework/typography.less
@@ -11,3 +11,10 @@
   white-space: nowrap;
   display: block;
 }
+
+.no-select {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+}
--- a/assets/styles/framework/ui.less
+++ b/assets/styles/framework/ui.less
@@ -61,3 +61,10 @@
 }
 
 @mini-light-border: 1px solid rgba(199, 206, 246, 0.25);
+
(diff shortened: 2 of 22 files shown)


phillsatellite:
Issue was fixed 👍  Just out of curiosity is there anything we can do to avoid being able to highlight everything on the page? 
https://user-images.githubusercontent.com/93608357/168118949-2e1904fe-be41-421d-b7cb-510bf9414c61.mov

josephmcg:
@phillsatellite sure, i removed user:select. I also went ahead and improved our device compatibility. some old browsers/devices still need vendor prefixes for that css rule, so I created a global class which applies all the prefixes

phillsatellite:
@josephmcg thanks for adding that! It worked really well on Chrome and Brave but I was still having some issues getting it to work in Safari. I inspected the element in both Safari and Chrome (where it works) and I saw the same CSS is being applied so the extend psuedo element is doing it's thing, and also verified on MDN both my browsers should support user-select:none
Screen Shot 2022-05-13 at 11 18 11 AM (2)
Screen Shot 2022-05-13 at 11 18 11 AM

josephmcg:
Since thats outside the scope of the original ticket, could we revisit that later?

phillsatellite:
> Since thats outside the scope of the original ticket, could we revisit that later?
> 
> Also, whats the issue? The webkit version is getting applied to safari in your screenshot

Oh Im sorry I forgot to add the actual issue in the comment lol it was strange it said that it was being applied in inspect but when I would click and drag the profile picture it would still highlight everything 

https://user-images.githubusercontent.com/93608357/168403965-d3a897fa-8a10-4403-81f4-106c83807ba9.mov

josephmcg:
I don't believe that's an issue. The user needs to be able to select text inside the chat

Jekrimo (approved):

WanderingHogan (approved):

stavares843 (changes requested):
image

clicking on the profiles no longer works


image

works on dev

stavares843:
issue no longer happens

stavares843 (approved):

PR #3160, opened 2022-05-13
fix(status): position and size of status

AlexZakablukov:
**What this PR does** 📖
fix position and size of status (on the message list on the sidebar it was slightly wrong)

[AP-1462](https://satellite-im.atlassian.net/browse/AP-1462)

**Additional comments** 🎤

### DIFF
--- a/components/ui/User/State/State.less
+++ b/components/ui/User/State/State.less
@@ -4,36 +4,40 @@
   align-items: center;
 
   .circle {
-    mask: radial-gradient(circle 8px at 85.35% 85.35%, #0000 98%, #000) 100%
-      100%/100% 100% no-repeat;
+    mask: radial-gradient(circle 8px at 86% 86%, #0000 98%, #000) 100% 100%/100%
+      100% no-repeat;
     overflow: hidden;
+    &.is-large {
+      mask: radial-gradient(circle 12px at 86% 86%, #0000 98%, #000) 100% 100%/100%
+        100% no-repeat;
+    }
   }
 
   .status {
     position: absolute;
-    left: 85.35%;
-    top: 85.35%;
-    font-size: @text-size;
-    text-shadow: 0 0 3px #000;
-    transform: translate(-50%, -50%);
-    display: inline-flex;
-    align-content: center;
+    left: 86%;
+    top: 86%;
+    border-radius: 50%;
+    width: 12px;
+    height: 12px;
+    margin-top: -6px;
+    margin-left: -6px;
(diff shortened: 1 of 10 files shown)


components/ui/User/State/State.less
```diff
@@ -4,36 +4,40 @@
...
-    mask: radial-gradient(circle 8px at 85.35% 85.35%, #0000 98%, #000) 100%
-      100%/100% 100% no-repeat;
+    mask: radial-gradient(circle 8px at 86% 86%, #0000 98%, #000) 100% 100%/100%
+      100% no-repeat;
     overflow: hidden;
+    &.is-large {
+      mask: radial-gradient(circle 12px at 86% 86%, #0000 98%, #000) 100% 100%/100%
+        100% no-repeat;
+    }
   }
```
josephmcg, line 14: I would probably setup a variable, then change the value of the variable for .is-large. something like this
`  .circle {
    --circle-size: 8px;
    mask: radial-gradient(circle var(--circle-size) at 86% 86%, #0000 98%, #000)
      100% 100%/100% 100% no-repeat;
    overflow: hidden;
    &.is-large {
      --circle-size: 12px;
    }
  }`

components/views/chat/group/Group.html
```diff
@@ -3,7 +3,7 @@
     <UiCircle
       :type="src ? 'image' :'random'"
       :seed="address"
-      :size="35"
+      :size="36"
```
josephmcg, line 6: any place you're passing in the default of 36 can be removed, right?

josephmcg (changes requested):

components/ui/User/State/State.less
AlexZakablukov: great point, thx)

components/views/chat/group/Group.html
AlexZakablukov: ofc, I removed them

AlexZakablukov:
fixed

josephmcg (approved):
great, thanks Alex

stavares843 (approved):

AlexZakablukov:
fixed

phillsatellite (approved):

molimauro:
On non-retina screens i see the problem persisting. I'm fixing it, in the meanwhile let this PR go in since it contains useful minor improvements :)

molimauro (approved):

stavares843 (approved):

PR #3187, opened 2022-05-17
fix(glyph): send glyph to active chat

jasonwoodland:
**What this PR does** 📖
Fixes the target recipient when sending a glyphs so that it is always sent to the active chat

AP-1563, AP-1550, AP-1544

**Additional comments** 🎤


josephmcg (approved):

josephmcg:
this should fix some other glyph related issues as well

PR #3196, opened 2022-05-17
feat(calls): properly unbind all listeners after call closure

aewing:
**What this PR does** 📖
unbinds all listeners from a call on close to prevent the peer from intercepting signaling data on the wrong call

**Additional comments** 🎤


WanderingHogan (approved):

phillsatellite:
Tested: When Hogan and I we're testing we noticed that if you answer in a different way than the person is calling you (if they are video calling you and you pick up with only voice for example) the call will immediately hangup. If you both are voice calling or both are video calling works very well

PR #3206, opened 2022-05-17
fix(notification): clear alert number on open

Jekrimo:
**What this PR does** 📖

Set's a variable to clear out the red number of alerts that are unseen. 

AP-1500

### DIFF
--- a/components/views/navigation/toolbar/Toolbar.html
+++ b/components/views/navigation/toolbar/Toolbar.html
@@ -30,14 +30,14 @@
       @click="toggleAlerts"
     >
       <UiCircle
-        v-if="allUnseenNotifications.length && !alertsOpened"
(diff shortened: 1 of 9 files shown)


josephmcg (changes requested):
Rather than create a new store value for tracking the number of read, I think we should stick with obtaining that value from the length of the unread getter.

How about we mark them all as read on notification open (clearAllNotifications), so that should make the number disappear. Then we can remove this `notification.state !== AlertState.READ` from the alert render condition
components/views/navigation/toolbar/alerts/Alerts.html (line 12).

Then make the 'clear all' button set notifications array to [] (setNotifications)

I wanted to fully test, so i ended up opening a PR here
https://github.com/Satellite-im/Core-PWA/pull/3221

PR #3217, opened 2022-05-18
fix(hounddog): copy houndog logic to store getters

josephmcg:
**What this PR does** 📖
- There were some hounddog related bugs recently, so I decided to take a look at the class. I discovered it's essentially a roundabout way of creating friends store getters
- copy hounddog methods to legit friends store getters
- in a future PR, i would like to refactor/clean up these getters and remove hounddog altogether

AP-1573

- this has 0 new functionality, just prep for future updates and hopefully getting validation that this is a good idea

**Additional comments** 🎤

### DIFF
--- a/store/friends/state.ts
+++ b/store/friends/state.ts
@@ -4,7 +4,7 @@ const InitialFriendsState: FriendsState = {
   incomingRequests: [],
   outgoingRequests: [],
   all: [],
-  activeConversation: null,
+  activeConversation: undefined,
 }
 
 export default () => InitialFriendsState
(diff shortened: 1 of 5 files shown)


store/friends/state.ts
```diff
@@ -4,7 +4,7 @@ const InitialFriendsState: FriendsState = {
   incomingRequests: [],
   outgoingRequests: [],
   all: [],
-  activeConversation: null,
+  activeConversation: undefined,
```
molimauro, line 7: why 'undefined' instead of `null`?

josephmcg: Ah i just did that to make the type declaration easier (?: type vs type | null). We can change it back if it makes a difference

Jekrimo (approved):

molimauro (approved):

phillsatellite (approved):

stavares843 (approved):

PR #3218, opened 2022-05-18
feat(sounds): combine deafen/undeafen into one toggle control

jasonwoodland:
**What this PR does** 📖
- Removes the undeafen sound toggle in the audio & video settings.
- Adds a `toggleSounds` method which accepts an array of `Sounds` to set the value
- Combines the deafen/undeafen sounds into one toggle control

AP-1537

**Additional comments** 🎤

### DIFF
--- a/components/views/settings/pages/audio/sounds/Sounds.html
+++ b/components/views/settings/pages/audio/sounds/Sounds.html
@@ -15,7 +15,7 @@
   />
   <SettingsPagesAudioSoundsItem
     :value="sound(Sounds.MUTE)"
-    @toggle="toggleSound(Sounds.MUTE, $event)"
+    @toggle="toggleSounds([Sounds.MUTE, Sounds.UNMUTE], $event)"
     :title="$t('pages.settings.notifications.sounds.mute')"
   />
   <SettingsPagesAudioSoundsItem
--- a/components/views/settings/pages/audio/sounds/Sounds.html
+++ b/components/views/settings/pages/audio/sounds/Sounds.html
@@ -20,13 +20,8 @@
   />
   <SettingsPagesAudioSoundsItem
     :value="sound(Sounds.DEAFEN)"
-    @toggle="toggleSound(Sounds.DEAFEN, $event)"
+    @toggle="toggleSounds([Sounds.DEAFEN, Sounds.UNDEAFEN], $event)"
     :title="$t('pages.settings.notifications.sounds.deafen')"
-  />
-  <SettingsPagesAudioSoundsItem
-    :value="sound(Sounds.UNDEAFEN)"
-    @toggle="toggleSound(Sounds.UNDEAFEN, $event)"
-    :title="$t('pages.settings.notifications.sounds.undeafen')"
   />
   <SettingsPagesAudioSoundsItem
     :value="sound(Sounds.UPLOAD)"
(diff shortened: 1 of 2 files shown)


components/views/settings/pages/audio/sounds/Sounds.vue
```diff
@@ -24,6 +24,11 @@ export default Vue.extend({
     toggleSound(key: Sounds, value: boolean) {
       this.$store.commit('sounds/set', { key, value })
     },
+    toggleSounds(keys: Sounds[], value: boolean) {
+      for (const key of keys) {
+        this.$store.commit('sounds/set', { key, value })
+      }
```
Jekrimo, line 30: can we use a forEach() here? lika:

keys.forEach((key) => this.$store.commit('sounds/set', { key, value }))

Jekrimo (changes requested):

components/views/settings/pages/audio/sounds/Sounds.vue
jasonwoodland: Hey @Jekrimo, nice to meet you! Can you explain the benefit of `forEach` in this scenario?
The for...of statement iterates over iterable objects, including arrays in this case.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

josephmcg: for of is good here, it also allows for object shorthand in the mutation params. this page provides some good distinctions between the various iterators https://thecodebarbarian.com/for-vs-for-each-vs-for-in-vs-for-of-in-javascript.html

marking as resovled

josephmcg (changes requested):
could you update the mute settings in the same way? it still has a sound effect when you toggle mute `on`.

Unrelated, but I was confused by this settings section at first. I think a sentence explaining that these are sound effects settings would be helpful. probably better in a separate ticket though

jasonwoodland:
@josephmcg combined unmute into mute.
I think the section title could change, maybe just "Notifications"?
If there's a paragraph you can write up I'll add it in another ticket.

components/views/settings/pages/audio/sounds/Sounds.html
```diff
@@ -15,19 +15,14 @@
   />
   <SettingsPagesAudioSoundsItem
     :value="sound(Sounds.MUTE)"
-    @toggle="toggleSound(Sounds.MUTE, $event)"
+    @toggle="toggleSound([Sounds.MUTE, Sounds.UNMUTE], $event)"
```
josephmcg, line 18: toggleSound**s**

josephmcg (changes requested):

josephmcg (approved):

molimauro:
> could you update the mute settings in the same way? it still has a sound effect when you toggle mute `on`.
> 
> Unrelated, but I was confused by this settings section at first. I think a sentence explaining that these are sound effects settings would be helpful. A user could easily think that these settings control the actual mute/deafen. probably better in a separate ticket though

I'm confused too honestly. Why should be combine mute/unmute and deafen/undeafen? I was reading those settings as: If i have the switch on for mute and off for unmute it means sound will be played once i mute but not when i unmute. So i would have add here only the unmute switch and nothing else. Or am i wrong? What do you guys think? :)

josephmcg:
@molimauro the whole concept just seems strange to me. I think audio indicators for significant setting changes (mute, video) are a good thing. I can't imagine wanting to disable them from a user perspective

molimauro:
i talked with Liz and it's ok to combine deafen/undeafen for now for keeping things simple. If users will complain in the future we'll split them ;) Thanks for your patience!

molimauro (approved):

stavares843:
thanks @molimauro 🔨

stavares843:
added ticket [here](https://satellite-im.atlassian.net/browse/AP-1619) 🔨

stavares843 (approved):

PR #3222, opened 2022-05-18
feat(friends): autofocus add friend input

jasonwoodland:
**What this PR does** 📖
- Adds an `autofocus` prop to the InputGroup component
- Autofocuses the add friend input field

AP-1572

**Additional comments** 🎤

### DIFF
--- a/components/views/friends/add/Add.html
+++ b/components/views/friends/add/Add.html
@@ -10,6 +10,7 @@
     @update="_searchFriend"
     v-model="accountID"
     :loading="searching"
+    :autofocus="hasNoFriends"
   >
     <user-plus-icon size="1x" />
   </InteractablesInputGroup>
--- a/components/views/friends/add/Add.html
+++ b/components/views/friends/add/Add.html
@@ -10,7 +10,7 @@
     @update="_searchFriend"
     v-model="accountID"
     :loading="searching"
-    :autofocus="hasNoFriends"
+    autofocus
   >
     <user-plus-icon size="1x" />
   </InteractablesInputGroup>
--- a/components/views/friends/add/Add.vue
+++ b/components/views/friends/add/Add.vue
@@ -35,9 +35,6 @@ export default Vue.extend({
     friendInviteUrl(): string {
       return `${location.origin}/#/friends/list/${this.accounts.active}`
     },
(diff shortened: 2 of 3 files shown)


components/views/friends/add/Add.vue
```diff
@@ -35,6 +35,9 @@ export default Vue.extend({
     friendInviteUrl(): string {
       return `${location.origin}/#/friends/list/${this.accounts.active}`
     },
+    hasNoFriends() {
+      return this.$store.state.friends.all.length === 0
+    },
```
josephmcg, line 40: 1. could you add typescript return doc `hasNoFriends(): boolean`? if any computed are missing return types, the context of `this` breaks in the file. plus documentation is good :)
2. this works, but we typically use mapState when accessing store. could you update the file like this?
`...mapState({
      activeAccount: (state) => (state as RootState).accounts.active,
      allFriends: (state) => (state as RootState).friends.all,
    }),`

name the mapped variables however you like, this was just an example

josephmcg (changes requested):

components/views/friends/add/Add.vue
josephmcg: also from a ux perspective I actually like this being autofocused regardless of how many friends the user has. if i were to click this page, it would most likely be to add a friend (just paste, rather than click and paste)

jasonwoodland: updated to autofocus unconditionally!

josephmcg (approved):

components/views/friends/add/Add.html
```diff
@@ -10,6 +10,7 @@
     @update="_searchFriend"
     v-model="accountID"
     :loading="searching"
+    autofocus
```
josephmcg, line 13: thought about this some more. could we autofocus true on desktop, false for mobile?
you can use `$device.isMobile`
the virtual keyboard coming up everytime would be kind of annoying

josephmcg: nvm, mobile layout doesn't render this component

josephmcg (approved):

stavares843:
sounds good, thanks 🔨

stavares843 (approved):

PR #3227, opened 2022-05-18
refactor(tooltip): changed and improved tooltips component

ThomBos:
**What this PR does** 📖
Improves tooltip performances and display and makes tooltips code easier to change

AP-1595

**Additional comments** 🎤


molimauro:
I saw the chatbar controls does not have the `cursor: pointer` attr ;)

molimauro:
Hey @ThomBos thanks! Now i have only 2 doubts, one is regarding the appearing effect of the tooltip, and the other is the arrow of the tooltip that we didn't have previously. Can you talk with Liz about these and let us know? Thanks!

ThomBos:
> Hey @ThomBos thanks! Now i have only 2 doubts, one is regarding the appearing effect of the tooltip, and the other is the arrow of the tooltip that we didn't have previously. Can you talk with Liz about these and let us know if we have to keep them? Thanks!

I had a talk with Liz, so we are now keeping both animations and arrows!! 🔥

components/interactables/Button/Button.html
```diff
@@ -5,8 +5,7 @@
   :type="htmlType"
   ref="button"
 >
-  <!-- {{$attrs['data-tooltip']}} -->
-  <UiGlowingCursorArea :hide="$attrs['data-tooltip'] ? true : false">
+  <UiGlowingCursorArea :hide="$attrs['data-popper-shown'] ? true : false">
```
aewing, line 8: I think you could shorten this with `:hide="!!$attrs['data-popper-shown']"`

assets/styles/base.less
```diff
@@ -709,24 +709,63 @@ video {
...
-    &:extend(.blur-less);
-    font-family: @primary-font !important;
-    color: @secondary-text !important;
-    &:extend(.fourth-layer);
+li .is-active,
+li.active,
+input.date-input {
+  &:extend(.background-flair);
+  &:extend(.glow-flair);
+  border: none !important;
```
aewing, line 720: This CSS rule is a little concerning at first glance. Are we sure that we want to style any item that has the `is-active` class inside of an `li`, any `li` with the `active class`, and all `date-input` with a `border: none !important`?

assets/styles/base.less
```diff
@@ -709,24 +709,63 @@ video {
...
+}
+
+// ------------- Ui Elements-------------
+
+* {
+  border-color: var(--flair-color);
```
aewing, line 726: This feels a little far-reaching also

components/interactables/Button/Button.html
ThomBos: Done! I didn't know it was possible to do that, thanks 😄

assets/styles/base.less
molimauro: Totally agree. @ThomBos why this rule in this PR?

ThomBos: this isn't on this pr, i removed the class above ".has-tooltip"

molimauro: ok ill open a ticket for this one and the other and we'll address those there

molimauro: https://satellite-im.atlassian.net/browse/AP-1635

molimauro (approved):

stavares843 (approved):

PR #3236, opened 2022-05-19
fix(sidebar): fix toggle mute dispatch

jasonwoodland:
**What this PR does** 📖
- update `$store.dispatch` calls for toggling audio/video muting in sidebar so that the buttons behave correctly.
- fix audio stream indicator visibility logic in media/user component so it only shows the audio stream indictor when unmuted.

AP-1581, AP-1578, AP-1466

**Additional comments** 🎤

### DIFF
--- a/components/views/media/actions/Actions.vue
+++ b/components/views/media/actions/Actions.vue
@@ -37,13 +37,13 @@ export default Vue.extend({
   computed: {
     ...mapState(['audio', 'video', 'webrtc']),
     audioMuted(): boolean {
-      return this.webrtc.streamMuted[p2p.id]?.audio
+      return this.audio.muted
     },
     videoMuted(): boolean {
-      return this.webrtc.streamMuted[p2p.id]?.video
+      return p2p.id && this.webrtc.streamMuted[p2p.id]?.video
     },
     screenMuted(): boolean {
-      return this.webrtc.streamMuted[p2p.id]?.screen
+      return p2p.id && this.webrtc.streamMuted[p2p.id]?.screen
     },
   },
   methods: {
@@ -54,11 +54,15 @@ export default Vue.extend({
      */
     toggleMute(kind = 'audio') {
       this.isLoading = true
-      this.$store.dispatch(
-        'webrtc/toggleMute',
-        { kind, peerId: p2p.id },
-        { root: true },
-      )
+      if (kind === 'audio') {
+        this.$store.dispatch('audio/toggleMute', {}, { root: true })
+      } else {
+        this.$store.dispatch(
+          'webrtc/toggleMute',
+          { kind, peerId: p2p.id },
+          { root: true },
+        )
+      }
       this.isLoading = false
     },
     /**
--- a/components/views/media/actions/Actions.vue
+++ b/components/views/media/actions/Actions.vue
@@ -1,4 +1,4 @@
-<template src="./Actions.html" />
+<template src="./Actions.html"></template>
 
 <script lang="ts">
 import Vue from 'vue'
--- a/components/views/media/actions/Actions.vue
+++ b/components/views/media/actions/Actions.vue
@@ -17,6 +17,7 @@ import { mapState } from 'vuex'
 import { Sounds } from '~/libraries/SoundManager/SoundManager'
 import { WebRTCEnum } from '~/libraries/Enums/enums'
 import { Peer2Peer } from '~/libraries/WebRTC/Libp2p'
+import { PeerMutedState } from '~/store/webrtc/types'
 const p2p = Peer2Peer.getInstance()
 
 export default Vue.extend({
@@ -52,7 +53,7 @@ export default Vue.extend({
      * @description Toggles mute for outgoing audio
      * @example
      */
-    toggleMute(kind = 'audio') {
+    toggleMute(kind: keyof PeerMutedState) {
       this.isLoading = true
       if (kind === 'audio') {
         this.$store.dispatch('audio/toggleMute', {}, { root: true })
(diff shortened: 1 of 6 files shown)


Jekrimo (approved):

phillsatellite:
Tested: Found a crash when trying to make a phone call Browsers used were Chrome and Brave 🚀 
https://user-images.githubusercontent.com/93608357/169329424-97b31c69-bfe5-42f8-9802-f4fcf69ecb17.mov

jasonwoodland:
@phillsatellite thanks for spotting it! Fixed in last commit

stavares843:
thanks @jasonwoodland 🔨

phillsatellite:
Retested: Crash is gone! Functionality was good on my end the only differences I noticed from this branch and dev is the sound feedback that happens when the user hits mute or video

https://user-images.githubusercontent.com/93608357/169571403-b4897233-06cb-49aa-a95a-87d6fcad55c2.mov

components/views/media/user/User.html
```diff
@@ -28,7 +28,7 @@
       type="random"
       :seed="user.address"
       :size="$device.isMobile ? 36 : 65"
-      :class="`${!audioStream ? 'calling' : ''}`"
+      :class="`${audioStream ? 'calling' : ''}`"
```
josephmcg, line 31: this has some unnecessary wrappers
`:class='{calling : audioStream}'`

josephmcg (changes requested):
mic and video toggles no longer work
image

components/views/media/user/User.html
jasonwoodland: fixed the class

jasonwoodland: I can't reproduce the bug you're getting with the toggle buttons

stavares843:
rebase wasn't working, added latest dev here to update Cypress

phillsatellite (approved):

Clayton215 (approved):

josephmcg (commented):
The mute/video toggles don't work while you are not in a call. If this is intended functionality, then this is ok

jasonwoodland:
@stavares843 I made some other changes to sync the muted state in the sidebar with the active calls. (see [234b7b7](https://github.com/Satellite-im/Core-PWA/pull/3236/commits/234b7b7dc533f0d7516544f0e275953431b66f79)) So please re-review!

store/audio/actions.ts
```diff
@@ -7,25 +7,24 @@ export default {
...
+      root: true,
+    })
+
+    if (call) {
+      if (muted) {
+        call.mute({ kind: 'audio' })
+      } else {
+        call.unmute({ kind: 'audio' })
+      }
     }
```
josephmcg, line 27: we typically use guard clauses, could you rewrite?
`if (!call) {
        return
      }
      if (muted) {
        call.mute({ kind: 'audio' })
        return
      }
      call.unmute({ kind: 'audio' })
`

components/views/media/actions/Actions.vue
```diff
@@ -54,11 +54,15 @@ export default Vue.extend({
      */
     toggleMute(kind = 'audio') {
```
josephmcg, line 55: this automatic value is never used. could we use keyof PeerMutedState for kind param

components/views/navigation/sidebar/controls/Controls.vue
```diff
@@ -51,17 +49,9 @@ export default Vue.extend({
      * @example
      */
     async toggleMute(kind = 'audio') {
```
josephmcg, line 51: could do same improvement here for types. also change self closing template at the top

josephmcg (changes requested):
change the first line of Actions.vue to this. I've noticed linting breaks if the template is self closing
image

store/audio/actions.ts
josephmcg: double check your implementation of this. we want to eliminate else. we also want to return if `!call`, right?

josephmcg (changes requested):
check previous comment

jasonwoodland:
@josephmcg fixed!

store/audio/actions.ts
jasonwoodland: my bad, fixed

store/audio/actions.ts
```diff
@@ -7,25 +7,25 @@ export default {
...
+    if (!call) {
       return
     }
-    if (call) call.mute({ kind: 'audio' })
-    dispatch('sounds/playSound', Sounds.MUTE, { root: true })
+
+    if (muted) {
+      call.mute({ kind: 'audio' })
+    }
+    call.unmute({ kind: 'audio' })
```
josephmcg, line 28: guard clause still funky, missing return

josephmcg (approved):

stavares843 (changes requested):
the video toggle doesn't work if you are not on a call, works on dev

Captura de ecrã 2022-05-26, às 18 48 13


btw I noticed when hovering it says ''Toggle'' but that issue is on dev as well so I added a ticket [here](https://satellite-im.atlassian.net/browse/AP-1637)

WanderingHogan:
@stavares843 I don't think we actually want to allow the video toggle if you are not on a call since we are forcing all calls to start as audio. I think another ticket to disable the button when not on a call might be warranted, what do you think?

stavares843:
sounds good, thanks @WanderingHogan 🔨

stavares843:
added ticket [here](https://satellite-im.atlassian.net/browse/AP-1647)

stavares843 (approved):

PR #3244, opened 2022-05-20
feat(call): add permission error alert for screen sharing

jasonwoodland:
**What this PR does** 📖
Adds a notification if the browser doesn't have permission to screen share

AP-1545

**Additional comments** 🎤

### DIFF
--- a/libraries/WebRTC/Call.ts
+++ b/libraries/WebRTC/Call.ts
@@ -6,6 +6,8 @@ import { Peer2Peer } from './Libp2p'
 import { WebRTCErrors } from './errors/Errors'
 import { Config } from '~/config'
 
+const PERMISSION_DENIED_BY_SYSTEM = 'Permission denied by system'
+
 export class CallPeer extends Peer {
   id: string
   name: string
@@ -404,7 +406,13 @@ export class Call extends Emitter<CallEventListeners> {
     try {
       screenStream = await navigator.mediaDevices.getDisplayMedia()
     } catch (e) {
-      throw new Error(WebRTCErrors.PERMISSION_DENIED)
+      if (
+        e instanceof DOMException &&
+        e.message.includes(PERMISSION_DENIED_BY_SYSTEM)
+      ) {
+        throw new Error(WebRTCErrors.PERMISSION_DENIED)
+      }
+      return
     }
 
     if (!this.streams[this.localId]) {
@@ -796,9 +804,7 @@ export class Call extends Emitter<CallEventListeners> {
 
     const stream = this.streams[peerId]?.[kind]
     if (!stream) {
-      throw new Error(
-        `webrtc/call: unmute - ${kind} stream not initialized for peer: ${peerId}`,
-      )
+      return
     }
 
     const track: MediaStreamTrack =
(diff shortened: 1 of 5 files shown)


libraries/WebRTC/Call.ts
```diff
@@ -401,7 +401,17 @@ export class Call extends Emitter<CallEventListeners> {
       throw new Error('WebRTC display media not supported')
     }
 
-    const screenStream = await navigator.mediaDevices.getDisplayMedia()
+    let screenStream: MediaStream
+
+    try {
+      screenStream = await navigator.mediaDevices.getDisplayMedia()
+    } catch (e) {
+      alert(
```
stavares843, line 409: can you add this on [locales](https://github.com/Satellite-im/Core-PWA/blob/dev/locales/en-US.js) pls? 🔨

stavares843 (changes requested):

libraries/WebRTC/Call.ts
jasonwoodland: @stavares843 done! Now it shows an error toast instead of a native alert too.

libraries/WebRTC/Call.ts
```diff
@@ -395,13 +395,17 @@ export class Call extends Emitter<CallEventListeners> {
...
-      throw new Error('WebRTC display media not supported')
+      throw new Error('errors.webRTC.permission_denied')
+    }
+
+    let screenStream: MediaStream
+
+    try {
+      screenStream = await navigator.mediaDevices.getDisplayMedia()
+    } catch (e) {
+      throw new Error('errors.webRTC.permission_denied')
```
josephmcg, line 406: since we're throwing the same string, could you make an enum for this? either create a new error file in the same folder, or add the enum to this class.

there are other error messages that should be added to that enum/translated too, but I understand how difficult it can be to track all of them down. We can do those in a separate ticket

josephmcg (approved):

stavares843:
thanks @phillsatellite 🔨

jasonwoodland:
@phillsatellite @stavares843 I tried with opera and opera GX and both crashed. It was crashing on dev and this branch, so I think it's an opera-specific issue.

https://user-images.githubusercontent.com/5356544/170407811-64c0df1b-bb8e-419a-9f16-293cfe42d078.mov

josephmcg:
good catch Phill, thanks for your thorough testing. While overall support would be wonderful, Opera is ~2% of market share at the moment. If it works on the major browsers, I would suggesting merging this and creating a separate ticket for Opera support

stavares843:
thanks @phillsatellite for testing 🔨 

and thanks for checking @josephmcg @jasonwoodland 💥 if this is opera only, I agree with joe as well in fixing separately

WanderingHogan (approved):

stavares843 (changes requested):
https://user-images.githubusercontent.com/29093946/171044281-f48297a6-c055-4ce8-8166-e4ede7379b81.mov

this branch vs dev

we don't need to trigger the permissions denied error when the user just cancelled the share

josephmcg (approved):

stavares843 (approved):

PR #3246, opened 2022-05-20
refactor(friends): move friend method from hounddog into getters

jasonwoodland:
**What this PR does** 📖
- Replace $hounddog calls inside components with vuex store getters
- improve action argument typing. will be expanded to other namespaces later

AP-1600

- maybe these getters can be refactored, but that can happen at a later date. tried to keep this update simple

**Additional comments** 🎤


Jekrimo (approved):

store/friends/actions.ts
```diff
@@ -164,7 +169,7 @@ export default {
       ).bytes,
     )
 
-    const friendExists = $Hounddog.friendExists(state, friendKey)
+    const friendExists = getters.friendExists(state)(friendKey)
```
WanderingHogan, line 172: this ()() is weird but i like it

josephmcg:
Jason and I collaborated on this one. I plan on building better types for other namespace getters separately

store/friends/actions.ts
josephmcg: I changed this up since the getter doesn't require any state based arguments, it can grab all that on it's own

phillsatellite (approved):

components/views/files/upload/Upload.vue
```diff
@@ -2,7 +2,7 @@
 <script lang="ts">
 import { FilePlusIcon, PlusIcon } from 'satellite-lucide-icons'
 import Vue, { PropType } from 'vue'
-import { mapState } from 'vuex'
+import { mapState, mapGetters } from 'vuex'
```
WanderingHogan, line 5: it looks like you took out an un-used computed item but didn't need to replace it with the getters you added?

WanderingHogan (approved):

components/views/files/upload/Upload.vue
josephmcg: it was no longer being used in the component. I think it was a leftover from when Samir did a larger refactor of these components

stavares843 (approved):

PR #3260, opened 2022-05-20
chore(imports): remove not used imports

stavares843:
**What this PR does** 📖

chore(imports): remove not used imports

### DIFF
--- a/libraries/SatelliteDB/SatelliteDB.ts
+++ b/libraries/SatelliteDB/SatelliteDB.ts
@@ -1,4 +1,4 @@
-import { Dexie, IndexableType } from 'dexie'
+import { Dexie } from 'dexie'
 import SearchIndex from './SearchIndex'
 import { Message } from '~/types/textile/mailbox'
 import { Config } from '~/config'
--- a/libraries/SatelliteDB/SatelliteDB.ts
+++ b/libraries/SatelliteDB/SatelliteDB.ts
@@ -1,4 +1,4 @@
-import { Dexie } from 'dexie'
+import { Dexie, IndexableType } from 'dexie'
 import SearchIndex from './SearchIndex'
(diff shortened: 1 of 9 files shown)


libraries/SatelliteDB/SatelliteDB.ts
```diff
@@ -1,4 +1,4 @@
-import { Dexie, IndexableType } from 'dexie'
```
WanderingHogan, line 1: This gets used as a type below

stavares843: Captura de ecrã 2022-05-22, às 18 27 20

shady

josephmcg: if it's only in jsdoc, it doesn't appear as 'used'

stavares843: so it's not really used and jsdocs are also wrong right? lol

josephmcg: the jsdocs are correct. in a perfect world, we would annotate return type with **both** jsdoc and typescript return annotation. this is just speculation, but I'm assuming the typescript compiler/linter/whatever can't find it because jsdoc is a comment

stavares843: so should this be reverted or not really? 🔨

josephmcg: this specific removal should be reverted since upsert is returning that type (as mentioned in jsdoc). if you notice something with jsdoc and no typescript return annotation, vscode will offer to do it for you.
image

All of these updates should be double checked and keep the imports that are mentioned in jsdoc (even better, copy it to ts return annotation as well)

phillsatellite (approved):

PR #3285, opened 2022-05-24
feat(buckets): upgrade abstract interfaces, switch chat to use SharedBucket class

josephmcg:
**What this PR does** 📖
- improve RFM interface and create abstract Bucket class
  - PersonalBucket for personal, encrypted files
  - SharedBucket for unencrypted files such as profile picture, chat files
  - I considered combining them and adding various return types, but it seems sloppy. this has better type safety and will be easier to scale as the buckets diverge with more features
- remove legacy BucketManager
- add bucket path (id) to FileMessage payload. We were previously using file name for the path. However, if a new file with the same name is uploaded, it will overwrite the old file. using uuid for path will ensure this doesn't happen
- switch chat to use streamsaver for downloads rather than blobs
- move filesystem import into PersonalBucket init rather than account actions.ts

AP-1607

- There is currently no index to keep track of your shared uploads, that will be a future ticket
- chat upload may behave strangely for large files still. that component needs additional updates, but I wanted to handle it separately. please ensure that small file uploads work in chat
- message component still keeps track of image blobs, but doesn't use them for download. will be ripped out in future update
- file save/download in chat will only work for newly uploaded files. previously uploaded files don't have id, so it won't be able to find the path

**Additional comments** 🎤


libraries/Files/remote/textile/SharedBucket.ts
```diff
@@ -0,0 +1,69 @@
...
+   * @description Initializes bucket
+   * @param name bucket name
+   */
+  async init({ name }: { name: string }) {
+    await this.getBucket({ name })
+    if (!this._buckets || !this._key) {
+      throw new Error(TextileError.BUCKET_NOT_INITIALIZED)
+    }
+    Vue.prototype.$Logger.log('Shared Bucket', 'Initialized')
+    // todo - add index tracking capabilities to shared bucket
```
Jekrimo, line 32: Can we delete commented?

Jekrimo (approved):

libraries/Files/remote/textile/SharedBucket.ts
josephmcg: Im going to work on it soon, so i'd prefer to leave it

Jekrimo: Just checkin!

phillsatellite:
Tested and found some issues that I haven't been able to replicate on dev. I got stopped short from Solana but I'll continue once its back up 

When in a chat, each user will have the same color profile avatar 

same color avatars

Other issue was I'm not seeing replies or reaction appear on other users side. I thought this might be textile but I wasn't able to replicate it on dev branch 

missing reply

josephmcg:
the avatar color was resolved in a separate PR, should be ok since Sara rebased this branch.
I'm not sure what's going on with the missing replies, I didn't alter anything that would affect those. I'll take a look once I'm able to test

libraries/Textile/TextileManager.ts
```diff
@@ -72,16 +72,16 @@ export default class TextileManager {
...
+    })
 
-    // Initialize bucket
-    this.bucket = new Bucket(textile)
-    await this.bucket.init(Config.textile.bucketName)
+    this.personalBucket = new PersonalBucket(textile)
+    await this.personalBucket.init({
+      name: Config.textile.personalBucket,
+      encrypted: true,
+    })
```
molimauro, line 84: Is there a reason why we await the sharedBucket init and then call the personalBucket one?

josephmcg: Good point, no reason to await. Feel free to fix that if you wanna clear up blocks. Otherwise I can change it tomorrow

stavares843:
the missing replies seems just the textile listener issue getting inactive and not related to this

phillsatellite (approved):

WanderingHogan (approved):

stavares843 (approved):

PR #3317, opened 2022-05-27
fix(message): disable save on empty message

jasonwoodland:
**What this PR does** 📖
- Prevents saving an edited message as an empty message
- Fixes styling and locale strings in the edit message UI

AP-1631

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/message/edit/Edit.vue
+++ b/components/views/chat/message/edit/Edit.vue
@@ -43,7 +43,8 @@ export default Vue.extend({
   },
   methods: {
     saveMessage() {
-      if (this.content.length === 0) {
+      if (this.content.trim().length === 0) {
+        this.$toast.error(this.$t('errors.chat.empty_message_error') as string)
         return
       }
       this.$emit('commitMessage', this.content.slice(0, this.maxChars))
--- a/components/views/chat/message/edit/Edit.vue
+++ b/components/views/chat/message/edit/Edit.vue
@@ -10,15 +10,6 @@ import { Config } from '~/config'
 import { KeybindingEnum } from '~/libraries/Enums/enums'
 import Editable from '~/components/views/chat/chatbar/Editable.vue'
 
-declare module 'vue/types/vue' {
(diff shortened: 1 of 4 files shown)


josephmcg (changes requested):
I like the solution. Could you remove `$data` and `$props` mentions and fix intellisense in the file? The computed need return types. Then you can remove the declaration stuff (screenshot below).

Please make a habit of fixing this whenever you see it

image

components/views/chat/message/edit/Edit.vue
```diff
@@ -40,19 +31,22 @@ export default Vue.extend({
...
-    this.$data.content = this.$props.message
+    this.content = this.message
   },
   methods: {
     saveMessage() {
-      this.$emit('commitMessage', this.$data.content.slice(0, this.maxChars))
+      if (this.content.length === 0) {
+        return
+      }
+      this.$emit('commitMessage', this.content.slice(0, this.maxChars))
```
WanderingHogan, line 49: why not just `if (this.content.length > 0) this.$emit('commitMessage', this.content.slice(0, this.maxChars))`

WanderingHogan (approved):

components/views/chat/message/edit/Edit.vue
josephmcg: I believe the current way falls more in line with our 'guard clause' philosophy. although I'm lazy, I would write `!this.content.length`

josephmcg (approved):

stavares843 (changes requested):
https://user-images.githubusercontent.com/29093946/171073731-b87a9f08-67de-42c3-828e-f1487283bf13.mov

if you try to save a empty message doesn't return any feedback to the user when clicking on save

josephmcg:
true. error toast would be enough.
a nice bonus feature would be making the blue outline red if error, that would be fairly involved though

stavares843 (approved):

PR #3350, opened 2022-05-30
fix(call): mute button logic

jasonwoodland:
**What this PR does** 📖
- Synchronises mute button with webrtc call state

AP-1647

**Additional comments** 🎤


components/views/media/actions/Actions.vue
```diff
@@ -56,7 +56,9 @@ export default Vue.extend({
     toggleMute(kind: keyof PeerMutedState) {
       this.isLoading = true
       if (kind === 'audio') {
-        this.$store.dispatch('audio/toggleMute', {}, { root: true })
+        this.$store.dispatch('audio/toggleMute', undefined, { root: true })
+      } else if (kind === 'video') {
```
Jekrimo, line 60: Can you make an Enum for 'video' and 'audio' if one does not exist yet? I feel like we have one, but def will be useful if we do not

phillsatellite (approved):

stavares843:
Captura de ecrã 2022-06-01, às 17 32 52

can we add a tooltip saying ''Not available outside a call'' or something similar? this way adds more feedback to the user

josephmcg:
im seriously questioning whether we need this button at all if it doesn't work outside of a call. Could we just remove it?

WanderingHogan (approved):

WanderingHogan:
@josephmcg sounds like a good idea to me, hide if not in call, show if in call. If you want to make a ticket and come back to it that works, it's very low priority right now compared to functionality things that need done, though, so mark it with POST

josephmcg:
@WanderingHogan that makes life much easier, thanks.

stavares843 (approved):

PR #3383, opened 2022-05-31
feat(notifications): Implement notifications for Group Chats

Jekrimo:
**What this PR does** 📖

Implement notifications for Group Chats

AP-1501


components/views/navigation/toolbar/alerts/alert/Alert.vue
```diff
@@ -17,14 +17,21 @@ export default Vue.extend({
   },
   computed: {
     ...mapState({ ui: (state) => (state as RootState).ui }),
+    ...mapGetters('accounts', ['getActiveAccount']),
```
WanderingHogan, line 20: I don't see where you're actually using this getter here or in the template, or am I missing something?

josephmcg: I can't see where `ui` is being used as well

Jekrimo: got it, nice catch thanks!

libraries/ui/Alerts.ts
```diff
@@ -23,7 +23,10 @@ export type Alert = {
   id?: string
   state?: AlertState
   type: AlertType
-  from: string
+  fromName: string
+  fromAddress?: string
+  group?: string
```
aewing, line 28: could this be `groupName` just to be consistent with the other signatures?

aewing (approved):

phillsatellite:
Hi Jeff! Awesome pr 💯  Few small things I found though 

1. Once each user has sent a message in the group chat, whenever any future alerts from groups come in they will come in twice as much so like if User A sent the group a message User B would get 2 of the same alerts saying User A messaged the group. The easiest way to replicate is start a new group, send 1 message from each user after that the next message you send you should get x2 alerts

https://user-images.githubusercontent.com/93608357/171700035-9cbeb11e-69ab-4130-b9fd-0f0ef7d1b829.mov


2. Other issue I found was if a user hasn't entered the group chat yet they won't receive any alerts from it. They have to go into the group chat then navigate to another chat to start receiving notifications for it. I tested dev's behavior with how it is if you have a 1on1 chat you haven't entered yet and was still getting alerts for it

https://user-images.githubusercontent.com/93608357/171700598-790b7e0d-df5d-46b4-9070-5567b469e7ca.mov

store/textile/actions.ts
```diff
@@ -731,6 +730,9 @@ export default {
     if (!$TextileManager.groupChatManager?.isInitialized()) {
       throw new Error(TextileError.EDIT_HOT_KEY_ERROR)
     }
+    if ($TextileManager.groupChatManager?.isSubscribed()) {
+      return
+    }
```
molimauro, line 735: Sorry @Jekrimo probably im missing some pieces. How is this supposed to work with 2 different groups? I was expecting something like `isSubscribed(groupId)` otherwise this will sub only on the first group or am i wrong?

Jekrimo: I lied, it's not workin - workin on it now

molimauro: great!

phillsatellite:
Tested: I think something broke since the last changes no notifications from groups seem to appear. I also tested this with Sara to see if it would appear on her side and nothing 😢 

https://user-images.githubusercontent.com/93608357/172469569-d8d736e8-b0c4-4549-b321-0fa437627c3c.mov

stavares843:
true 🔨 i also didn't get the group notifications

Jekrimo:
> Tested: I think something broke since the last changes no notifications from groups seem to appear. I also tested this with Sara to see if it would appear on her side and nothing 😢
> 
>  no.notifications.mov

pushed fix. PRAY FOR MEEEE

phillsatellite (approved):

stavares843 (approved):

PR #3395, opened 2022-06-01
feat(video): option to mirror local video stream

jasonwoodland:
**What this PR does** 📖
- Adds textile record field for `flipVideo` boolean
- Adds flip video switch to audio/video settings page
- sets scaleX(-1) on local video stream if flipVideo is enabled

AP-1574

**Additional comments** 🎤


components/views/media/user/User.less
```diff
@@ -1,4 +1,4 @@
-.user {
+true {
   position: relative;
 }
```
josephmcg, line 3: what's going on here?

components/views/media/user/User.html
```diff
@@ -14,9 +14,10 @@
     ></video>
   </div>
 
-  <div v-else-if="videoStream" class="video-stream-container">
+  <div v-else-if="true" class="video-stream-container">
```
josephmcg, line 17: `v-else-if="true"`?

components/views/settings/pages/audio/Audio.html
```diff
@@ -168,6 +168,17 @@
...
+    <SettingsUnit
+      v-if="getInitialized"
+      :title="$t('pages.settings.video.flipVideo.title')"
+      :text="$t('pages.settings.video.flipVideo.subtitle')"
+    >
+      <InteractablesSwitch
+        v-model="flipVideo"
+        :isLocked="loading.includes('flipVideo')"
+      />
     </SettingsUnit>
```
josephmcg, line 182: show a loader while textile is not initialized

josephmcg (changes requested):
a couple things. could you also clean up the `any` return types to be more specific?

phillsatellite (approved):

molimauro:
The code looks great good job!
One question, why are we saving the `flipVideo` in textile? :)

josephmcg:
I thought it would be good to add our settings somewhere remote so they're the same across app instances (different browser/device, same account)

molimauro (approved):

molimauro:
@josephmcg can you also review? Thanks

josephmcg (approved):
sure. I wrote part of it, but i think it looks good ;)

stavares843:
Captura de ecrã 2022-06-08, às 15 35 15

we could use this copy that joe added when he did the textile things on settings, but minor

Captura de ecrã 2022-06-08, às 15 35 09

stavares843 (approved):

PR #3401, opened 2022-06-01
fix(groups): prevent submit if groupname too short

josephmcg:
**What this PR does** 📖
- guard clause for group name errors (empty, less than 3, greater than 64)
- change conditional rendering of quick chat modal, rather than watching and clearing values. should help with memory too since it was always rendered before.
- css refactor, use gap rather than repetitive margin
- remove funky type declarations for 'friends'. improve type sense in file
- remove unused classes & id
- improve flexibility of error component, allows `<small>` version
- minor gender neutral language update
- improve type declaration for modals
- red border on input if invalid (shorter than 3 char)
- minlength support for input component

AP-1153

**Additional comments** 🎤
- I would like to expand that red border for invalid inputs to other places as well


phillsatellite:
Hey Joe! Works really well I only found a few minor things when I was exploring the branch

1. User can save group name with all spaces 

https://user-images.githubusercontent.com/93608357/171709683-89c5e768-eab3-4909-8481-b94c8a1fc65b.mov

2. When the user has less than 3 char's the red border appears, should we do the same for when user has reached the max char limit? 

https://user-images.githubusercontent.com/93608357/171710497-736e2a54-a338-4112-8bfd-f4aa1e859d37.mov

3. I'm not really sure if this one is an issue or not but figured I'd bring it to your attention anyway, should I be able to have just 2 emoji's as my group name? I was able to save it as that 
emoji name

josephmcg:
thanks @phillsatellite, these are some great things to fix
1. I adjusted the input to account for spaces
2. i had considered that, but using all 64 characters is valid. What we do at the moment is prevent further input, I think thats fine
3. emojis are super weird. I added something that should be really solid on everything except firefox. firefox will work OK, but not as well

components/views/friends/quick/Quick.vue
```diff
@@ -1,72 +1,74 @@
...
-        this.friends = []
-        this.isLoading = false
-        this.name = ''
-        this.error = ''
-      }
+    isInvalid(): boolean {
+      return (
+        !this.name ||
+        this.name.trim().length < 3 ||
+        this.name.trim().length > 64
```
molimauro, line 25: Can we put this in the config? :)

molimauro (approved):

molimauro (approved):

phillsatellite:
Found 2 things when retesting on iOS when user creates a new account the New Chat modal will be open and user can't close it, other thing was when I would try to create the group I would get this textile error 

IMG_5548
IMG_5549

josephmcg:
@phillsatellite thanks, forgot to update the conditional rendering on mobile too. It shouldn't appear now, it seems like this feature isn't ready on mobile.

josephmcg:
@molimauro fair question, no worries. if the input has pseudo invalid, yes. we can remove that, and only rely on the class if you prefer

josephmcg:
better ux would probably be to only show the border if they tried submitting and got an error, I'll update that

josephmcg:
fixed, will only apply to this input at the moment. will only show error border if they already tried to submit && current input is invalid

store/ui/types.ts
```diff
@@ -109,6 +109,7 @@ export enum ModalWindows {
   CALL_TO_ACTION = 'callToAction',
   RENAME_FILE = 'renameFile',
   CROP = 'crop',
+  CREATE_GROUP = 'creategroup',
```
stavares843, line 112: createGroup

stavares843:
added minor comment above, merging

stavares843 (approved):

PR #3403, opened 2022-06-01
fix(chat): add timestamps on hover for subsequent messages

josephmcg:
**What this PR does** 📖
- add timestamp on hover for subsequent messages in a group
- slightly adjust chat margin to make room for the timestamp

AP-1662

**Additional comments** 🎤
- I didn't realize we switched to static timestamps for messages, i love it!
- I feel like the special string timestamps (now, yesterday) don't add much value considering the effort it takes. I think those should be adjusted to static timestamps as well, then we could get rid of all the interval stuff

image

### DIFF
--- a/components/views/chat/group/Group.html
+++ b/components/views/chat/group/Group.html
@@ -7,7 +7,7 @@
       <span v-on:click="showQuickProfile">
         <UiUserName :username="username" :badge="badge" />
       </span>
-      <TypographyText :text="timestamp" :data-cy="`chat-timestamp`" />
+      <TypographyText :text="timestamp" data-cy="chat-timestamp" />
     </div>
     <div class="group-message">
       <Message :group="group" :message="group.message" :from="username" />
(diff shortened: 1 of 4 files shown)


Jekrimo (approved):

phillsatellite (approved):

stavares843 (changes requested):
Captura de ecrã 2022-06-01, às 20 21 34

lacking padding here

seed
end electric swear hurdle business priority salad ancient wire change please purpose

browser - chrome

josephmcg:
thanks for the account, fixed

stavares843 (changes requested):
same seed

Captura de ecrã 2022-06-02, às 02 23 31
 big message, when hovering no timestamp


then
Captura de ecrã 2022-06-02, às 02 23 47
Captura de ecrã 2022-06-02, às 02 23 52


these 2 messages when hovering one has a timestamp the other not, even though maybe they had the exact timestamp but then I saw these have the same timestamp, and both show

Captura de ecrã 2022-06-02, às 02 24 04
Captura de ecrã 2022-06-02, às 02 24 06

stavares843:
as joe explained, theres no timestamp if its the first message in a group


thanks @josephmcg 🔨

stavares843 (approved):

PR #3451, opened 2022-06-02
chore(faucet): testing different dev faucet

WanderingHogan:
**What this PR does** 📖

**Additional comments** 🎤

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -51,7 +51,7 @@ export const Config = {
         ],
   },
   solana: {
-    customFaucet: 'https://faucet.satellite.one',
+    customFaucet: 'https://dev-faucet.satellite.one',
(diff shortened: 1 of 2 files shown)


phillsatellite:
Tested and went over a little with Hogan, works a lot faster now but getting a error 400 when creating a new account so Hogan is gonna take another look at it and make some changes 
Screen_Shot_2022-06-03_at_12 30 51_PM

WanderingHogan:
@phillsatellite @stavares843 I think this is working better with the new faucet in the config. Solana is doing a dev update tomorrow around 7am ET - so it will probably be down again for a few hours then... Not much we can do.

We have a backup faucet we can use from quicknode (that will also go down at that time) that we can hopefully use if figment doesn't come up within a few hours.

josephmcg (approved):

stavares843 (approved):

PR #3453, opened 2022-06-03
fix(volume): bind output slider volume to stream volume

maljuburi:
**What this PR does** 📖
- add a volume watcher to update the audioStream element
- set initial volume when audioStream is set to the current volume state

AP-352

**Additional comments** 🎤
- 🎊  This is my first PR, please review carefully and don't hesitate to give feedback and share knowledge 🙂

### DIFF
--- a/components/ui/Chat/TypingIndicator/TypingIndicator.vue
+++ b/components/ui/Chat/TypingIndicator/TypingIndicator.vue
@@ -22,7 +22,7 @@ export default Vue.extend({
    */
   methods: {
     checkUsersTyping() {
-      if (this.users.length <= 3 && this.users.length !== 0) {
+      if (this.users.length <= 3 && this.users.length) {
         return 1
       }
       if (this.users.length > 3) {
--- a/components/ui/Chat/TypingIndicator/TypingIndicator.vue
+++ b/components/ui/Chat/TypingIndicator/TypingIndicator.vue
@@ -22,7 +22,7 @@ export default Vue.extend({
    */
   methods: {
     checkUsersTyping() {
-      if (this.users.length <= 3 && this.users.length != 0) {
+      if (this.users.length <= 3 && this.users.length !== 0) {
         return 1
       }
       if (this.users.length > 3) {
--- a/components/views/media/Media.vue
+++ b/components/views/media/Media.vue
@@ -203,6 +203,18 @@ export default Vue.extend({
         })
       }
     },
+    'audio.volume': {
+      handler(volume) {
+        // Bind stream audio element volume to slider volume
+        const audioStreamElements = document.getElementsByClassName(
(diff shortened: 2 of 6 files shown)


components/ui/Chat/TypingIndicator/TypingIndicator.vue
```diff
@@ -22,7 +22,7 @@ export default Vue.extend({
    */
   methods: {
     checkUsersTyping() {
-      if (this.users.length <= 3 && this.users.length != 0) {
+      if (this.users.length <= 3 && this.users.length !== 0) {
```
molimauro, line 25: Can be refactored to `if (this.users.length <= 3 && this.users.length)`

components/views/media/Media.vue
```diff
@@ -203,6 +203,18 @@ export default Vue.extend({
         })
       }
     },
+    'audio.volume': {
+      handler(volume) {
+        // Bind stream audio element volume to slider volume
+        const audioStreamElements = document.getElementsByClassName(
+          'remote-audio-stream',
+        ) as HTMLCollectionOf<HTMLAudioElement>
```
molimauro, line 211: Instead of getting the element through `getElementsByClassName` you can use `ref`. Or are there more than one audio elements?

molimauro (changes requested):
Hey Muhammad! Thanks for the first PR :) just a couple of minors and the rest looks good!

components/views/media/Media.vue
maljuburi: as far as I understand, we would expect multiple audio elements in a group call. so I tried to future proof it. good feedback though

components/views/media/user/User.vue
```diff
@@ -142,6 +142,18 @@ export default Vue.extend({
       loadVideos()
     })
   },
+  updated() {
+    // When audio is streamed, initialize stream volume to current volume.
+    if (!this.isLocal && !this.audio.deafened && this.audioStream) {
+      const audioStreamElements = document.getElementsByClassName(
+        `remote-audio-stream`,
+      ) as HTMLCollectionOf<HTMLAudioElement>
```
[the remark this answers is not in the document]
maljuburi: addressed above

components/views/media/Media.vue
molimauro: Ok thanks for the clarification! :)

molimauro (approved):

stavares843:
Hogan has a couple changes to request on these code changes.

WanderingHogan (approved):

WanderingHogan:
i dunno what i was thinking on that call earlier. When i skimmed it i thought this was doing something else. Good job, Muhammad!

phillsatellite (approved):

PR #3460, opened 2022-06-03
fix(groups): members order sorted alpha

josephmcg:
**What this PR does** 📖
- set group members with an alpha sort. it seems like the backend returns them in a random order every time
- improve performance w/ find rather than iterating 100% of the time. a couple other mutations here could use this same treatment

AP-1698

**Additional comments** 🎤

### DIFF
--- a/components/views/navigation/toolbar/Toolbar.vue
+++ b/components/views/navigation/toolbar/Toolbar.vue
@@ -89,7 +89,7 @@ export default Vue.extend({
     },
     ModalWindows: () => ModalWindows,
     src(): string {
-      const hash = this.recipient?.profilePicture
+      const hash = (this.recipient as Friend).profilePicture
       return hash ? `${this.$Config.textile.browser}/ipfs/${hash}` : ''
     },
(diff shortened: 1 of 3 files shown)


molimauro (approved):

stavares843 (changes requested):
Captura de ecrã 2022-06-03, às 16 53 17

fixed

Captura de ecrã 2022-06-03, às 16 49 37

but the order is not the same then the one that we have on friends list

phillsatellite:
One little thing I wanted to ask was when you are adding users in the new group chat modal should they be listed alphabetical or add as user selected? 
Screen Shot 2022-06-03 at 12 05 24 PM

josephmcg:
These changes do not affect the quick chat popup. That pulls from the 'friends' store namespace. this sets the 'groups' namespace list to alphabetical order so the group chat icons/names don't change order every time you click on a group

stavares843 (approved):

stavares843:
thanks for clarifying

PR #3461, opened 2022-06-03
refactor(style): clean up base css

ThomBos:
**What this PR does** 📖
Remove unused code and remove most of the general css expressions

AP-1635

**Additional comments** 🎤


phillsatellite:
Found 2 small UI issues on IOS when testing on the account creation page there is a green border around the Set Profile picture button and also on the Files page when you are uploading a file the loading button will have the same green border around it 
PIN
FILE

ThomBos:
> Found 2 small UI issues on IOS when testing on the account creation page there is a green border around the Set Profile picture button and also on the Files page when you are uploading a file the loading button will have the same green border around it 

thanks! i'm checking 👍

ThomBos:
fixed! 🔨

Jekrimo (approved):

molimauro (approved):

ThomBos:
> Noticed one more thing appear on iOS. To replicate you can't add the app to your homescreen you have to view through the browser but on account creation the Import Account and "I saved it" button will be blocked off and user can't scroll to view them Thanks @ThomBos !

How do you find these weird bugs?? 😂 

going to fix it 🐛💥

molimauro:
@ThomBos i think that is because you set height 100vh on body instead of 100% so the safe paddings are not considered

ThomBos:
@phillsatellite please let me know if it works (no hurry), i cannot test it on a ios mobile device 😢

molimauro:
> > @ThomBos i think that is because you set height 100vh on body instead of 100% so the safe paddings are not considered
> 
> ah yeah... do you think there's anyway to keep both padding and 100vh?

Just use 100% as we do for the width.. Otherwise you have to calc(100vh - safepaddings) but i dont like that too much

molimauro (approved):

phillsatellite (approved):

PR #3475, opened 2022-06-06
fix(store): switch accounts getters to direct store access

josephmcg:
**What this PR does** 📖
- I'm in the process of improving type sense for our getters. Since these weren't adding any value, I removed them in favor of mapState

AP-1728

**Additional comments** 🎤


Jekrimo (approved):

molimauro (approved):

phillsatellite:
Tested: Had some weird behavior on here, when User A would message User B, User B would not receive any of those messages until they navigated to a different page and back to chat 

https://user-images.githubusercontent.com/93608357/172475702-a07c09cc-9004-4ea6-92cf-1228459f31cd.mov

josephmcg:
that must be unrelated, I'm essentially just accessing the same variables with a different name

phillsatellite (approved):

stavares843:
that issue seems to be the textile listener issue after it goes inactive

stavares843 (approved):

PR #3506, opened 2022-06-08
feat(interactables): user picker component

jasonwoodland:
**What this PR does** 📖
- add new checkbox component
- add reusable friend picker component
- upgrade quick component to use new friend picker
- fix button animation speed when switching `type` prop in quick component
- add missing type definition

AP-1742
AP-1710

**Additional comments** 🎤

### DIFF
--- a/components/interactables/UserPicker/UserPicker.html
+++ b/components/interactables/UserPicker/UserPicker.html
@@ -12,29 +12,5 @@
       :selected="isSelected(friend)"
       @click="toggle(friend)"
     />
-    <InteractablesUserPickerListItem
-      v-for="friend in filteredFriends"
-      :friend="friend"
-      :selected="isSelected(friend)"
-      @click="toggle(friend)"
-    />
-    <InteractablesUserPickerListItem
-      v-for="friend in filteredFriends"
-      :friend="friend"
-      :selected="isSelected(friend)"
-      @click="toggle(friend)"
-    />
-    <InteractablesUserPickerListItem
-      v-for="friend in filteredFriends"
-      :friend="friend"
-      :selected="isSelected(friend)"
-      @click="toggle(friend)"
-    />
-    <InteractablesUserPickerListItem
-      v-for="friend in filteredFriends"
-      :friend="friend"
-      :selected="isSelected(friend)"
-      @click="toggle(friend)"
-    />
   </UiScroll>
 </div>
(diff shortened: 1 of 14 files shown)


components/interactables/UserPicker/UserPicker.vue
```diff
@@ -0,0 +1,48 @@
...
+  computed: {
+    ...mapState(['friends']),
+    filteredFriends() {
+      if (this.filter) {
+        const filterLower = this.filter.toLowerCase()
+        return this.friends.all.filter((friend: Friend) => {
+          return friend.name.toLowerCase().includes(filterLower)
+        })
+      }
+      return this.friends.all
```
josephmcg, line 23: i think you can skip the guard clause, includes('') will be true

components/interactables/UserPicker/UserPicker.vue
```diff
@@ -0,0 +1,48 @@
...
+    },
+  },
+  watch: {
+    selected() {
+      console.log('emit', this.selected)
```
josephmcg, line 28: remove logs

josephmcg (changes requested):

components/interactables/UserPicker/UserPicker.vue
jasonwoodland: I feel if we can skip a superfluous filter call then we should from the perspective of performance.

josephmcg (approved):

josephmcg:
making a note that we want to add some sort of list to keep track of currently selected users. thought it would be best in a separate PR to keep things simple

jasonwoodland:
@phillsatellite fixed :)

stavares843:
thanks @jasonwoodland 🔨

phillsatellite (approved):

stavares843:
very neat 🎉 

i noticed that it doesn't show feedback if the search doesn't have results and the user tries to click on chat now even if the button is already not tinted but maybe we could improve that on other PR

Screenshot 2022-06-10 at 00 08 25

this branch vs dev where we don't show the button if there are no results

but this component is way prettier 🎉 

cc @josephmcg

stavares843 (approved):

components/interactables/UserPicker/UserPicker.html
```diff
@@ -0,0 +1,16 @@
+<div class="user-picker">
```
stavares843, line 1: Captura de ecrã 2022-06-10, às 14 47 05

this is adding a warning when we build locally, could fix you on a separate PR? @jasonwoodland 🔨

stavares843: added ticket here - [AP-1766](https://satellite-im.atlassian.net/browse/AP-1766)

PR #3521, opened 2022-06-08
fix(chat): fix scroll to bottom in chat not working properly when non fixed size images are present

vimercati-samir:
What this PR does 📖
Fix scroll to bottom in chat not working properly when non fixed size images are present

[AP-1731](https://satellite-im.atlassian.net/browse/AP-1731)

Additional comments 🎤


molimauro:
Ok @vimercati-samir good draft, let me dm you for some changes!

stavares843:
is this ready? 🔨

molimauro:
Yes it is!

utilities/ImageSize.ts
```diff
@@ -0,0 +1,76 @@
...
+  source,
+  sizeType,
+}: {
+  source: string
+  sizeType: string
+}): string => {
+  if (source.includes('/$1/')) {
+    return source.replace('$1', sizeType)
+  }
+  return source
```
molimauro, line 22: why this?

molimauro:
Images with error keep loading:
image

utilities/ImageSize.ts
vimercati-samir: This logic was already there I only moved it, basically every sticker has an URL associated, for example https://satellite.mypinata.cloud/ipfs/QmXmpYrNUankzCCWR1U5ASDHcwhSJynQayMy1gT7RTU4ck/$1/AHHcloseup.gif,  this link return not found, but if you replace "$1" with "small" you can see the sticker, I don't know if it's intended to have links associated with stickers with that portion in the URL

molimauro: ok thanks!

vimercati-samir:
> Images with error keep loading: image

This bug occurs only when using older chats with glyphs in them (meaning chats that have been created without the fixes of this branch)

components/views/files/upload/Upload.vue
```diff
@@ -248,7 +253,21 @@ export default Vue.extend({
...
+          const originalSize = await getOriginalSizeFromDataUrl(fileAsDataURL)
+
+          await this.dispatchFile({
+            ...file,
+            ...getSizeFromAspectRatio(originalSize),
+          })
+        } else {
+          await await this.dispatchFile({
+            ...file,
+          })
```
molimauro, line 269: it is necessary to spread this?

vimercati-samir: nope, has been removed in commit 0133d4fd884e065fdfd9950862b01b70e48a1a1c

vimercati-samir:
I made some more changes because I wasn't saving the sizes in the textile payload and as a result of that the recipient wasn't receiving any sizes and the images would scale without control, now everything should be working fine

molimauro (changes requested):
Something broke in the last changes, can you check @vimercati-samir? Thanks

molimauro:
got also this:
image

vimercati-samir:
> The image aspect ratio is not correct: image

resolved

vimercati-samir:
> got also this: image

Starting from a clean chat from this branch (with the latest commits) would result in not having this error in console

molimauro (approved):

stavares843 (changes requested):
the images are way bigger on this PR

this branch

Captura de ecrã 2022-06-15, às 19 51 28

vs dev

Captura de ecrã 2022-06-15, às 19 52 13


same image, same seed, same browser

seed
whisper tired egg over alert reduce carpet asset draft whisper must lawn

stavares843:
no, these was images that were already there

thanks for clarifying

phillsatellite:
Noticed a small bug when testing, seems like when the user sends images or files in chat there will be a second loading bar when the picture is posting 
https://user-images.githubusercontent.com/93608357/174103759-ff222530-b27f-4778-9e8c-b7d2382804ce.mov

phillsatellite:
> > Noticed a small bug when testing, seems like when the user sends images or files in chat there will be a second loading bar when the picture is posting
> > 2nd.loading.bar.mov
> 
> It's not a bug, every image has a loading state that's independent from the upload state, if you refresh the page you will see that every image loads before appearing, maybe we can improve the ux by adding a label to distinguish the two types of loading (image and upload)

Awesome! Thanks for clarifying!

stavares843:
Screenshot 2022-06-20 at 23 51 16

if you send a image that will fail to load such as a svg for example will not scroll to the bottom properly

vimercati-samir:
> Screenshot 2022-06-20 at 23 51 16
> 
> if you send a image that will fail to load such as a svg for example will not scroll to the bottom properly

Should be resolved in commit 1bdec1ad0345a71b02cba06e5e194b5461c2cb6c, I also rebased and resolved conflicts

components/ui/Chat/Image/Image.less
```diff
@@ -3,30 +3,40 @@
   border-radius: @corner-rounding;
 
   .image-container {
-    overflow: hidden;
-    max-width: 300px;
```
josephmcg, line 7: bring back `overflow: hidden`. It prevents a blurred nsfw image from blurring the surrounding area
image

components/ui/Chat/Image/Image.html
```diff
@@ -13,13 +13,21 @@
...
   </figure>
-  <div v-else class="image-container">
+  <div
+    v-else
+    class="image-container"
+    :style="{ width: `${size.width}px`, height: `${size.height}px` }"
+  >
+    <div v-if="loading" class="image-loader">
+      <UiLoadersLoadingBar />
+    </div>
```
josephmcg, line 23: I don't particularly enjoy the loading icon. I think a skeleton image would be better (just a div with darker background color than the surrounding area)

components/ui/Chat/Image/Image.html
```diff
@@ -13,13 +13,21 @@
...
     <img
       class="is-image"
-      :class="{'blur-image' : image.nsfw && blockNsfw}"
+      :class="`${(image.nsfw && blockNsfw) ? 'blur-image' : ''} ${ loading ? 'hidden' : 'visible'}`"
```
josephmcg, line 26: This is somewhat hard to read, avoid using ternary if one of the cases is blank.

using template literals inside v-bind is helpful if you need to mix plain strings in with js values (like you're doing with height and width for the `image-container` above). Try this out instead:
`:class="[{'blur-image': image.nsfw && blockNsfw }, loading ? 'hidden': 'visible']"`

josephmcg: Even better, I would probably apply one of the visibility statues to the plain is-image class. Then you're only applying one class based on `loading`

components/views/chat/message/Message.html
```diff
@@ -70,6 +72,7 @@
     <UiChatEmbedsFile
       v-else-if="message.type === 'file'"
       :file="message.payload"
+      :size="{width: message.payload.width, height: message.payload.height}"
```
josephmcg, line 75: it's bad practice to pass data twice. you're already passing size values inside the file prop

components/ui/Chat/Embeds/File/File.html
```diff
@@ -1,5 +1,10 @@
 <div :class="isEmbeddable ? 'embedded-image' : 'embedded-file'">
-  <UiChatImage v-if="isEmbeddable" :image="file" :alt="file.name" />
+  <UiChatImage
+    v-if="isEmbeddable"
+    :image="file"
+    :size="size"
+    :alt="file.name"
```
josephmcg, line 6: since the file prop here contains size and width, you shouldnt double up the data here, just extract directly from file in child component
image

components/ui/Chat/Embeds/File/File.vue
```diff
@@ -15,6 +15,11 @@ export default Vue.extend({
       type: Object as PropType<FileMessagePayload>,
       required: true,
     },
+    size: {
+      type: Object as PropType<{ width: number; height: number }>,
+      default: () => {},
+      required: false,
+    },
```
josephmcg, line 22: delete this

components/ui/Chat/Image/Image.html
josephmcg: you can switch size to be image.height and image.width

components/ui/Chat/Image/Image.vue
```diff
@@ -22,12 +22,17 @@ export default Vue.extend({
       type: Object as PropType<FileMessagePayload>,
       required: true,
     },
+    size: {
+      type: Object as PropType<{ width: number; height: number }>,
+      required: true,
+    },
```
josephmcg, line 28: redundant, remove

components/views/chat/message/Message.vue
```diff
@@ -136,6 +136,8 @@ export default Vue.extend({
       setTimestamp,
       Config.chat.timestampUpdateInterval,
     )
+
+    console.log(this.$props.message)
```
josephmcg, line 140: remove console

components/views/files/upload/Upload.vue
```diff
@@ -110,15 +115,28 @@ export default Vue.extend({
...
             progress: 0,
             thumbnail: thumbnail ? await blobToBase64(thumbnail) : '',
           },
           address,
-        })
+        }
+
+        if (isFileSizeFixed) {
+          const fileAsDataURL = window.URL.createObjectURL(file.file)
+          const originalSize = await getOriginalSizeFromDataUrl(fileAsDataURL)
```
josephmcg, line 131: anytime you createObjectURL, you need to revoke or it causes memory leaks

components/views/chat/message/Message.html
```diff
@@ -51,8 +51,10 @@
     </template>
     <MessageGlyph
       v-if="message.type === 'glyph'"
-      :source="message.payload"
+      :source="message.payload.src"
       :pack="message.pack"
+      :size="{width: message.payload.width, height: message.payload.height}"
+      :sizeType="message.payload.sizeType"
```
josephmcg, line 57: if this component is only used here, it would make sense to pass the payload as one prop to reduce the complexity here. If we want to use this in other places, I could see leaving this

components/views/chat/message/glyph/Glyph.html
```diff
@@ -1,9 +1,16 @@
...
+<div
+  class="image-container"
+  :style="{ width: `${size.width}px`, height: `${size.height}px` }"
+>
+  <div v-if="loading" class="image-loader">
+    <UiLoadersLoadingBar />
+  </div>
   <img
-    class="glyph"
+    :class="`glyph ${ loading ? 'hidden' : 'visible'}`"
```
josephmcg, line 9: dont use template literals for classes if you dont have to

josephmcg (changes requested):
looking good Samir, a couple stylistic things

let me know if you have any questions

components/views/files/upload/Upload.vue
josephmcg: It would probably be better to refactor the getOriginalSizeFromDataUrl to accept a blob instead, and create the data url inside that utility function. That way you only need to create and revoke url once

utilities/ImageSize.ts
```diff
@@ -0,0 +1,76 @@
+import { StringToHashBucketFastInputs } from '@tensorflow/tfjs'
+import { Config } from '~/config'
+
+export const shouldFileSizeBeFixed = (type: string) => {
+  if (type.match('image.*')) {
+    return true
+  }
+
+  return false
+}
```
josephmcg, line 10: im not sure how this would handle corrupt files or unembeddable images. Try using stuff from the FileType utilities instead

vimercati-samir:
Thanks @josephmcg, all the issues should be resolved in commit d134bb2433518fbb9e7a67924c6cb144a30ca99a

phillsatellite:
Tested with most recent commits: found a few issues if the chat has a glyph in it before you switch to this branch it'll crash when you try to enter that chat 
Screen Shot 2022-06-24 at 2 59 32 PM

Another issue I had was whenever I would send a new glyph after already switching to this branch it would appear huge 
Screen Shot 2022-06-24 at 2 59 32 PM

josephmcg:
@molimauro @iltumio seems like that action call was moved [here](https://github.com/Satellite-im/Core-PWA/pull/3383/files). not sure how long this has been an issue. I would suggest moving it back unless Jeff moved it here for a good reason. its current location doesn't really fit with the rest of the method. 

@phillsatellite I'll let Samir confirm, but I believe this is intended to be a breaking change. At least the glyph size, maybe crash should be resolved with some optional chaining

josephmcg (approved):
wonderful, thanks Samir. I really like the new loading animation!

vimercati-samir:
As @josephmcg mentioned to test the functionality of this PR you have to start on a fresh account and send messages from this branch, crashes or wrong sizes on old messages occurs due to the fact that this pr changed the behaviour of how the messages are saved and displayed

stavares843:
we need to prevent the crash

 I was aware of the bad sizing, but we can't merge crashes on purpose

also, regarding the glyphs phil said

''I would send a new glyph after already switching to this branch it would appear huge'' so the glyph was sent already on this branch not before

vimercati-samir:
crashes and wrong images/glyphs size due to being sent on another branch or to an older version of this branch have been resolved in commit 7ae52e20c37068ba3c185c44bd7c3168f6f5cc61

molimauro:
I dont totally understand why do we want to prevent old chats from breaking since this is an intended breaking change. However Samir already fixed that so fine for now!

components/views/chat/enhancers/glyphs/item/Item.vue
```diff
@@ -54,27 +55,39 @@ export default Vue.extend({
         this.$store.commit('ui/setHoveredGlyphInfo', undefined)
       }
     },
-    sendGlyph() {
+    async sendGlyph() {
       const { id, address } = this.$route.params
       const activeFriend = this.findFriendByAddress(address)
+      const sizeType = 'small'
```
molimauro, line 61: Is this ever going to change?

vimercati-samir: Added comment in the code to address this concern in commit 909c2776c1a3816a0338ed8c3639fc4fca090e75

components/views/chat/message/Message.html
```diff
@@ -51,8 +51,9 @@
     </template>
     <MessageGlyph
       v-if="message.type === 'glyph'"
-      :source="message.payload"
+      :message="message.payload"
       :pack="message.pack"
+      :sizeType="message.payload.sizeType"
```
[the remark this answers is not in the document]
vimercati-samir: Changed accordingly in commit 909c2776c1a3816a0338ed8c3639fc4fca090e75

molimauro (approved):

molimauro:
Btw love the new loader

molimauro:
Sorry @vimercati-samir noticed a small reflow on the image placeholder, lets keep it at the same height of the original image thanks!

vimercati-samir:
> Sorry @vimercati-samir noticed a small reflow on the image placeholder, lets keep it at the same height of the original image thanks!

Should be resolved in commit bb5df4f848b77d16feb05d4882757bc87d140341

components/ui/Chat/Image/Image.html
```diff
@@ -1,5 +1,9 @@
 <div class="image">
-  <figure class="image-placeholder" v-if="usePlaceholder">
+  <figure
+    class="image-placeholder"
+    :style="{ width: `${image.width}px`, height: `${image.height}px` }"
```
molimauro, line 4: sorry @vimercati-samir, isnt this going to change the aspect ratio of the placeholder image? Its important to have the same height, but if we also force the same width this can lead to a strange aspect ratio right?

vimercati-samir: Should be resolved in commit 0b2a64f14341422c2e6dc751d54fb7ae75e31fa1

molimauro (approved):

phillsatellite (approved):

stavares843:
https://user-images.githubusercontent.com/29093946/176811049-05e856ea-af3d-4df9-85df-b5c2e36d4f2f.mov

is there a way to prevent these glitches? sometimes doesn't scroll all the way down or it scrolls but goes up and then down

seed
oval dignity twist coyote frozen exchange fiscal april doll conduct pudding meat

macOS, Chrome

molimauro:
> Gravacao.do.ecra.2022-07-01.as.03.22.42.mov 
> is there a way to prevent these glitches? sometimes doesn't scroll all the way down or it scrolls but goes up and then down
> 
> seed oval dignity twist coyote frozen exchange fiscal april doll conduct pudding meat
> 
> macOS, Chrome

@vimercati-samir at 0:17 the bitmap images create a reflow, can you see it?

molimauro:
Isnt the scrollbar glitch pretty much this one? https://satellite-im.atlassian.net/browse/AP-1925
I dont think its related to this PR. Does this happen on dev too? :)

molimauro:
> Gravacao.do.ecra.2022-07-01.as.03.22.42.mov 
> is there a way to prevent these glitches? sometimes doesn't scroll all the way down or it scrolls but goes up and then down
> 
> seed oval dignity twist coyote frozen exchange fiscal april doll conduct pudding meat
> 
> macOS, Chrome

@stavares843 was this one on this branch or on an old one with images already in the chat?

stavares843:
> old one with images already in the chat

stavares843:
> Does this happen on dev too? :)


yes but I was wondering if this PR would address that as well 😂

stavares843 (commented):
testing webhook

stavares843 (commented):
testing webhook 2

WanderingHogan:
closing but leaving branch to come back to if we need it

PR #3528, opened 2022-06-09
chore(cleanup): remove unused markup

josephmcg:
**What this PR does** 📖
- remove leftover interval code
- switch some code to shorthand
- remove unused case since all image messages are encoded as 'file'. I removed all of the other associated code a week or two ago, missed this one bit

**Additional comments** 🎤

### DIFF
--- a/components/views/user/User.vue
+++ b/components/views/user/User.vue
@@ -43,7 +43,6 @@ export default Vue.extend({
     return {
       existConversation: false,
       isLoading: false,
-      timestampRefreshInterval: null,
     }
   },
   computed: {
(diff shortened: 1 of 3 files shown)


WanderingHogan (approved):

components/views/user/User.vue
```diff
@@ -115,10 +114,6 @@ export default Vue.extend({
       })
     })
   },
-  beforeDestroy() {
-    clearInterval(this.$data.timestampRefreshInterval)
-    this.$store.commit('ui/toggleContextMenu', false)
```
WanderingHogan, line 120: we should totally take this out of state. I cant think of why we would want this to be stored there in the first place

WanderingHogan: but we can do it later when there is more time, or an overall state cleanup task if you think it makes sense

josephmcg: now i remember why we added the context menu mutation. it was to handle bugs related to removing friends and clicking the context menu mid operation. will fix

WanderingHogan:
still ready for QA after you put the beforeDestroy toggleContextMenu = false back

stavares843 (approved):

PR #3538, opened 2022-06-09
feat(notifications): add browser alert and notification image

Jekrimo:
**What this PR does** 📖
Adds browser push notification, adds profile thumbnail to in app notification

AP-1633

**Additional comments** 🎤
Needs some investigation into adding flair into the browser alert, ie. icon, theme color, localization, etc - ticket is made at AP-1755 & AP-1756

### DIFF
--- a/utilities/Notifications.ts
+++ b/utilities/Notifications.ts
@@ -11,7 +11,6 @@ const isSupported = (): boolean =>
 export const Notifications = class Notifications {
   currentPlatform: PlatformTypeEnum = PlatformTypeEnum.ANDROID
   notificationPermission: string = 'denied' // web: 'denied' 'granted' 'default'
-  sendNotification: any = this.sendNotifications
   $Config: typeof Config = Config
 
   constructor() {


WanderingHogan:
image

utilities/Notifications.ts
```diff
@@ -135,6 +139,29 @@ export const Notifications = class Notifications {
         body: message,
       })
     }
+    if (this.currentPlatform === PlatformTypeEnum.WEB) {
```
WanderingHogan, line 142: What do you think about using the built in notification api right above this for Web and Electron, modified a little bit?

Screen Shot 2022-06-09 at 3 58 53 PM

It lets us include our icon on the chrome notification
image

WanderingHogan: if you go this route you can remove the config import/parts too

WanderingHogan: icon might work with the image hash you passed, in too

libraries/Textile/NotificationManager.ts
```diff
@@ -115,6 +115,15 @@ export class NotificationManager {
       CollectionName,
       Query.where('_id').eq(notificationId[0]),
     )
+    const makeNotificationMessage = `New message from ${payload.from}`
```
WanderingHogan, line 118: can you add this to the translations, an example that would allow you to pass the from you can see if you search for `user.registration.username_error`

WanderingHogan (changes requested):
can you check out the feedback and lmk

libraries/Textile/NotificationManager.ts
WanderingHogan: don't worry about this translation right now - as you are showing me in a DM we are having an issue using vues translation library outside of vue in the utilities

Jekrimo: https://media.tenor.com/images/f4ec7060f9dc5ed3bc0fb1d4e87f3e14/raw

Jekrimo: done and done

utilities/Notifications.ts
Jekrimo: YESSSS it worked with the image hash for the local!! yus, yus, yus. Thank you!

libraries/Textile/NotificationManager.ts
Jekrimo: @stavares843 If/when you test, can you tell us if the browser notification comes through already translated without us doing anything? (I feel like it should)

utilities/Notifications.ts
```diff
@@ -117,24 +119,51 @@ export const Notifications = class Notifications {
...
-      await new Notification(type, {
-        tag: titleText,
+      await new Notification(`${type} ${titleText}`, {
+        tag: String(new Date().getTime()),
         body: message,
+        icon: `${this.$Config.textile.browser}/ipfs/${image}`,
+        badge: `${this.$Config.textile.browser}/ipfs/${image}`,
       })
     }
+    // if (this.currentPlatform === PlatformTypeEnum.WEB) {
```
WanderingHogan, line 144: @stavares843 @phillsatellite can we leave this commented code in for now and we remove it after this all evolves a little bit before EA

WanderingHogan (approved):

phillsatellite (approved):

stavares843 (changes requested):
I tried this on both macOS and windows - Chrome.

Even after activating Chrome notifications on macOS settings, I received no alerts.

On windows, I got the pop-up asking permission for browser notifications, I said yes but no alerts.

stavares843:
neat, thanks 🔨 checking

utilities/Notifications.ts
iltumio, line 144: do we need to keep commented code?

utilities/Notifications.ts
```diff
@@ -11,6 +12,7 @@ export const Notifications = class Notifications {
   currentPlatform: PlatformTypeEnum = PlatformTypeEnum.ANDROID
   notificationPermission: string = 'denied' // web: 'denied' 'granted' 'default'
   sendNotification: any = this.sendNotifications
```
iltumio, line 14: is it possible to use a proper type instead of 'any'?

utilities/Notifications.ts
Jekrimo: Yeah, Hogan want's to leave this as we will be needing it for mobile and other such things.

WanderingHogan (approved):

iltumio (approved):
It's approved from my side. I found one small issue related to the notification removal. If you have 2 notifications and you try to remove one of them, both notifications are canceled

Jekrimo:
> works on windows
> 
> image
> 
> but on macOS, even with the permissions activated on macOS settings I don't have any alerts from chat
> 
> Captura de ecrã 2022-06-16, às 15 46 27

Keep me posted if it's the mac's chrome settings like we discussed on discord

stavares843:
yes will check 💥

PR #3550, opened 2022-06-10
fix(chat): upload rework

josephmcg:
**What this PR does** 📖
### Code related
- helpful new getters
  - isGroup - determine whether active conversation is group or DM
  - recipient - fetch active recipient data (Group or User). This will be used instead of passing recipient prop all over the place. its still being used as a prop for unrelated components, but I thought that should be handled separately
  - getFiles - fetch files for active conversation based on recipient address
- generate smaller thumbnails for file uploads, should help w/ memory
  - move this thumbnail function from filesystem to utilities
- move & rename types associated w/ chat upload
- rather than updating local state inside the components for files with a subscription, move all files to store only
- rename upload nested components to be more concise
- move file count error outside preview for better v-if rendering (save memory)
- reduce redundancy for progress tracking. add to files store value rather than use a separate list just for progress
- refactor preview css to use flex (was using inline-block 😵‍💫)
- move file upload method to chatbar since everything is in store. easier than calling a method of a ref

### Functionality
- big file uploads with streams api
- allow sending of nsfw files, but attach an nsfw boolean to it.
  - if a user has the 'block nsfw content' setting on, it will blur
  - In a future update, I will add a privacy setting to blur images from people not on your friends list (provide option to click and remove blur)
- not functionality, but make upload preview closer to figma

AP-1470, AP-1650, AP-1651, AP-1844

**Additional comments** 🎤
### things to come
- aforementioned privacy blur setting
- index to track chat uploads
- improve file limit ux. rather than preventing upload, we could slice the file list and add up to the 8 file limit
- show preview immediately and load thumbnail async


components/views/chat/chatbar/Chatbar.html
```diff
@@ -1,30 +1,16 @@
-<div
-  id="chatbar"
-  ref="chatbar"
-  :class="ui.replyChatbarContent.id ? 'has-reply' : ''"
->
+<div id="chatbar" ref="chatbar">
   <Enhancers @click="handleChatTextFromOutside" :recipient="recipient" />
```
molimauro, line 2: We can remove recipient also here right?

components/views/files/upload/preview/Preview.html
```diff
@@ -0,0 +1,18 @@
+<div
+  class="file-preview"
+  :class="{bordered : getFiles.length}"
+  @touchmove="handleTouchPreview"
+  @touchstart="handleTouchPreview"
+  @touchend="handleTouchPreview"
+>
+  <UiScroll horizontalScroll scrollbarVisibility="scroll">
```
molimauro, line 8: Can we use here `UiSimpleScroll`? We're ditching the old one thanks :)

components/views/files/upload/Error.vue
```diff
@@ -0,0 +1,39 @@
...
+<style scoped lang="less">
+.file-error {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: @normal-spacing;
+  width: calc(@full - 2rem);
+  margin: 0 @normal-spacing;
+  background-color: #181f2c;
+  border-bottom-color: rgba(87, 101, 116, 0.3);
```
molimauro, line 32: new colors?

phillsatellite:
One thing I found when testing, if I send an NSFW image in chat, then block NSFW when I open the image in chat should it be blurred? 
https://user-images.githubusercontent.com/93608357/173372783-c3f79afc-5346-4652-b95d-551fd89f366d.mov

phillsatellite:
One thing I found when testing, if I send an NSFW image in chat, then block NSFW when I open the image in chat should it be blurred? 
https://user-images.githubusercontent.com/93608357/173372783-c3f79afc-5346-4652-b95d-551fd89f366d.mov

components/views/files/upload/Upload.vue
```diff
@@ -65,192 +53,84 @@ export default Vue.extend({
...
-        const newFiles: File[] = [...event.target.files]
+      this.$store.commit('chat/setCountError', false)
+      if (!this.recipient) {
+        return
+      }
+      const newFiles: File[] = [...event.target.files]
 
-        if (newFiles.length + this.files.length > 8) {
-          this.$store.commit('chat/setCountError', true)
+      if (newFiles.length + this.getFiles.length > 8) {
```
molimauro, line 73: 8 sounds like a config ;)

utilities/Thumbnail.ts
```diff
@@ -0,0 +1,71 @@
...
+      format: 'JPEG',
+      quality: 1,
+    })
+    const fileJpg = new File([outputBuffer.buffer], file.name, {
+      type: 'image/jpeg',
+    })
+    if (await _tooLarge(fileJpg)) {
+      return
+    }
+    return blobToBase64(await skaler(fileJpg, { width }))
```
molimauro, line 33: why we return the base64 instead of the image?

components/views/files/upload/Error.vue
josephmcg: leftover from copy paste, thanks

utilities/Thumbnail.ts
josephmcg: originally this was only working within the context of file system. in that case I used base64 so I can easily add it to the json index. In this case, it doesnt need base64. I can rework that

josephmcg:
@phillsatellite @molimauro 
Thank you both for taking a look. everything has been addressed

utilities/Thumbnail.ts
molimauro: ok perfect thanks a lot!

molimauro (approved):

molimauro:
Good to me, thanks Joe!!

phillsatellite:
Blur when preview is fixed thanks Joe! I found 2 other things when I retested, it seems like the "view original" has moved from the left side of the preview to the center 
Screen Shot 2022-06-14 at 11 15 38 AM

I also saw this error if I cancel a 2GB file upload in chat half way through. This error will appear each time you navigate into the chat as well
https://user-images.githubusercontent.com/93608357/173612903-a652fbd1-fee9-40ed-a8c3-47d8b1fa5a36.mov

josephmcg:
@molimauro I realized why we were using base64. I was encountering bugs where the rendered preview items would lose track of the images if you deleted the ones in the middle. I switched back to base64 for now since it's simpler. maybe we can revisit later. I'm glad we split up the thumbnail and base64 utils though.

@phillsatellite I adjusted the style of that 'view original' link. I removed the box shadow and improved the context due to discussions we were having in [this ticket](https://satellite-im.atlassian.net/browse/AP-1561). 
For now, I changed it so you cannot remove a file from the list once upload has started. I would like to look into cancelling uploads at a later date

josephmcg:
heads up that gifs will not move inside the preview. I would like to improve that separately

molimauro:
> @molimauro I realized why we were using base64. I was encountering bugs where the rendered preview items would lose track of the images if you deleted the ones in the middle. I switched back to base64 for now since it's simpler. maybe we can revisit later. I'm glad we split up the thumbnail and base64 utils though.
> 
> @phillsatellite I adjusted the style of that 'view original' link. I removed the box shadow and improved the context due to discussions we were having in [this ticket](https://satellite-im.atlassian.net/browse/AP-1561). For now, I changed it so you cannot remove a file from the list once upload has started. I would like to look into cancelling uploads at a later date

Ok perfect @josephmcg, thanks for explaining! All good then ;)

stavares843 (changes requested):
dev

Captura de ecrã 2022-06-15, às 23 47 14


this branch

Captura de ecrã 2022-06-15, às 23 48 34


same image, same account, same browser

josephmcg:
this did not occur for me. could you make sure you have the latest changes and you `yarn dev` after switching to this branch? not switching while it is already running
image

stavares843:
no longer happens

strange, i also did that yesterday and was still reproducible


also

dev vs this branch

Captura de ecrã 2022-06-16, às 15 59 14
Captura de ecrã 2022-06-16, às 15 59 19


on dev, if you send an image and fails to load such as an SVG, appears the failed to load placeholder, but on this branch doesn't appear

josephmcg:
I do not have that issue :/
I just rebased, hoping that will fix it for you

image

stavares843 (approved):

PR #3551, opened 2022-06-10
chore(loading): removed loading screen on chat

WanderingHogan:
**What this PR does** 📖
Since we are pulling messages from a local db and not from ipfs/textile every time now for all data, this takes away that big loading screen that makes chat feel slow. New messages are still fetched. Currently re-using the send animation.

This was pitched to Liz to see if she was OK with it, and requested we make ui updates for it later if she is okay with this change.

This was added to make @stavares843 happy
[AP-1799](https://satellite-im.atlassian.net/browse/AP-1799) This ticket was to make @tooshel happy

Before:

https://user-images.githubusercontent.com/2993032/172978017-d974a02d-f7ee-4a5b-ac00-522036894e1f.mov

After:

https://user-images.githubusercontent.com/2993032/172978052-26f724b7-39f6-419f-a635-96aa49104add.mov

**Additional comments** 🎤

### DIFF
--- a/components/ui/Chat/InfiniteScroll/InfiniteScroll.vue
+++ b/components/ui/Chat/InfiniteScroll/InfiniteScroll.vue
@@ -43,7 +43,6 @@ export default {
         this.$emit('intersect', {
           loaded: () => {
             this.debounceTimeout = setTimeout(() => {
-              this.isIntersecting = false
               this.scrollObserver.observe(this.$refs.root)
             }, 100)
           },
--- a/components/views/chat/chatbar/Chatbar.html
+++ b/components/views/chat/chatbar/Chatbar.html
@@ -19,7 +19,6 @@
   >
     <div class="control-wrap file-upload-wrap">
       <FilesUpload
-        v-if="recipient"
         :recipient="recipient"
         ref="file-upload"
         :files="files"
--- a/components/views/chat/conversation/Conversation.html
+++ b/components/views/chat/conversation/Conversation.html
@@ -24,7 +24,7 @@
         </div>
       </div>
       <div
-        v-if="textile.messageLoading || !initializingStatus"
+        v-if="textile.messageLoading"
         data-cy="loading-indicator"
         class="zoop"
       >
--- a/components/views/chat/conversation/Conversation.html
+++ b/components/views/chat/conversation/Conversation.html
@@ -24,7 +24,7 @@
         </div>
       </div>
       <div
-        v-if="textile.messageLoading"
+        v-if="textile.messageLoading || !initializingStatus"
         data-cy="loading-indicator"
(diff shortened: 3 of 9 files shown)


josephmcg (changes requested):
can you map the getter directly in the component rather than passing as a prop?

I love the ux though, great

molimauro (approved):

phillsatellite:
Noticed this "loading messages" flashes when navigating into chat, messaged Hogan about it but wanted to leave a comment on here as well 
https://user-images.githubusercontent.com/93608357/173143667-e8dec0fb-0375-463d-98cf-b07391e52af2.mov

WanderingHogan:
Are you okay with not considering this an actual issue? I noticed that this is related to the virtual scroll. This was introduced here: `components/views/chat/conversation/Conversation.html`, in the templates for UiChatInfiniteScroll component. In the attached video you can see this happens on dev, and the more i think about it, i don't think it's a bad thing or a bug or anything; what do you think? I think it would make more sense to ditch that 'end to end encryption' message that gets pushed off the screen after a few messages in favor of a permanent small icon that makes the user confident the messages are encrypted, and a background screen when there are no messages that says 'all messages are protected with end to end encryption', that will go away when messages are sent, after EA or when there is time?

https://user-images.githubusercontent.com/2993032/173243710-8bc6769f-496c-48fb-a12f-885dae840a0c.mov

stavares843:
>  i don't think it's a bad thing or a bug or anything

agree as well, updated label

stavares843:
also, this is way better compared with dev

phillsatellite:
Not sure if anyone else had this issue or not but can't send message in chat. The curser will always show as disabled
https://user-images.githubusercontent.com/93608357/173369394-cddf62d5-5254-44bc-9129-db5b591a2280.mov

stavares843:
I have that while is loading, after is loaded, that is gone and I can send messages 🔨

stavares843:
That's on purpose, because we don't allow user to be sending messages if we are still fetching data.

WanderingHogan:
@josephmcg I undid moving the getter to the conversation file, doing that screwed something when you were loading accounts with images loaded

stavares843:
both Phil account and Cypress load messages fine now

stavares843:
Hogan is gonna look into adding this on groups chats as well

josephmcg:
@WanderingHogan I think you removed the getter from the initial file too. I was saying to keep the getter in the original file, because it relies on it. but rather than passing as a prop to the child, map it there too

WanderingHogan:
@josephmcg good call, thank you, Joe

phillsatellite (approved):

stavares843 (approved):

PR #3581, opened 2022-06-12
fix(groupCreateForm): allow enter on name input to create group

WanderingHogan:
**What this PR does** 📖
Combination of allowing you to hit enter in the text input field to create a group (image), and minor refactor of groupIcon when investigating an icon issue

Screen Shot 2022-06-12 at 2 47 04 PM

AP-1680

**Additional comments** 🎤


josephmcg (approved):

josephmcg:
yeah, it's all within the form. I can't speak for Hogan, but that's how I envisioned it

stavares843:
enter should be only on the group name because that's the field is gonna creat the group, the search should not create a group

WanderingHogan:
I made it so you could specific if you want to submit the input field on enter, so hitting enter on the search didn't create a group

stavares843 (approved):

PR #3585, opened 2022-06-13
fix(toolbar): remove secondary create group modal

josephmcg:
**What this PR does** 📖
- remove secondary modal, in favor of making quickchat the only place to create a group. If we feel like there should be a way to create groups from the toolbar, we should render quickchat there instead of this

AP-1779

**Additional comments** 🎤


phillsatellite (approved):

Jekrimo (approved):

molimauro (approved):

stavares843:
@josephmcg why do you think we should remove the icon on the toolbar? we also have ''duplicated'' places to do a video call for example context menu vs toolbar

also, was this change checked with Liz?

josephmcg:
I don't necessarily have anything against a create group option here. I just thought it was silly to have multiple modals that essentially do the same thing. it's twice the maintenance for us. I can add in the fleshed out quickchat here after [this](https://github.com/Satellite-im/Core-PWA/pull/3588) goes through. the css will make it easier

stavares843:
sounds good

stavares843 (approved):

PR #3663, opened 2022-06-16
fix(groups): message glitch flashing in group chat

vimercati-samir:
What this PR does 📖
Fix message glitch flashing in group chat

[AP-1745](https://satellite-im.atlassian.net/browse/AP-1745)

Additional comments 🎤

### DIFF
--- a/components/views/chat/conversation/Conversation.html
+++ b/components/views/chat/conversation/Conversation.html
@@ -9,7 +9,7 @@
     <div class="conversation-content">
       <div
         class="group"
-        v-for="group in currentChat.messages"
+        v-for="group in messages"
         :id="group.id"
         :key="group.id"
       >
--- a/components/views/chat/conversation/Conversation.vue
+++ b/components/views/chat/conversation/Conversation.vue
@@ -16,6 +16,12 @@ export default Vue.extend({
       default: '',
     },
   },
+  data() {
+    return {
+      hasConversationBeenLoaded: false,
+      messages: [],
+    }
+  },
   computed: {
     ...mapState({
       ui: (state) => (state as RootState).ui,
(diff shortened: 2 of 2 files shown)


molimauro:
This is breaking the infinite scroll functionality and i dont like too much having messages in a local state. Does this happen also on directs?

vimercati-samir:
> This is breaking the infinite scroll functionality and i dont like too much having messages in a local state. Does this happen also on directs?

I'm working on it to prevent the scroll functionality from breaking, this bug only happens on group chat, essentially messages arrive before the flag "textile.conversationLoading" goes from true to false causing this scenario:

1) you send a message
2) the loader would appear
3) the message would appear but the loader would be still there for a split second
4) the loader would disappear causing the message to shift down to take the space where the loader was

To prevent this we wait for the flag "textile.conversationLoading" to go from true to false and then we show the new message that takes the space of the loader instantly

josephmcg:
why not just change the store value earlier then so this doesn't happen? This seems a little over-engineered

josephmcg:
Rather than focus on fixing the loading indicator.. Could we discuss whether we should even have it to begin with?
It makes sending a message feel relatively slow. It would probably be better ux to display the message as if it were sent, perhaps dimming the font color while is it still sending. what do you think @molimauro ?

molimauro:
> Rather than focus on fixing the loading indicator.. Could we discuss whether we should even have it to begin with? It makes sending a message feel relatively slow. It would probably be better ux to display the message as if it were sent, perhaps dimming the font color while is it still sending. what do you think @molimauro ?

I discussed this with Liz in the past and she suggested not to remove it but to show the loader and if the network is slow (lets say after a second) dim the font color while pending

josephmcg:
I typically try to make loaders more subtle.

I think of differently colored text as a skeleton for the actual message. Similar to Samir's recent image loader changes, favoring the skeleton (darker bg rectangle) over the standard loading icon. I can't speak for all users, but an icon feels slow and somewhat jarring for me when the text does go through. Maybe for longer, less common actions I would enjoy an animation like that, but sending a message is probably going to be the most common activity on the app.

molimauro:
Honestly i think it too, maybe should we reopen the discussion in the ui/ux channel?

PR #3668, opened 2022-06-16
fix(chat): is_typing indicator

molimauro:
**What this PR does** 📖
Restore typing notification in direct and groups chat

AP-1734

Should resolve the Textile listener as well.
The typing event will be sent only after the users are connected (as for starting a call for example)

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/chatbar/Chatbar.vue
+++ b/components/views/chat/chatbar/Chatbar.vue
@@ -173,22 +173,11 @@ export default Vue.extend({
   },
   methods: {
     /**
-     * @method typingNotifHandler
-     * @description Wraps the event handler for dispatching typing notifications
-     * TODO: Right now this is hard coded to the WebRTC Data method, in the future this should be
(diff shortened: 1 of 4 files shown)


iltumio (approved):

molimauro:
> Hi Mauro! I checked with Sara and she didn't think this was a bug but she wanted me to run it by you as well! User A refreshes page while User B is typing a message, when User A loads back into account it'll automatically show User B is typing but it seems to show longer than the usual timeout we have (personal opinion lol)
> 
>  bug.mov

Hey @phillsatellite! Im aware of this, on refresh itll show the typing indicator till textile is loaded if the other user was previously typing. Thats because on refresh we lose the timeout that normally removes the indicator and so the timing is a bit different.
I dont think its too annoying now so in my opinion we can consider this as intended :)

phillsatellite:
Sounds good to me @molimauro ! Gonna retest now 😄

phillsatellite (approved):

stavares843 (approved):

PR #3560, opened 2022-06-10
fix(settings-modal): fixed modal style

ThomBos:
**What this PR does** 📖
Fixes settings' modal style

**Which issue(s) this PR fixes** 🔨
AP-1765

**Special notes for reviewers** 🗒️

**Additional comments** 🎤

PR #3669, opened 2022-06-16
feat(select): new keyboard accessible select component

josephmcg:
**What this PR does** 📖

AP-1816

**Additional comments** 🎤


josephmcg:
> Wow, cool!! looks very, very promising 🚀 
> 
> 
> 
> 
> 
> 
> 
> P.S. Idk if the fix i've done for selects will affect this (not merged yet)

Im probably going to be removing the current select component and replacing it with this. So don't worry about existing select bugs

components/interactables/Select/Select.less
```diff
@@ -1,94 +1,85 @@
...
+  margin-top: @xlight-spacing;
+
+  &::-webkit-scrollbar {
+    width: @scroll-bar-size;
   }
-  &.down {
-    .items {
-      top: 30px;
-    }
+  &::-webkit-scrollbar-track {
```
molimauro, line 64: are we redefining the scrollbar attrs instead of using our scrollbar component for a particular reason? :)

josephmcg: i just thought this was easier. I don't fully understand how the scroll component works, and I do not understand the benefits

ThomBos: Yeah, that was much easier (i've tried that solution as well with selects 😅), but we are trying to keep the same scroll component all around the app to make it easier to find its style and modify it.

Atm the component is really easy, it just takes his container's height and verify that it's content isn't overflowing, if it overflows, it shows the scroll, that's it. 
So all the problems can be solved by modifying its container style.

The benefits of keeping this components are: 
- a lightweight native component that is more programmable and flexible than a custom more javascript-based
- having the style almost (you can use #simple-scroll on css to controll for example the max-height) in a unique file

On the other hand, i've noticed that with selects it has some issues, that's why i added a prop to make it invisible or it would have kept a small padding to the right. (check #3560 to see that prop)

I'm open betters solutions tho, let me know if you have any ideas!! 💥🚀

josephmcg: Thanks @ThomBos, I appreciate your insight. I can experiment with integrating the scroll component next week and let you know how it goes

molimauro: Yes basically those are the main reasons. We had 2/3 external libraries for scrollbars and one custom so it was a mess. Much better relying on 1 for virtualization and 1 for the rest.

In this case im not against making some common css rules and used them in our component and where needed, as long as we have all coherent ;)

molimauro (approved):
Im down with this, will take care of the scrollbar question later on!

components/interactables/Select/Select.less
josephmcg: my typical philosophy is to do it with css rather than js if possible. I still don't really what you can achieve with this component that can't be accomplished with css (style and overflow-y:auto)

josephmcg: @molimauro I added a base style to * which will apply our styled scrollbar. I think we could possibly handle scroll with plain css in most places

stavares843 (changes requested):
Captura de ecrã 2022-06-21, às 00 01 17
Captura de ecrã 2022-06-21, às 00 01 21

dev vs this branch

no arrow displaying when the field is locked, should display anyone to show the user is usable when is not locked

stavares843 (changes requested):
Captura de ecrã 2022-06-21, às 00 04 14
Captura de ecrã 2022-06-21, às 00 01 45

dev vs this branch

we have space to display the full list of the themes, we shouldn't add a scrollbar there

stavares843 (changes requested):
Captura de ecrã 2022-06-21, às 00 02 24

Captura de ecrã 2022-06-21, às 00 02 20


dev vs this branch

that field no longer has the default value displaying

stavares843:
thanks 🔨

stavares843 (approved):

PR #3677, opened 2022-06-16
fix(sidebar): get last message and display it in sidebar

AlexZakablukov:
**What this PR does** 📖
get last message and display it in sidebar

[AP-1410](https://satellite-im.atlassian.net/browse/AP-1410)

**Additional comments** 🎤

### DIFF
--- a/components/views/user/User.html
+++ b/components/views/user/User.html
@@ -9,14 +9,11 @@
   <div class="user-info">
     <TypographyTitle :size="6" data-cy="sidebar-user-name" :text="user.name" />
     <div
-      v-if="!lastMessageLoading"
+      v-if="lastMessage"
       v-html="wrapEmoji(markdownToHtml(lastMessage))"
       :class="`subtitle ${containsOnlyEmoji(lastMessage) ? 'bigmoji' : ''}`"
       ref="subtitle"
     />
-    <div v-else data-cy="loading-indicator" class="user-zoop">
-      <UiLoadersLoadingBar />
-    </div>
   </div>
   <div class="user-chat-details" v-if="!lastMessageLoading">
     <TypographyText :size="6" :text="timestamp" v-if="hasMessaged" />
--- a/components/views/user/User.html
+++ b/components/views/user/User.html
@@ -9,7 +9,6 @@
   <div class="user-info">
     <TypographyTitle :size="6" data-cy="sidebar-user-name" :text="user.name" />
     <div
-      v-if="lastMessage"
       v-html="wrapEmoji(markdownToHtml(lastMessage))"
       :class="`subtitle ${containsOnlyEmoji(lastMessage) ? 'bigmoji' : ''}`"
       ref="subtitle"
--- a/components/views/user/User.html
+++ b/components/views/user/User.html
@@ -15,7 +15,7 @@
       ref="subtitle"
     />
   </div>
-  <div class="user-chat-details" v-if="!lastMessageLoading">
+  <div class="user-chat-details">
     <TypographyText :size="6" :text="timestamp" v-if="hasMessaged" />
     <TypographyTag v-if="unreadMessageCount" :text="unreadMessageCount" />
   </div>
--- a/components/views/user/User.html
+++ b/components/views/user/User.html
@@ -9,12 +9,16 @@
   <div class="user-info">
     <TypographyTitle :size="6" data-cy="sidebar-user-name" :text="user.name" />
     <div
+      v-if="!lastMessageLoading"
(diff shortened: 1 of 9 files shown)


phillsatellite:
Is there anything we can do about all the loading bars appearing when user logs in? Sara and I we're both thinking it looked kinda messy 
https://user-images.githubusercontent.com/93608357/174358463-a2a513dd-be46-4cb1-8589-a72e409cf97f.mov

stavares843:
agree

molimauro:
We can change the loader or we can remove it and show the last msg we have in store (even though can be that is not the last) and once the real last is fetched show finally that

AlexZakablukov:
Fixed, now it shows the last message from the store while the actual last message is loading, without any loaders

Clayton215:
The sidebar works as should, however, I am not able to send pictures out of local. They show as sent on the local account but they don't actually go through. 
Screen Shot 2022-06-21 at 7 07 16 PM

josephmcg (changes requested):
whats the goal of switching it from the textile store namespace to friends? I thought we already had a decent foundation and would just build that out

AlexZakablukov:
> whats the goal of switching it from the textile store namespace to friends? I thought we already had a decent foundation and would just build that out

I made it because we use friends and groups but not conversations in sidebar, when we add a new friend conversation does not create and till we enter the chat a conversation will not create, but a friend can write us and the last message should be 
 changed. 
But if you think it is the wrong way, I can return it to textile

josephmcg:
Thanks Alex, I'm honestly not sure what the 'correct place' for it is. I think it would be ideal if it could be more centralized, rather than spread across multiple store modules. but that's probably a separate discussion altogether

molimauro:
Sara has the same problem too, @AlexZakablukov can you double check please?

stavares843:
correct tried here, and was the same

I was able to send images but not receiving them 🔨

AlexZakablukov:
done, check pls)

store/textile/getters.ts
```diff
@@ -12,6 +12,14 @@ const getters = {
       ? Object.values(state.conversations[address].messages)
       : []
   },
+  getConversationLastMessage:
+    (state: TextileState, getters: any) => (address: string) => {
+      const conversation = getters.getConversation(address)
+      if (!conversation) {
+        return null
+      }
```
molimauro, line 20: this can be removed since we have the `?`

components/views/user/User.html
```diff
@@ -9,6 +9,7 @@
   <div class="user-info">
     <TypographyTitle :size="6" data-cy="sidebar-user-name" :text="user.name" />
     <div
+      v-if="lastMessage"
```
molimauro, line 12: isnt this always true?

store/textile/mutations.ts
```diff
@@ -168,6 +168,17 @@ const mutations = {
...
+  setConversationLastMessage(
+    state: TextileState,
+    {
+      conversationId,
+      lastMessage,
+    }: { conversationId: string; lastMessage: MessageTrackerValues | null },
+  ) {
+    const conversations = { ...state.conversations }
+    conversations[conversationId].lastMessage = lastMessage
+    state.conversations = conversations
```
molimauro, line 180: are you spreading for reactivity right?

AlexZakablukov: Yes, I don't want to mutate the state

components/views/user/User.html
AlexZakablukov: you are right

store/textile/mutations.ts
molimauro: Ok perfect then!

store/textile/getters.ts
AlexZakablukov: fixed

components/views/user/User.html
molimauro: still can see this

AlexZakablukov: changed

stavares843:
https://user-images.githubusercontent.com/29093946/176716230-a73f044c-489d-45ef-bafe-856eac7da15b.mov

the messages were both showing 1:00 as a timestamp until user opened the conversation

AlexZakablukov:
fixed

phillsatellite (approved):

molimauro (approved):

WanderingHogan:
Closing this one and leaving the branch for now. We can come back to it when we get to it. This PR touches on a major change that iridium brings - the textile store and messages being stored in indexeddb are going away. We can come back to this close PR and branch though when we are able to port any needed changes to iridium

PR #3680, opened 2022-06-16
fix(friends): online status is hidden if not friend

vimercati-samir:
What this PR does 📖
Hide online status of an user if not a friend

[AP-1820](https://satellite-im.atlassian.net/browse/AP-1820)

Additional comments 🎤

### DIFF
--- a/components/ui/User/State/State.less
+++ b/components/ui/User/State/State.less
@@ -10,18 +10,13 @@
     --circle-radius: 10px;
   }
 
-  &.has-status {
-    .circle {
-      mask: radial-gradient(
-          circle calc(var(--circle-radius) + var(--border-width)) at 86% 86%,
-          #0000 98%,
-          #000
-        )
-        100% 100%/100% 100% no-repeat;
-    }
-  }
-
   .circle {
+    mask: radial-gradient(
+        circle calc(var(--circle-radius) + var(--border-width)) at 86% 86%,
+        #0000 98%,
+        #000
+      )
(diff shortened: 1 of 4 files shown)


phillsatellite (approved):

molimauro (approved):

stavares843 (changes requested):
dev vs this branch

Screenshot 2022-06-17 at 17 23 50

Screenshot 2022-06-17 at 17 23 32

vimercati-samir:
> dev vs this branch
> 
> Screenshot 2022-06-17 at 17 23 50
> 
> Screenshot 2022-06-17 at 17 23 32

Should be resolved in commit 020de50dff1ba2d9e12b9bb6515544e1173b0ebe

josephmcg (changes requested):
I think it would be simpler to just render UiCircle rather than UiUserState inside the friend component if send prop is true
components/views/friends/friend/Friend.html

What you're doing here is essentially forcing UiUserStateto render as if it were UiCircle. Also always add typescript return annotation to computed

vimercati-samir:
> I think it would be simpler to just render UiCircle rather than UiUserState inside the friend component if send prop is true components/views/friends/friend/Friend.html
> 
> What you're doing here is essentially forcing UiUserState to render as if it were UiCircle. Also always add typescript return annotation to computed

Agreed, changed accordingly in commit 68949dad6be2d2d1aa97bb4af1ced03cc33bc216

stavares843 (approved):

PR #3714, opened 2022-06-17
fix(chat): unknown name while loading

josephmcg:
**What this PR does** 📖
- show name instead of 'unknown' while loading messages

AP-1829

- this solution feels quick and dirty, should be refactored when time allows

**Additional comments** 🎤
- that `getUsernameFromState` is really slow and should be phased out


molimauro:
Agree lets add a ticket for the `getUsernameFromState ` thanks!

molimauro (approved):

stavares843 (approved):

PR #3759, opened 2022-06-20
test(store): create mock root state that will automatically keep in sync

josephmcg:
**What this PR does** 📖
- in an effort to make our test files readable by humans, make a central root state mock that can be imported into various other test files
- remove unused meta store namespace

AP-1763

**Additional comments** 🎤
- if you need to mock individual rootState values, just spread that specific namespace with the desired value


Jekrimo (approved):

WanderingHogan:
@drepram Can we hold this until you have a chance to review?

josephmcg:
@stavares843 I don't think that this specific store is necessary to update the title. you could just feed it a translation string based on the length of unread notifications. for example..

0 | any number

pageTitle: 'Satellite.im [maybe include files or chat name here] | Satellite.im [page name] ({notifCount})',

josephmcg:
added back for now, but we should consider a simpler approach
https://v3.nuxtjs.org/guide/features/head-management/

stavares843 (approved):

PR #3767, opened 2022-06-20
fix(conversation): offline status

aewing:
**What this PR does** 📖

**Additional comments** 🎤


josephmcg (approved):
I think when we setup pinia, we should set the default to not persist, then specify which properties we do want to persist

stavares843 (approved):

PR #3772, opened 2022-06-21
fix(files): update local files to match remote updates

josephmcg:
**What this PR does** 📖
- textile listener to clear out your local filesystem and import fresh on any remote updates
  - maybe one day we can just check the delta and make that change. this is still relatively fast and seems like a decent first step
  - i tried running the import without clearing first, skipping/adding files as needed, but it seemed relatively error prone
- add try catch to all files operations to catch any remote differences in case you're trying multiple operations at the same time on different devices. The out of date instance will cancel your operation and sync local with updated remote version
- move file system errors up a component so Browse component can adjust as needed
- update created user thread to match initial store value

AP-1477

- the files upload status being in the `ui` store module is weird. I don't know what I was thinking when I set that up. I would like to move it to Textile namespace after this

**Additional comments** 🎤
- I found a way to translate strings directly inside vuex, this was really helpful in reducing repetitive logic

### DIFF
--- a/components/views/files/controls/Controls.html
+++ b/components/views/files/controls/Controls.html
@@ -58,13 +58,7 @@
     :text="filesUploadStatus"
     :size="6"
   />
-  <div class="error-container" v-if="errors.length">
-    <alert-triangle-icon size="1.3x" />
-    <div>
-      <TypographyError v-for="error in errors" :key="error" :text="error" />
-    </div>
-    <x-icon size="1.3x" class="close" @click="errors=[]" />
-  </div>
+  <slot></slot>
   <div class="switcher-container">
     <TypographyText :size="6" :text="$t('pages.files.browse.files')" />
     <FilesViewSwitcher :changeView="changeView" />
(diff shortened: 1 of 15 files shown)


phillsatellite:
Tested: Found a few small things I ran by Sara and she wanted me to comment, I imported my account and had them both open, when I would delete file off of account 1, account 2 would not update until I refresh the browser page 

https://user-images.githubusercontent.com/93608357/175656537-15f81251-3d69-401d-8177-59da5e74c915.mov

Another issue I found was if Account 1 sent files in a chat, Account 2 would not update even after refreshing the browser 

https://user-images.githubusercontent.com/93608357/175656706-274721f7-20e9-4291-b519-dec764840291.mov

stavares843:
thanks @phillsatellite 🔨

josephmcg:
this doesn't deal with chat files at all, we don't have an index to keep track of those yet.

I was able to reproduce the account import bug. The second time I tried importing, it was working as intended. There's some logic in place to catch if the auto update failed. If you try submitting a change in the out of date window, it will cancel it and pull down latest changes. I'm not sure what caused it to stop listening for updates

josephmcg:
this is probably silly to work until we see how it works with iridium, closing. Dont delete branch yet though, there is some useful stuff I'd like to pull out

PR #3773, opened 2022-06-21
chore(vscode): default imports  - non relative

josephmcg:
**What this PR does** 📖
prefer imports from root w/ intelliense. this makes it easier to copy/paste if you want to add the import to another file. plus it complains less about import order if it's consistent.

will enforce a format like this
image

**Additional comments** 🎤


WanderingHogan (approved):
i dont think this will hurt anything

stavares843 (approved):

PR #3825, opened 2022-06-27
fix(chat): always show file size, fix overflow for long mime type

josephmcg:
**What this PR does** 📖
switch long file embeds to this
image
 instead of this
image

AP-1901

**Additional comments** 🎤


molimauro (approved):

stavares843 (changes requested):
Captura de ecrã 2022-06-29, às 15 44 20

dev vs this branch

if you dont have a subtitle, you have a hifen without any content on front, that way seems something is missing - first arrow

the ''q'' and ''p'' are cut off - second arrow

seed
end electric swear hurdle business priority salad ancient wire change please purpose

Chrome, macOS

josephmcg:
@stavares843 thanks, should be all better.
I noticed mobile embeds were flowing off the screen too, fixed that as well. I think it was introduced by my timestamp hover stuff

josephmcg:
I think I can clean this up after this goes through
https://github.com/Satellite-im/Core-PWA/pull/3858/files

phillsatellite:
Retested: Made sure I got most recent commits and some other stuff was fixed on here for me to verify but noticed the "p" was still cutoff
this branch
Screen Shot 2022-06-30 at 12 37 37 PM
 
 de
Screen Shot 2022-06-30 at 12 38 41 PM
v

josephmcg:
oh interesting, I'll have to take another look at that. we need to add overflow:hidden for the ellipsis feature to work, so that must be what's doing it. may i could adjust the line height

PR #3840, opened 2022-06-28
fix(search): add markdown to search results

josephmcg:
**What this PR does** 📖
title

AP-1905

**Additional comments** 🎤


phillsatellite:
tested: worked a lot better only one I couldn't get to appear in search results was messages with the Spoiler markdown

Screen Shot 2022-06-28 at 12 24 24 PM

josephmcg:
thanks Phill, that's a good bug. I think it's outside the scope of this ticket though. it has to do with how the search action + indexing is done. If you search for the markup string `||value||`, your result will appear. we need to adjust the search to pick up values that also match the various markup versions

molimauro (approved):

phillsatellite (approved):

stavares843 (changes requested):
on search i can click on the spoiler text if is the first result, but on the other pages, clicking on it doesn't work

https://user-images.githubusercontent.com/29093946/176491132-e083adf4-2914-4a8f-9687-f2b7d973c832.mov

seed
oval dignity twist coyote frozen exchange fiscal april doll conduct pudding meat

keyword when searching
Non nihil iusto consectetur quis nostrum.



Chrome, macOS

josephmcg:
oh yeah makes sense, It was only applying it on component mount. The search result is only mounted a single time, then the data changes with every new page. I added a watcher to check the page for spoilers every time the data changes

josephmcg:
Phill mentioned a similar issue above. I think that's outside the scope of this ticket. All this does is render markdown in search results. anything related to the actual search logic should be handled separately. I have a draft PR for that [here](https://github.com/Satellite-im/Core-PWA/pull/3857). I think I can find a better way to do it

stavares843 (approved):

PR #3847, opened 2022-06-28
fix(files): keyboard accessible, hit esc to close

josephmcg:
**What this PR does** 📖
- keyboard nav for files preview. It's difficult to see while we still disable outline on focus, but the tooltips should help a bit

AP-1909

This is ONLY the fullscreen file preview for filesystem. not chat preview, and not regular files page
image

**Additional comments** 🎤

### DIFF
--- a/components/views/files/view/View.vue
+++ b/components/views/files/view/View.vue
@@ -45,7 +45,7 @@ export default Vue.extend({
      * also takes a bit to get started for large files, this adds loading indicator
      */
     async download() {
-      // assign varaible in case the user closes modal and removes store value before download is finished
+      // assign variable in case the user closes modal and removes store value before download is finished
       const file = this.file
       if (file) {
         this.$store.commit('ui/addFileDownload', file.name)
--- a/components/views/files/view/View.vue
+++ b/components/views/files/view/View.vue
@@ -45,20 +45,22 @@ export default Vue.extend({
      * also takes a bit to get started for large files, this adds loading indicator
      */
     async download() {
-      if (this.file) {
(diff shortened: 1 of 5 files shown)


phillsatellite (approved):

components/views/files/view/View.vue
```diff
@@ -60,10 +63,18 @@ export default Vue.extend({
     },
     /**
      * @method share
-     * @description Emit to share item - pages/files/browse/index.vue
+     * @description copy link to clipboard
+     */
+    async share() {
+      this.$toast.show(this.$t('todo - share') as string)
+    },
```
molimauro, line 70: can we remove the async?

molimauro (changes requested):

components/views/files/view/View.vue
josephmcg: that's currently a dummy function that has no way of being called (ui button has been removed because it isn't ready). I imagine it will have async logic since it needs to interact with textile/iridium

josephmcg: commented out for now

molimauro: perfect

josephmcg:
oh interesting. maybe this was an existing bug, I'll add a check to prevent this

josephmcg:
should be resolved now. I also cleaned up some css since my other pr added a global rule

stavares843:
neat, thanks 💥 

yes could be a existing bug that was more reproducible with the esc button 🔨 i was only able to reproduce with the flow above

stavares843 (approved):

PR #3857, opened 2022-06-29
fix(search): make spoiler results appear

josephmcg:
**What this PR does** 📖
search results wasn't yielding spoiler versions of query

AP-1916

maybe there's a more elegant solution, let me know if you think of anything

**Additional comments** 🎤


josephmcg:
gonna try making this better, perhaps changing the fuzzy value

PR #3859, opened 2022-06-29
feat(modal): add fade animation

jasonwoodland:
**What this PR does** 📖
- Add fade in/out animation to modal

AP-1917

**Additional comments** 🎤


assets/styles/framework/animations.less
```diff
@@ -1,6 +1,7 @@
 // Animations
 @animation-speed: 0.05s;
 @animation-speed-micro: 0.05s;
+@animation-speed-slow: 0.25s;
```
josephmcg, line 4: could you clean these up while you're working on it? maybe remove micro and switch all instances of that to animation-speed

components/ui/Modal/Modal.less
```diff
@@ -92,3 +94,25 @@
...
+  .modal-background {
+    background-color: transparent !important;
+    backdrop-filter: blur(0px) !important;
+  }
+
+  .modal-card {
+    opacity: 0 !important;
+    transform: perspective(400px) rotateX(2deg) !important;
+    transform-origin: 50% 200%;
+    filter: @light-blur !important;
```
josephmcg, line 116: any way to avoid important? if this is overriding other important, try removing those instead

josephmcg (changes requested):

assets/styles/framework/animations.less
jasonwoodland: you mean consolidate all mentions of `@animation-speed-micro` to just `@animation-speed`?

I know they're the same value now, but maybe we might want to differentiate in the future? Or is it safe to do so?

components/ui/Modal/Modal.less
jasonwoodland: oops, forgot to remove that 😅

assets/styles/framework/animations.less
josephmcg: Yep, thats what i mean. If theyre the same value, they might as well use the same variable. If somebody wants to add a third speed later, they can

josephmcg (approved):

phillsatellite:
Found small UI issue when navigating from friends page to the Coming Soon modal looks like you can see the modal sliding in
https://user-images.githubusercontent.com/93608357/176947139-c53c0282-449c-4522-9173-f7562dbc8fa9.mp4

jasonwoodland:
@phillsatellite good spot, was this on safari? I think it was animating the transform-origin value which caused the glitch. Anyway I've moved the transform-origin outside of the animation ruleset.

phillsatellite:
> @phillsatellite good spot, was this on safari? I think it was animating the transform-origin value which caused the glitch. Anyway I've moved the transform-origin outside of the animation ruleset.

Hey @jasonwoodland! It was on Safari 🔥

josephmcg (approved):

PR #3874, opened 2022-06-30
fix(sidebar): add specific relative timestamps to sidebar

josephmcg:
**What this PR does** 📖
- timestamps for same day, yesterday, 2 days ago, and onward

AP-1918

**Additional comments** 🎤
- I couldn't justify adding an interval to every single conversation to add 'now' for 1 minute, until it's replaced by another timestamp 1 minute later. I think when we add unread message notifications, that will serve a similar purpose that 'now' would.

### DIFF
--- a/components/views/chat/search/result/item/Item.vue
+++ b/components/views/chat/search/result/item/Item.vue
@@ -22,13 +22,13 @@ export default Vue.extend({
       const msgTimestamp = this.$dayjs(this.data.at)
       // if today
       if (this.$dayjs().isSame(msgTimestamp, 'day')) {
-        return `${this.$t('search.result.today')} ${this.getTimestamp({
+        return `${this.$t('time.today')} ${this.getTimestamp({
           time: this.data.at,
         })}`
       }
       // if yesterday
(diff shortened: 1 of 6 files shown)


Jekrimo (approved):

molimauro (approved):

phillsatellite:
Hey Joe! Was just curious how long it should take to show sidebar as "Today" It changed from "now" to the time message was sent perfectly 
now
time

josephmcg:
Hey Phill, right now its configured to switch after 30 seconds. We can adjust that as needed

stavares843:
i believe what phil was asking was, when can we expect to see the ''today''?

https://github.com/Satellite-im/Core-PWA/pull/3874/files#diff-93b6f569059fddd160ddb58774259cd699eba83c03c74903cbe2097a7e146c14R42

seems is in code but you mentioned wouldn't appear

> If you're asking when it will switch to 'today', the answer is that it will not switch to that value.

either way, merging this for now 🔨

stavares843 (approved):

PR #3880, opened 2022-06-30
fix(scrollbar-style): scrollbars now disappear when not hovered on de…

ThomBos:
…sktop devices

**What this PR does** 📖
Hides scrollbar when not hovered by cursor
AP-1929

**Additional comments** 🎤

### DIFF
--- a/assets/styles/base.less
+++ b/assets/styles/base.less
@@ -30,19 +30,34 @@ body {
 *:active {
   outline: none;
 }
+
 * {
   &::-webkit-scrollbar {
     width: @scroll-bar-size;
+
+    :hover {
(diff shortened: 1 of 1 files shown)


phillsatellite (approved):

josephmcg:
i added pretty much the same rules in my draft scroll PR. I'd prefer if we could just handle it all there.

ThomBos:
> i added pretty much the same rules in my draft scroll PR. I'd prefer if we could just handle it all there.

ah... i see... i close this pr then 👍

ThomBos:
> i added pretty much the same rules in my draft scroll PR. I'd prefer if we could just handle it all there.

ah... i see... i close this pr then 👍

PR #3893, opened 2022-06-30
feat(friends): can click on a friend and open direct message

maljuburi:


aewing (approved):
some half-completed thoughts working toward chat here, but let's merge this so we can keep the ball rolling on the rebase & chat in another PR

PR #3918, opened 2022-07-05
feat(profile): remove photo when user already uploaded

nathan-power:
**What this PR does** 📖
- Allow user to remove profile picture in profile settings
- Correct an object undefined error
- Fix "add profile photo plus-button" from hovering above modals

AP-1840

**Additional comments** 🎤

### DIFF
--- a/components/views/settings/pages/profile/Profile.html
+++ b/components/views/settings/pages/profile/Profile.html
@@ -67,6 +67,9 @@
                   />
                 </div>
               </div>
+              <button class="remove-button" @click="removeProfileImage">
+                <span>{{ $t('pages.settings.profile.remove_image') }}</span>
+              </button>
               <div class="username">
                 <TypographyTitle :size="6" :text="accounts.details.name" />
               </div>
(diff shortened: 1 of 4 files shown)


components/views/settings/pages/profile/Profile.html
```diff
@@ -59,6 +59,11 @@
                 class="plus-button"
               />
             </div>
+            <div class="profile-photo-remove-button" v-if="!!src">
+              <button @click="removeProfileImage">
+                <span>{{ $t('pages.settings.profile.remove_image') }}</span>
+              </button>
+            </div>
```
josephmcg, line 66: could you get rid of the outer div here? It's not adding a lot of value. I think you can just slap the v-if and css rules on the button itself. other than that, good to go

josephmcg (approved):

components/views/settings/pages/profile/index.vue
```diff
@@ -126,6 +126,10 @@ export default Vue.extend({
         reader.readAsDataURL(files[0])
       }
     },
+    removeProfileImage(e: any) {
```
jasonwoodland, line 129: If we're not using `e` can we remove it?

PR #3921, opened 2022-07-05
fix(chatbar-focus): fixed conflict between chatbar and searchbar focus

ThomBos:
**What this PR does** 📖
Fixes search input's loss of focus caused by the chatbar 
AP-1949

**Additional comments** 🎤

### DIFF
--- a/components/views/chat/chatbar/Chatbar.vue
+++ b/components/views/chat/chatbar/Chatbar.vue
@@ -149,6 +149,9 @@ export default Vue.extend({
      * @description blur chatbar
      */
     blurChatbar() {
+      if (this.$device.isDesktop) {
+        return
+      }
       this.$refs.editable?.$refs.editable.blur()
     },
     /**
--- a/components/views/chat/chatbar/Chatbar.vue
+++ b/components/views/chat/chatbar/Chatbar.vue
@@ -149,7 +149,7 @@ export default Vue.extend({
      * @description blur chatbar
      */
     blurChatbar() {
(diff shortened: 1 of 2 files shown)


josephmcg (changes requested):
could you help me understand the goal of this? is it specifically for mobile?

ThomBos:
> could you help me understand the goal of this? is it specifically for mobile?

yeah it should be just for mobile but, i don't see the if that filters clickoutside tho...

josephmcg:
> > could you help me understand the goal of this? is it specifically for mobile?
> 
> yeah it should be just for mobile but, i don't see the if that filters clickoutside tho...

thanks, you could call a method with a guard clause.
`if (this.$device.isDesktop){
return
}
blur logic here`

WanderingHogan:
well that was a disaster

aewing (approved):

aewing:
@ThomBos can we cherry-pick the valid commits from this PR onto `iridium-dev` so we can close this PR?

stavares843:
> could you help me understand the goal of this? is it specifically for mobile?

the bug wasn't mobile only, was on desktop as well

[AP-1949](https://satellite-im.atlassian.net/browse/AP-1949)

josephmcg:
@stavares843 I think the initial bug here was caused by a 'fix' for chatbar keeping mobile focus when it should not.
We'll build this out again if we're still having the same issue later. there have been several changes to chat, especially when it comes to mobile

PR #3929, opened 2022-07-06
feat(files): unblock files ui, basic upload and download

josephmcg:
**What this PR does** 📖
- upload and download unencrypted files directly through Iridium.ipfs add and cat
  - https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md
- remove textile bucket init
- move nsfw check and heic check inside iridium upload methods. I will do this for chat upload as well once we're ready. My goal was to keep the component code as simple as possible, and let the classes handle more complex logic
  - also combined these two checks, so you only need to call nsfw check. will convert to heic for you as needed
- I really like how the IridiumManager class was setup to be used across different files. I took a similar approach with FilSystem and removed the Vue prototype from the classLoader
- our byte check package only returns `image/heif` for any type of heic, so I adjusted the enum to reduce bloat
- move blobToStream to a utility file. consolidate with other blobToBase64 utility
- files import/export

AP-1950

- will be switched to encrypted files next, just wanted to unblock ui and get something working.
- I would also like to move the types around, specifically the file types enum, since that is used all over the app.

**Additional comments** 🎤


libraries/Files/types/file.ts
```diff
@@ -72,8 +72,7 @@ export enum FILE_TYPE {
 
   // image - non-embeddable
   BMP = 'image/bmp',
-  HEIC = 'image/heic', // non-embeddable, but we convert as needed
-  HEIF = 'image/heif', // non-embeddable, but we convert as needed
+  HEIC = 'image/heif', // non-embeddable, but we convert as needed
```
molimauro, line 75: This is on purpose right? i mean the heif on the right

josephmcg: yeah I added a comment in the pr description, but the byte check only returns 'image/heif'. If you'd rather name our enum HEIF, we can

molimauro: Oh sorry i missed that! Good then :)

molimauro:
Gonna merge this

PR #3935, opened 2022-07-08
fix(files): remove leftover ui blockers and fix exports

josephmcg:
**What this PR does** 📖
- I missed a few ui blockers in my last PR, this removes any remaining ones. I think it was only a concern with textile having an outdated root. with a direct ipfs connection, I don't think there's a concern about concurrent operations
- export fs after every files operation (like, rename, delete). missed these in my last PR as well

I'm going to attempt to rearchitect this to work more smoothly with our new stack, but this at least returns all functionality while that work is in progress

AP-1953

**Additional comments** 🎤

### DIFF
--- a/components/views/files/aside/List/List.html
+++ b/components/views/files/aside/List/List.html
@@ -1,8 +1,8 @@
 <ul class="menu-list">
   <li
     v-for="item in items"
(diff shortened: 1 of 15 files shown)


WanderingHogan (approved):

josephmcg:
im just gonna roll w/ this branch and try building it out here

PR #3942, opened 2022-07-08
fix(signup): fix glitch

molimauro:
**What this PR does** 📖
When registering there is a small amount of time in which the login page (not the signup) is shown, causing a bad UX. This should fix that.

**Additional comments** 🎤

### DIFF
--- a/pages/auth/unlock/index.vue
+++ b/pages/auth/unlock/index.vue
@@ -130,7 +130,7 @@ export default Vue.extend({
     async deleteAccount() {
       await this.$store.dispatch('settings/clearLocalStorage')
     },
-    async action() {
+    action() {
       this.status = 'loading'
       return this.step === 'login' ? this.decrypt() : this.create()
     },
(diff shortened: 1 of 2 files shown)


josephmcg:
I can't login on this branch
image

pages/auth/unlock/index.vue
```diff
@@ -14,13 +14,17 @@ export default Vue.extend({
...
+  } {
     return {
       pin: '',
+      status: 'idle',
       error: '',
-      decrypting: false,
-      peer: null,
-      showChangeLog: false,
+      step: 'signup',
     }
```
josephmcg, line 28: I typically just cast the data as type
`status: 'idle' as 'idle' | 'loading'`

molimauro: Do you think its too verbose? I'll change then for consistency :)

josephmcg: I think either way is ok, I prefer fewer lines whenever possible. less scrolling lol

PR #3946, opened 2022-07-11
feat(iridium): add settings manager

jasonwoodland:
**What this PR does** 📖
- Implement iridium SettingsManager
- Refactor IridiumManager so that the managers are not typed as optional (removes need for null checking at component-level)
- Move the following settings to iridium:
  - Theme/flair
  - Flip video
  - Notification sounds
  - Keybinds
  - Privacy

AP-1952

**Additional comments** 🎤


assets/styles/base.less
```diff
@@ -386,7 +386,7 @@ button {
 @media not all and (min-resolution: 0.001dpcm) {
   @supports (-webkit-appearance: none) {
     ::selection {
-      background: rgba(var(--flair-color-rgb), 0.99) !important;
```
molimauro, line 389: the `rgba(var(--flair-color-rgb), 0.99) !important` was needed for a text selection bug in Safari

molimauro (changes requested):

assets/styles/base.less
josephmcg: I think he meant to switch this to a new variable. Rather than keep track of rbg, we decided to append 'fe' to the regular primary color, making an 8 digit hex code that achieves the same thing

jasonwoodland: @molimauro @josephmcg you're right I did forget to update the variable name here. I decided a better solution would be to generate the rgb value instead so we can keep it like this with less overhead. See https://github.com/Satellite-im/Core-PWA/pull/3946/commits/cead93821ee9c751986a86f1b84eb35157b5bb39

josephmcg (approved):
excellent work Jason, sets a strong foundation and creates some great utilities that can be implemented in other iridium modules

molimauro (approved):

PR #3953, opened 2022-07-12
feat(settings-profile): account link component/markup

nathan-power:
**What this PR does** 📖
Add the markup for the accounts list in the profile settings page.

AP-1957
https://satellite-im.atlassian.net/browse/AP-1957

As current design in Figma had a few layout restraints, the design was tweaked after discussion with Joe so that it is now easier to access all elements of the original design on all devices.

**Additional comments** 🎤


josephmcg (approved):
image

PR #3955, opened 2022-07-12
fix(friends-requests): improved requests and notifications display

ThomBos:
**What this PR does** 📖
Fixed friends requests display
AP-1954

**Additional comments** 🎤
Still missing some data to show (name and image, just when the other user hasn't accepted the request)

### DIFF
--- a/components/views/friends/add/Add.vue
+++ b/components/views/friends/add/Add.vue
@@ -23,11 +23,14 @@ export default Vue.extend({
       user: null as User | null,
     }
   },
-  mounted() {
+  async mounted() {
     if (this.$route.params && this.$route.params.id) {
       this.$data.friendId = this.$route.params.id
       this._searchFriend()
     }
+    iridium.friends?.on('request/error', (err) => {
+      this.error = err
+    })
   },
   methods: {
     _searchFriend: debounce(async function (this: any) {
@@ -61,11 +64,7 @@ export default Vue.extend({
       }
       this.searching = false
     },
-    onFriendRequestSent(error?: string) {
-      if (error) {
-        this.error = error
-        return
-      }
+    onFriendRequestSent() {
       this.request = null
       this.user = null
       this.friendId = ''
--- a/components/views/friends/add/Add.vue
+++ b/components/views/friends/add/Add.vue
@@ -52,7 +52,7 @@ export default Vue.extend({
         this.error = this.$t('friends.self_add') as string
         return
       }
-      const hasFriend = iridium.friends?.isFriend(friendId)
+      const hasFriend = iridium.friends.isFriend(friendId)
       if (hasFriend) {
         this.error = this.$t('friends.already_friend') as string
       }
--- a/components/views/friends/add/Add.vue
+++ b/components/views/friends/add/Add.vue
@@ -54,18 +54,11 @@ export default Vue.extend({
         this.error = this.$t('friends.already_friend') as string
       }
 
-      try {
-        const request = await iridium.friends?.requestCreate(friendId)
-        this.user = {
-          did: friendId,
-          name: friendId,
-          status: 'offline',
-        }
-        this.request = request
-      } catch (error) {
-        this.error = this.$t('friends.invalid_id') as string
+      this.user = {
+        did: friendId,
+        name: friendId,
+        status: 'offline',
       }
-
       this.searching = false
     },
     onFriendRequestSent(error?: string) {
--- a/components/views/friends/friend/Friend.vue
+++ b/components/views/friends/friend/Friend.vue
@@ -67,9 +67,13 @@ export default Vue.extend({
   methods: {
     async createFriendRequest() {
       this.loading = true
-      const req = await iridium.friends?.requestCreate(this.user.did)
+      try {
+        const req = await iridium.friends?.requestCreate(this.user.did)
+      } catch (err) {
+        return this.$emit('requestSent', err)
+      }
       this.loading = false
-      this.$emit('requestSent', typeof req === 'string' && req)
+      this.$emit('requestSent')
     },
     async acceptFriendRequest() {
       this.loading = true
--- a/components/views/friends/friend/Friend.vue
+++ b/components/views/friends/friend/Friend.vue
@@ -67,7 +67,7 @@ export default Vue.extend({
   methods: {
     async createFriendRequest() {
       this.loading = true
-      await iridium.friends?.requestCreate(this.user.did)
+      await iridium.friends?.requestInsert(this.user.did)
       this.loading = false
       this.$emit('requestSent')
     },
@@ -83,7 +83,7 @@ export default Vue.extend({
     },
     async removeFriend() {
       this.loading = true
-      await iridium.friends?.remove(this.user.did)
+      await iridium.friends?.removeFriend(this.user.did)
       this.loading = false
     },
     async cancelRequest() {
--- a/components/views/friends/friend/Friend.vue
+++ b/components/views/friends/friend/Friend.vue
@@ -67,11 +67,7 @@ export default Vue.extend({
   methods: {
     async createFriendRequest() {
       this.loading = true
-      try {
-        const req = await iridium.friends?.requestCreate(this.user.did)
-      } catch (err) {
-        return this.$emit('requestSent', err)
-      }
+      await iridium.friends?.requestCreate(this.user.did)
       this.loading = false
       this.$emit('requestSent')
     },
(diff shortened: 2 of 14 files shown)


components/views/friends/add/Add.vue
```diff
@@ -54,18 +54,11 @@ export default Vue.extend({
...
-        const request = await iridium.friends?.requestCreate(friendId)
-        this.user = {
-          did: friendId,
-          name: friendId,
-          status: 'offline',
-        }
-        this.request = request
-      } catch (error) {
-        this.error = this.$t('friends.invalid_id') as string
+      this.user = {
```
ThomBos, line 57: i think we should add a check function here to see if the id exists and show the confirm button (now we are just showing it, after clicking it we check the id)
the previous version sent automatically the request when the id was valid

libraries/Iridium/friends/FriendsManager.ts
```diff
@@ -322,7 +322,7 @@ export default class FriendsManager extends Emitter<IridiumFriendPubsub> {
     }
     const list = this.state?.list?.filter((f) => f !== friendId)
     await this.set('/list', list)
-    await this.set(`/details/${friendId}`, undefined)
+    await this.set(`/details/${friendId}`, null)
```
ThomBos, line 325: this doesn't really remove the friend (just removes from the display) 😬 
other user doesn't get any event or notification

molimauro: because is still not implemented in iridium?

ThomBos: no idea, i have to ask Drew, but i don't see any other function that removes friends

ThomBos: > because is still not implemented in iridium?

the only thing i know that's missing rn on iridium (so in this pr), is the name display when sending a friend request (we can see just the id)

libraries/Iridium/friends/FriendsManager.ts
```diff
@@ -193,11 +193,11 @@ export default class FriendsManager extends Emitter<IridiumFriendPubsub> {
   async requestCreate(remotePeerDID: string, incoming = false, user: User) {
     if (this.isFriend(remotePeerDID)) {
       logger.error('iridium/friends', 'already a friend', { remotePeerDID })
-      throw new Error(FriendsError.FRIEND_EXISTS)
+      return FriendsError.FRIEND_EXISTS
     }
     if (this.hasRequest(remotePeerDID)) {
       logger.error('iridium/friends', 'request already exists')
-      throw new Error(FriendsError.REQUEST_ALREADY_SENT)
+      return FriendsError.REQUEST_ALREADY_SENT
```
josephmcg, line 200: why return errors rather than throw?

ThomBos: i need to show those error strings in the ui

components/views/friends/friend/Friend.vue
```diff
@@ -67,8 +67,13 @@ export default Vue.extend({
   methods: {
     async createFriendRequest() {
       this.loading = true
-      await iridium.friends?.requestCreate(this.user.did)
+      try {
+        const req = await iridium.friends?.requestCreate(this.user.did)
+      } catch (err) {
+        return this.$emit('requestSent', err)
```
aewing, line 73: Do we still need this emitter? Could we just listen to the emitted event on FriendsManager?

ThomBos: yeah i think it would work the same

components/views/navigation/sidebar/Sidebar.vue
```diff
@@ -45,23 +45,30 @@ export default Vue.extend({
...
-      friends: () => iridium.friends?.state,
       groups: () => iridium.groups?.state,
       conversations: () => ({}),
     }),
   },
   async mounted() {
-    iridium.friends?.on('request/changed', () => this.$forceUpdate())
+    iridium.friends?.on('request/changed', () => {
+      this.friends = { ...iridium.friends?.state }
+    })
```
josephmcg, line 71: are we fully replacing the state value inside iridium anytime a request change comes in? Then we lose the observer in the component (setup in data). If you did a deepMerge, or just added the new request data to store individually, you wouldn't need to do this with the observers.

Jason implemented the deepMerge approach here
https://github.com/Satellite-im/Core-PWA/pull/3946

components/views/friends/friend/Friend.vue
```diff
@@ -67,8 +67,13 @@ export default Vue.extend({
...
+      }
       this.loading = false
+      this.$emit('requestSent')
```
josephmcg, line 76: I think we should just emit the did and pass it to a method in the parent component. Then that component would handle the try catch. Having an optional error param in the parent make it more complicated than it needs to be

josephmcg (changes requested):

components/views/navigation/sidebar/Sidebar.vue
ThomBos: a deepmerge seem a good solution, but can it work if it's done on the friendsmanager? (since it's there where we manage requests)

components/views/friends/friend/Friend.vue
ThomBos: yeah, i'm going to remove this and add what you and Drew suggested 👍

components/views/navigation/sidebar/Sidebar.vue
josephmcg: Good point, i'll take a closer look at this tomorrow

ThomBos:
@josephmcg let me know what you think about these changes! 🔨

components/views/friends/add/Add.vue
```diff
@@ -23,11 +23,14 @@ export default Vue.extend({
...
   },
-  mounted() {
+  async mounted() {
     if (this.$route.params && this.$route.params.id) {
       this.$data.friendId = this.$route.params.id
       this._searchFriend()
     }
+    iridium.friends?.on('request/error', (err) => {
+      this.error = err
+    })
```
josephmcg, line 33: would it be simpler to wrap the possible error creating methods in a try/catch (or promise.catch if async)?

components/views/navigation/sidebar/Sidebar.vue
josephmcg, line 71: I see why this is needed with the current architecture. Vue observers can't pickup newly created objects. If we switched details & requests to arrays instead, we could get rid of this
image

libraries/Iridium/friends/FriendsManager.ts
```diff
@@ -191,32 +193,38 @@ export default class FriendsManager extends Emitter<IridiumFriendPubsub> {
...
+      const request: FriendRequest = {
+        user: user || { did: remotePeerDID, name: remotePeerDID },
+        status: 'pending',
+        incoming,
+        at: Date.now(),
+      }
+      return this.requestSave(remotePeerDID, request, incoming)
+    } catch (err) {
+      this.emit('request/error', err)
     }
```
josephmcg, line 219: I mentioned in a similar comment above, but I think it would be easier to directly throw the error, then let the component method handle it

josephmcg (changes requested):
a couple ideas regarding FriendsManager state structure. If we switched details and requests to arrays, it would simplify a lot of this logic and improve Vue reactivity

components/views/navigation/sidebar/Sidebar.vue
ThomBos: yeah, currently i've removed details and moved all it's content into a list[user object].
also we are keeping requests just if pending and it's now an array

libraries/Iridium/friends/FriendsManager.ts
ThomBos: don't look at the code now, because it's going to change pretty soon, but the logic is the same... so i can tell you that i'm keeping the try catch also on the new code, but the reason is that both user are calling that function to create the request

ThomBos:
**Quick recap:**

- all states are now arrays instead of objects, removed details state because it's now stored into list
- requestSave clears non-pending requests and sends requests change events using requestSendUpdate
- removeFriend is disconnected from requestSendUpdate (but it does basically the same thing) because 'accepted' or 'rejected' requests are no more stored in the state
- removed unused functions

**Errors management is still in progress and has to be improved.** 🔨
At the moment i'm thinking of adding a new state to contain an error string, so the components can render that as it changes, this would cause to all throw to be removed though, in order to remove all try/catches around the app... 
 

**Any suggestion to improved this code are well welcomed!!** 💥🚀

PR #3956, opened 2022-07-13
refactor(imageCropper): reduce anti pattern of passing functions as props

nathan-power:
**What this PR does** 📖
- reduce anti pattern of passing functions as props
- adjust config (config.ts) to use blob rather than base64
- reduce height & width in config.ts

AP-1956

**Additional comments** 🎤


josephmcg:
will review again after we rip out the blob -> base64 -> conversion for the cropper

josephmcg (approved):

PR #3991, opened 2022-07-14
feat(wallet): phantom integration on the Iridium-dev

JustZacca:
As the name implies, this PR introduces Phantom Wallet to the Satellite, allowing users to sign up and run Iridium with it.
This PR also modifies the classic initialization of Iridium.

Now both by connecting the external wallet and using the internal wallet, iridium starts up by signing a message.

I corrected the tests that I could correct, those related to my PR, but according to Manuel's directives, I completely disabled the non-working tests.

This PR is temporary, as much as it does its job, it is only a small part, as we still have to modify the store and part of the workflow.


store/accounts/actions.ts
```diff
@@ -255,8 +269,8 @@ export default {
       photoHash: imagePath,
       address: walletAccount.publicKey.toBase58(),
     })
-    dispatch('initializeEncryptionEngine', walletAccount)
-    dispatch('startup', walletAccount)
+    // dispatch('initializeEncryptionEngine', walletAccount)
```
stavares843, line 272: do we need this commented out? 🔨

libraries/Phantom/PhantomFriends.ts
```diff
@@ -0,0 +1,678 @@
...
+      this.buildEventHandler(FriendsEvents.NEW_REQUEST),
+      Config.solana.defaultCommitment,
+      [incomingRequestsFilter],
+    )
+
+    // Filter for new friends checks only if an outgoing request has been accepted
+    // because we suppose the incoming request acceptance to be catch directly after the
+    // success of the acceptRequest action
+    // This filter checks the sender public key (our) and the status
+    // [32 bytes (sender public key)][1 byte (status)][32 bytes (recipient public key)]
```
stavares843, line 576: maybe we could use multi-line comment instead for this

stavares843 (commented):
can you skip the tests using

image


instead of renaming pls? thanks 🔨

iltumio:
> > can you skip the tests using
> > image
> > instead of renaming pls? thanks 🔨
> 
> I changed the name of the files according to Manuela's instructions, but if you want I'll use skip instead

the rename is better because if we have errors like missing imports due to a whole library change it will fail even if you skip it

stavares843:
yes i know those was the instructions before 🔨 , but in case there are no imports missing or anything is better to just skip the test and just renaming only if really needed

stavares843:
this PR is re-adding things that are already on iridium-dev so some rebase was made incorrectly

PR #4012, opened 2022-07-18
feat(wallet): phantom integration on the Iridium-dev

JustZacca:
As the name implies, this PR introduces Phantom Wallet to the Satellite, allowing users to sign up and run Iridium with it.
This PR also modifies the classic initialization of Iridium.

Now both by connecting the external wallet and using the internal wallet, iridium starts up by signing a message.

I corrected the tests that I could correct, those related to my PR, but according to Manuel's directives, I completely disabled the non-working tests.

This PR is temporary, as much as it does its job, it is only a small part, as we still have to modify the store and part of the workflow.

### DIFF
--- a/libraries/BlockchainClient/BlockchainClient.ts
+++ b/libraries/BlockchainClient/BlockchainClient.ts
@@ -4,7 +4,6 @@ import {
   RpcResponseAndContext,
   SignatureResult,
 } from '@solana/web3.js'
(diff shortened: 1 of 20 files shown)


aewing (approved):
Nice work!!!

WanderingHogan (approved):
so neat

JustZacca:
Thanks!

PR #4024, opened 2022-07-20
refactor(friendsmanager): different error handling aproach

ThomBos:
**What this PR does** 📖
Improves error handling in the friendsmanager 

**Additional comments** 🎤

### DIFF
--- a/components/views/friends/add/Add.html
+++ b/components/views/friends/add/Add.html
@@ -15,12 +15,7 @@
   />
   <TypographyError v-if="friendRequestError" :text="$t(friendRequestError)" />
   <UiLoadersLoadingBar v-else-if="searching" />
-  <FriendsFriend
-    v-else-if="user && user.did"
-    :user="user"
-    @requestSent="onFriendRequestSent"
-    isPreview
-  />
+  <FriendsFriend v-else-if="user && user.did" :user="user" isPreview />
   <!-- <UiSpacer :height="10" />
     <TypographyHorizontalRuleText plaintext value="OR" />
     <UiSpacer :height="10" />
--- a/components/views/friends/add/Add.html
+++ b/components/views/friends/add/Add.html
@@ -13,7 +13,7 @@
     ref="input"
     autofocus
   />
-  <TypographyError v-if="error" :text="$t(error)" />
+  <TypographyError v-if="friendRequestError" :text="$t(friendRequestError)" />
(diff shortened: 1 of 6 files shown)


ThomBos:
i was wondering if changing the error's display from text to a popup (like the "toast" component but red), was a good idea or not.... changing this here would mean change errors display all around the app. 
was thinking about this because errors, like lost connection, can occur while the user is on any page and he doesn't have any feedback whenever this happens

josephmcg (changes requested):
i don't understand the design decision of throwing an error and catching it inside the same method. Throwing errors are helpful if you intend to bubble it up and catch it elsewhere.

An example of the above technique can be found in this method `validateName` (https://github.com/Satellite-im/Core-PWA/blob/iridium-dev/libraries/Iridium/files/FilesManager.ts#:~:text=*/-,private,-validateName(name). It will throw an error which is then caught all the way back in `components/views/files/controls/Controls.vue` `addFolder` and `handleFile`

instead of try/catch in the same method, you may as well just manually set error to whatever value, then return.

*more importantly - With regards to the overall structure. Could you explain your thought process?

My thoughts are that an error pertains to a single friends operation (add, send request, accept, etc), so you don't need to keep the error stored in iridium. You should be able to throw it, and then let the component handle it

ThomBos:
I'd have gone the way you said if the components which display the error were made in a different way... I've tried as well returning error strings instead of throwing them, but the "issue" is still the same.
Right now we have a child component which calls friendsmanager's functions (it would be the place where to catch the errors) and the parent component that reads the errors computed property, then displays it.
I went for this solution mainly to avoid emitting all the stuff from the child component...

So actually i have no idea how to handle it avoiding errors emits in the child component (or friendmanager) and get the same try catch process you suggested

josephmcg:
gotcha, I can take a look on monday to see if there's a simpler approach. If it can't wait until then, maybe Mauro would be able to guide you in the right direction

josephmcg:
since Thomas is out of office, I didn't check this today. will fix sometime this week. I was more focused on adding new functionality in chat

ThomBos:
> since Thomas is out of office, I didn't check this today. will fix sometime this week. I was more focused on adding new functionality in chat
> 
> -- edit because the friends connection stuff is a bit wonky, decided it was better to wait

Since lot as changed with Muhammad's work, is this Pr still useful or it can be deleted?

PR #4033, opened 2022-07-22
feat(chat): create groups and render in sidebar

josephmcg:
**What this PR does** 📖
- refactor userpicker to use iridium friends list
- refactor and move GroupIcon component
  - change size to be the same as UiUserState (36px) instead of 40
  - simplify styles and html
- flesh out quickchat popup
  - remove render condition from store since its only handled in sidebar component
  - use v-model.trim to avoid needing to trim in guard clause (learn something new everyday)
- remove unused GroupClickable now that the sidebarItem component exists

AP-2002

**Additional comments** 🎤


components/interactables/UserPicker/UserPicker.vue
```diff
@@ -21,17 +20,17 @@ export default Vue.extend({
...
-        if (!this.exclude.length) {
-          return all
-        }
-        return all.filter((friend) => !this.exclude.includes(friend.address))
-      },
-    }),
+    friends() {
+      if (!this.exclude.length) {
+        return this.friendsList
+      }
```
molimauro, line 29: isnt this reduntant given the filter right after?

josephmcg: I had the same thought when I gave my initial review lol.
Jason mentioned a performance benefit if we don't do an unnecessary filter

josephmcg:
oh man i wrote up some longwinded explanation of everything here in the PR description but forgot to save, 1 sec

components/interactables/UserPicker/UserPicker.vue
molimauro: Oh okay nice!

WanderingHogan (approved):

PR #4046, opened 2022-07-25
feat(chat): display uploaded files in ui

josephmcg:
**What this PR does** 📖
- display uploaded files in conversation (kept in chat vuex store tied to conversation id) 
- drag and drop functionality for chat and personal files uploads
  - hook up consent scan check to iridium settings manager and created action to simplify that logic
- paste file functionality for chat file uploads
- refactor chatbar css
  - there was a missing class container, so links weren't getting properly styled.
  - I removed a lot of unnecessary css and switched that whole area to flex

AP-2008, AP-2012

- files will not send, that will be separate functionality

**Additional comments** 🎤


components/views/chat/chatbar/Editable.vue
```diff
@@ -325,19 +321,20 @@ export default Vue.extend({
...
+  &:extend(.font-primary);
 
   .placeholder {
     color: @text-muted;
     position: absolute;
-    white-space: nowrap;
-    text-overflow: ellipsis;
-    overflow: hidden;
-    left: 0;
-    right: 0;
```
molimauro, line 340: potentially can be longer than the width in the future, thats why the ellipsis, but can easily circle back later

josephmcg:
@molimauro i thought this was consistent with the current dev branch. i'll take another look tomorrow

components/views/chat/chatbar/Editable.vue
josephmcg: true, I can add the utility class there

josephmcg:
this appears to be consistent with dev branch.

The scope of this is already quite large, so I think we should improve chatbar styles later

top 2 -dev
bottom 2 -this branch

image

image

image

image

jasonwoodland (approved):

molimauro:
Strange, the cursor appears different on my screen from dev and this, also the height. In any case no problem, we can track this and come back later, thanks for checking ;)

PR #4057, opened 2022-07-25
fix(friends): fix crash when removing a friend

vimercati-samir:
What this PR does 📖
Fix crash when removing a friend

[AP-2021](https://satellite-im.atlassian.net/browse/AP-2021)

Additional comments 🎤


josephmcg:
The goal of this component was to keep friends and groups accessible in the sidebar. This appears to be filtering out groups.

Instead, we should change the removeFriend method to work and conditionally change the options based on the conversation type. Although I believe this task falls within the scope of my teams current effort, considering I created a ticket for this task yesterday https://satellite-im.atlassian.net/browse/AP-2015

Closing this out

PR #4065, opened 2022-07-26
fix(sidebar): conditionally render context menu items based on type

josephmcg:
**What this PR does** 📖
- conditionally render sidebar context menu items based on group vs friend

AP-2015

- remove friend may be a little buggy. will fix separately 

**Additional comments** 🎤


molimauro (approved):

molimauro:
Samir is fixing removing friend ;)

josephmcg:
> Samir is fixing removing friend ;)

it involves removing from the conversation manager too. I'd prefer if we can just handle it some other day

PR #4103, opened 2022-08-02
fix(images): implement checkered background for transparent images

nathan-power:
**What this PR does** 📖
- Adds a ImageContainer component for dynamic styling (nsfw styling)
- Adds checkered background for transparent svgs and pngs (in ImageContainer component)

AP-2040

**Additional comments** 🎤

### DIFF
--- a/components/ui/ImageContainer/ImageContainer.vue
+++ b/components/ui/ImageContainer/ImageContainer.vue
@@ -7,13 +7,9 @@
 <script>
 export default {
   props: {
-    src: {
-      type: String,
-      required: true,
-    },
     nsfw: {
       type: Boolean,
-      required: false,
+      default: false,
     },
   },
 }
--- a/components/ui/ImageContainer/ImageContainer.vue
+++ b/components/ui/ImageContainer/ImageContainer.vue
@@ -1,0 +1,47 @@
+<template>
+  <div class="image-container" :class="{ 'image-container-nsfw': nsfw }">
+    <slot />
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    src: {
+      type: String,
+      required: true,
+    },
+    nsfw: {
+      type: Boolean,
+      required: false,
+    },
+  },
+}
+</script>
+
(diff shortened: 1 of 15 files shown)


components/ui/ImageContainer/ImageContainer.vue
```diff
@@ -0,0 +1,47 @@
...
+      #eee 25%,
+      transparent 25%,
+      transparent 75%,
+      #eee 75%
+    ),
+    linear-gradient(45deg, #eee 25%, transparent 25%, transparent 75%, #eee 75%);
+  background-size: 16px 16px;
+  background-position: 0 0, 8px 8px;
+  filter: brightness(0.75);
+
```
josephmcg, line 37: i think we don't necessarily want this brightness filter on every image. I added it to files grid specifically so you could have enough contrast to click the heart overlay button.

although I think I removed a z-index value without throughly testing because it was coming through to modals. maybe we can just change the grid item markup quick-actions to the bottom of the container so it doesn't need z index

components/views/files/view/View.less
```diff
@@ -51,7 +51,7 @@
     max-height: 60vh;
     &:extend(.round-corners);
     overflow: hidden;
```
josephmcg, line 53: can this overflow be removed?

josephmcg (changes requested):
good, a couple minor things. Could you remove the blur-image utility class too? since we're just applying blur to the container now

components/ui/ImageContainer/ImageContainer.vue
nathan-power: > maybe we can just change the grid item markup quick-actions to the bottom of the container so it doesn't need z index
Yeah sounds like a good idea. In that case would we need the brightness filter for the images?

josephmcg: Try uploading some white images and see how the white heart on hover looks. If its enough contrast, feel free to leave it out

josephmcg (approved):

PR #4112, opened 2022-08-02
fix(buttons): fixed loader on friends cards

ThomBos:
**What this PR does** 📖
Substitue buttons loader in the friends cards with the loader that is already in the chat and main screen
AP-2058

**Additional comments** 🎤

### DIFF
--- a/assets/styles/base.less
+++ b/assets/styles/base.less
@@ -286,6 +286,22 @@ button {
 
   #custom-cursor-area {
     &:extend(.font-primary);
+
+    .spinner {
+      display: inline-flex;
+      align-items: center;
+      justify-content: space-between;
+      width: 60px;
+
+      .loading-bar.button-loader-bar .bar {
+        animation: buttonloading 2s cubic-bezier(0.17, 0.37, 0.43, 0.67)
+          infinite;
+      }
+
+      .spin {
+        margin-right: @xlight-spacing;
+      }
+    }
   }
 
   &.is-dark.is-outlined {
--- a/assets/styles/base.less
+++ b/assets/styles/base.less
@@ -286,22 +286,6 @@ button {
 
   #custom-cursor-area {
     &:extend(.font-primary);
-
-    .spinner {
-      display: inline-flex;
-      align-items: center;
-      justify-content: space-between;
-      width: 60px;
-
-      .loading-bar.button-loader-bar .bar {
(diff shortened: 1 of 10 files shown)


components/interactables/Button/Button.vue
```diff
@@ -73,7 +73,7 @@ export default Vue.extend({
      */
     loadingText: {
       type: String,
-      default: 'Loading...',
+      default: null,
```
molimauro, line 76: is this text still needed?

ThomBos: mmhh need to check around the app if this loader is triggered.... going for it

assets/styles/base.less
```diff
@@ -640,6 +640,15 @@ button {
...
 
+.friend {
+  .loader-container {
+    width: 60px;
+    .loading-bar .bar {
+      animation: shorterloading 2s cubic-bezier(0.17, 0.37, 0.43, 0.67) infinite;
+    }
+  }
+}
+
```
josephmcg, line 651: we want to slim down this file as much as possible, could you add this to scoped component css?

ThomBos: i've tried to do that, but it couldn't overwrite the animation

josephmcg: Sounds like its better off as a new loading component then. Or if its really similar to the other one, you can pass a prop

josephmcg: If its using the extend stuff, that often complicates the css priority. I try to avoid if possible

molimauro: I'd avoid as well, just use a prop

ThomBos: yeah the prop is definitely better, thanks for clarifying! 👍

components/ui/Loaders/LoadingBar/LoadingBar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="loading-bar">
+  <div class="loading-bar" :class="short ? 'short' : ''">
```
josephmcg, line 2: a ternary is good if there are two values that could be applied. Try this syntax instead
```
:class="{short : short}"
```

ThomBos: it also looks much better while reading the code 👍

PR #4145, opened 2022-08-08
refactor(text): improve basic text component

josephmcg:
**What this PR does** 📖
This is a work in progress. All text components have been adjusted, along with a few Titles/subtitles. I would like to continue that separately 

- I thought using 'primary' and 'secondary' for font families was somewhat arbitrary. We end up using the 'secondary' family more frequently. Would like to switch this to 'heading' and 'body'. will replace global rules in a future PR
- font colors were using white all over the place. updated colors to match figma. will remove incorrect global colors later
- switched to slot based instead of a prop. this will be way more flexible

AP-526

**Additional comments** 🎤


nathan-power (approved):
!(LGTM)[https://storage.googleapis.com/lgtm-cat-cats/9b6d5d0a-5811-40b6-b33b-9d905be2a6f2.gif]

PR #4162, opened 2022-08-09
fix(friends): reactivity when adding friends

josephmcg:
**What this PR does** 📖
- remove Object.values getter on friendsmanager that breaks reactivity. update components accordingly. I noticed when you do an iridium.set before Vue.set, reactivity breaks

AP-2073

**Additional comments** 🎤


nathan-power (approved):
LGTM

PR #4186, opened 2022-08-10
fix(modal): modal animation

josephmcg:
**What this PR does** 📖

**Additional comments** 🎤

### DIFF
--- a/components/ui/Modal/Modal.less
+++ b/components/ui/Modal/Modal.less
@@ -4,12 +4,12 @@
   justify-content: center;
   position: fixed;
   inset: 0;
-  backdrop-filter: @light-blur;
+  backdrop-filter: @light-blur brightness(80%);
 
   .modal {
     transform-origin: 50% 200%;
     position: relative;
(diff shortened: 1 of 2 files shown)


components/ui/Modal/Modal.less
```diff
@@ -24,6 +25,8 @@
 
 .modal-enter,
 .modal-leave-to {
+  backdrop-filter: blur(0px);
```
jasonwoodland, line 28: Had to add this default value so css would animate the backdrop filter correctly

components/ui/Modal/Modal.less
```diff
@@ -4,11 +4,12 @@
   justify-content: center;
   position: fixed;
   inset: 0;
-  backdrop-filter: @light-blur;
+  backdrop-filter: @light-blur brightness(80%);
```
jasonwoodland, line 7: Added this brightness filter to differentiate the modal from the background a bit more

jasonwoodland (approved):

PR #4187, opened 2022-08-10
fix(mute-toggle): cleared previous sounds played

ThomBos:
**What this PR does** 📖
Clears previous sounds played when mute/unmute button is pressed

**Additional comments** 🎤

### DIFF
--- a/libraries/SoundManager/SoundManager.test.ts
+++ b/libraries/SoundManager/SoundManager.test.ts
@@ -60,7 +60,7 @@ describe('Manage sounds', () => {
 
   test('sound stops', () => {
     const spy = jest.spyOn(inst.sounds[Sounds.CALL], 'stop')
-    const result: any = inst.stopSound(Sounds.CALL)
+    const result: any = inst.stopSounds([Sounds.CALL])
 
     expect(spy).toHaveBeenCalled()
     expect(result).toMatchSnapshot()
--- a/libraries/SoundManager/SoundManager.test.ts
+++ b/libraries/SoundManager/SoundManager.test.ts
@@ -59,7 +59,6 @@ describe('Manage sounds', () => {
   })
 
   test('sound stops', () => {
-    const spy = jest.spyOn(inst.sounds[Sounds.CALL], 'stop')
     const result: any = inst.stopSounds([Sounds.CALL])
 
     expect(result).toMatchSnapshot()
--- a/libraries/SoundManager/SoundManager.test.ts
+++ b/libraries/SoundManager/SoundManager.test.ts
@@ -62,7 +62,6 @@ describe('Manage sounds', () => {
     const spy = jest.spyOn(inst.sounds[Sounds.CALL], 'stop')
     const result: any = inst.stopSounds([Sounds.CALL])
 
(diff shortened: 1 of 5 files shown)


libraries/SoundManager/SoundManager.test.ts
```diff
@@ -60,7 +60,7 @@ describe('Manage sounds', () => {
 
   test('sound stops', () => {
     const spy = jest.spyOn(inst.sounds[Sounds.CALL], 'stop')
-    const result: any = inst.stopSound(Sounds.CALL)
+    const result: any = inst.stopSounds([Sounds.CALL])
```
stavares843, line 63: thanks for updating 🔨

ThomBos: no prob!  👍

libraries/SoundManager/SoundManager.ts
```diff
@@ -58,20 +58,29 @@ export default class SoundManager {
...
+   * Stops specified sounds
+   * @name stopSounds
+   * @argument soundList Array of the sound file to stop.
    * @returns null
    */
-  stopSound(sound: Sounds) {
-    this.existsSound(sound)
+  stopSounds(soundList: Array<Sounds>) {
+    soundList.forEach((sound: Sounds) => {
+      this.existsSound(sound)
```
aewing, line 72: since existsSound throws, the first sound that isn't loaded will short circuit this forEach and prevent the following sounds from stopping. Should we forEach playingSounds()?

aewing (changes requested):

libraries/SoundManager/SoundManager.ts
ThomBos: it should throw only if the sound passed as payload doesn't exist, doesn't throw if the sound isn't playing

libraries/SoundManager/SoundManager.test.ts
ThomBos: had to remove it to... does it work like this? 🙈😅

libraries/SoundManager/SoundManager.ts
```diff
@@ -58,20 +58,31 @@ export default class SoundManager {
...
+   * @name stopSounds
+   * @argument soundList Array of the sound file to stop.
    * @returns null
    */
-  stopSound(sound: Sounds) {
-    this.existsSound(sound)
+  stopSounds(soundList: Array<Sounds>) {
+    soundList.forEach((sound: Sounds) => {
+      if (this.playingSounds().includes(sound)) {
+        this.existsSound(sound)
```
aewing, line 73: this line is redundant now that we are checking for inclusion in sounds

aewing (approved):

libraries/SoundManager/SoundManager.test.ts
stavares843: yes all green 🎉

libraries/SoundManager/SoundManager.ts
ThomBos: oh yeah, playingSounds already executes existsSound on its own

PR #4196, opened 2022-08-11
chore(reactivity): reactivity without vue set

maljuburi:
This is just a proof of concept for Vue reactivity.

it seems like we don't have to `Vue.set` in Iridium Managers,
we can simply update the manager state and then reference the state directly in a `computed` method.


josephmcg:
blocked while we figure out whats going on

aewing:
@maljuburi I think we should close this to avoid further debate.

PR #4291, opened 2022-08-16
feat(jest): 15% coverage for IdentityManager

drepram:
**What this PR does** 📖

Self-explanatory

AP-2128

**Additional comments** 🎤


libraries/Iridium/IdentityManager.test.ts
```diff
@@ -0,0 +1,60 @@
+import IdentityManager from './IdentityManager'
+const crypto = require('crypto')
+
+Object.defineProperty(global.self, 'crypto', {
+  value: {
+    getRandomValues: (arr) => crypto.randomBytes(arr.length),
+  },
+})
+
+describe('', () => {
```
stavares843, line 10: can we add a name on the describe instead of being empty?

stavares843 (changes requested):

libraries/Iridium/IdentityManager.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  test('generateEntropyMessage', () => {
+    const localParam = {
+      address: 'address',
+      secret: 'secret',
+    }
+    const result = IdentityManager.generateEntropyMessage(
+      localParam.address,
+      localParam.secret,
+    )
+    expect(result).toEqual(
```
aewing, line 20: I think we should snapshot this

libraries/Iridium/IdentityManager.test.ts
```diff
@@ -0,0 +1,60 @@
+import IdentityManager from './IdentityManager'
+const crypto = require('crypto')
+
+Object.defineProperty(global.self, 'crypto', {
```
aewing, line 4: Nice, can we maybe put this in a location where we can reuse it as a mock?

libraries/Iridium/IdentityManager.test.ts
stavares843: agree, avoids the spam below

libraries/Iridium/IdentityManager.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  value: {
+    getRandomValues: (arr) => crypto.randomBytes(arr.length),
+  },
+})
+
+describe('', () => {
+  test('generateEntropyMessage', () => {
```
aewing, line 11: I might have structured these tests with a describe for each function, i.e.:

describe someFn:
  it should throw an error if a string is passed in
  it should return a uint8array
  it should execute prescribed some computation

The verbosity here helps us to be semantic about what we are testing and why, and makes it clear from a quick review which features we are promising or supporting.

libraries/Iridium/IdentityManager.test.ts
aewing: Snapshots are useful when you get a lot of text or markup output because it's often easier to just gatekeep changes to it manually.

libraries/Iridium/IdentityManager.test.ts
stavares843: btw I also added additional jest matchers [here](https://github.com/Satellite-im/Core-PWA/pull/4300) that will be available when the PR gets merged that could be useful when creating tests

libraries/Iridium/IdentityManager.test.ts
stavares843: also true

stavares843:
Captura de ecrã 2022-08-18, às 12 05 31

drepram:
Closing this PR because something's wrong with GitHub Actions. Will open a new one!

PR #4292, opened 2022-08-16
feat(jest): 100% coverage for Iridium/chat/types

drepram:
**What this PR does** 📖

Self-explanatory

AP-2132

**Additional comments** 🎤


stavares843:
this needs to be added in iridium repo instead, please add this on that repo

libraries/Iridium/chat/types.test.ts
```diff
@@ -0,0 +1,12 @@
+import { ChatError } from './types'
+
+describe('', () => {
```
stavares843, line 3: can we add a name on the describe instead of being empty?

stavares843 (changes requested):

drepram:
After I committed my changes, 1 test failed `libraries/BlockchainClient/adapters/Phantom/PhantomAdapter.test.ts`.  

I do not see the relation between the file in this PR with that test that's returning an error, but I find the use in bringing it up so that we may resolve it quickly.

stavares843:
Captura de ecrã 2022-08-18, às 12 03 41

its all green on CI

is also all green locally, not sure why you got that error

stavares843 (approved):

PR #4293, opened 2022-08-16
feat(jest): 91% coverage for Iridium/utils

drepram:
**What this PR does** 📖

Self-explanatory

AP-2130

**Additional comments** 🎤


stavares843:
this needs to be added in iridium repo instead, please add this on that repo

libraries/Iridium/utils.test.ts
```diff
@@ -0,0 +1,19 @@
+import { setInObject } from './utils'
+
+describe('', () => {
```
stavares843, line 3: can we add a name on the describe instead of being empty?

stavares843 (changes requested):

libraries/Iridium/utils.test.ts
aewing: I second this, as a dev it's really helpful if the test text is descriptive enough to tell me what this function was testing specifically so I can go straight to the code that it impacts without needing to read the test file.

aewing: (thought I should read the test file anyway)

drepram:
After I committed my changes, 3 snapshots of 1 test suite (`libraries/WebRTC/encoders.test.ts`) failed. If this error is reproducible in other people's machine I see that we must open a ticket to resolve the error.

I do not see the relation between the file in this PR with that test that's returning an error, but I find the use in bringing it up so that we may resolve it quickly.

stavares843:
Captura de ecrã 2022-08-18, às 12 06 07

is all green on CI and locally not sure how you got that error

stavares843 (approved):

PR #4314, opened 2022-08-17
fix(files): fix rename reactivity and delete action

nathan-power:
**What this PR does** 📖
- Fixes reactivity for file name changes
- Fixes deletion from removing incorrect files

AP-2168

- These problems occurred due to the item data in "store/files/getters.ts" becoming cloned to avoid direct mutational changes from happening to the file state. As the file actions (including rename, delete, and like) were relying on directly mutating the file state they stopped working.
- @josephmcg Would you prefer the current fix and prevent direct mutation of the iridium file state to avoid any potential unintended side-effects? Or should we keep the original structure so the state can be directly mutated? Can understand the benefit of not cloning whenever a change is made, but considering the scope of the project it might be a good idea to avoid relying on mutating the state whenever possible.

**Additional comments** 🎤

### DIFF
--- a/store/files/getters.ts
+++ b/store/files/getters.ts
@@ -1,7 +1,6 @@
 import Vue from 'vue'
 import { GetterTree } from 'vuex'
 import fuzzysort from 'fuzzysort'
-import { cloneDeep } from 'lodash'
 import { FileRouteEnum, FileSortEnum } from '~/libraries/Enums/enums'
 import { IridiumDirectory, IridiumItem } from '~/libraries/Iridium/files/types'
 import { FilesState } from '~/store/files/types'
@@ -25,7 +24,7 @@ const getters: GetterTree<FilesState, RootState> & FilesGetters = {
         return files
       }
 
-      let items: IridiumItem[] = cloneDeep(files)
+      let items: IridiumItem[] = files
       const itemsFlat = flatDeep(items)
 
       if (state.search.searchAll) {
@@ -59,7 +58,7 @@ const getters: GetterTree<FilesState, RootState> & FilesGetters = {
       const key = state.sort.category
       const searchAllActive = state.search.searchAll && state.search.value
 
-      let items: IridiumItem[] = cloneDeep(filteredFiles)
+      let items: IridiumItem[] = filteredFiles
       const allItems = flatDeep(allFiles)
 
       // If "SearchAll" is inactive, we only want to sort the items that are in the current directory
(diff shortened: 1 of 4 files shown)


josephmcg:
Cant fully look yet, but id say avoid cloning at all if we can. Why was that introduced?

store/files/getters.ts
```diff
@@ -24,7 +25,7 @@ const getters: GetterTree<FilesState, RootState> & FilesGetters = {
         return files
       }
 
-      let items: IridiumItem[] = structuredClone(files)
```
tooshel, line 27: How is the lodash cloneDeep different from the built in structuredClone?

phillsatellite (approved):

store/files/getters.ts
nathan-power: It shouldn't be, but the objects from the second layer onwards in the structured clones were still keeping their references for some reason. Switching to cloneDeep seemed to fix that, but if possible I'd rather stick with using structuredClone for the performance.

nathan-power:
> Cant fully look yet, but id say avoid cloning at all if we can. Why was that introduced?

The file data eventually gets passed down to the grid/list components as props, because Vue's keeping track of them it's usually not a good idea to mutate them, which is what was happening. It's been rewritten so that we're not relying on mutations anymore so it should be safe to remove the clones now, it's just that we'd have to be aware that we'd potentially be mutating Vue state.

store/files/getters.ts
josephmcg: i thought clonedeep was faster

store/files/getters.ts
```diff
@@ -24,7 +24,7 @@ const getters: GetterTree<FilesState, RootState> & FilesGetters = {
...
+      let items: IridiumItem[] = files
```
josephmcg, line 27: since we're no longer cloning, this may not be needed

nathan-power: Will fix it up now

nathan-power: Done

store/files/getters.ts
```diff
@@ -73,7 +60,7 @@ const getters: GetterTree<FilesState, RootState> & FilesGetters = {
         }
         // If set to "Recent", get 15 most recently edited files (no directories)
       } else if (route === FileRouteEnum.RECENT) {
-        const recentItems = allItems
+        const recentItems = iridium.files.flat
           .filter((e) => !('children' in e))
           .sort((a, b) => b.modified - a.modified)
           .slice(1, 14)
```
josephmcg, line 66: Wouldn't this skip the most recent item? I think you'd want 0,14

nathan-power: Good catch. This was a part of the original code so didn't notice, but probably should have. I'll get it fixed up now 👍

josephmcg (approved):

PR #4319, opened 2022-08-17
chore(webrtc): basic webrtc reimplementation

molimauro:
**What this PR does** 📖
Reintroduce the basic audio/video calls functionalities.

AP-1985

* Using only one `webrtc` sub as suggested;
* atm keeping track of the online status in friends, to be moved in usermanager later;
* created a hook for using the webrtc properties, if you like it we should consider using a similar approach for other managers too;
* wait till iridium and the managers are loaded before showing the ui

**Additional comments** 🎤


WanderingHogan:
🚀

PR #4378, opened 2022-08-19
feat(jest): fix and improve skipped tests

drepram:
**What this PR does** 📖

This is just one part of this ticket. More will be done in the following days using the same JIRA ticket (if that is what was meant by combining them into 1 ticket).

AP-2169

**Additional comments** 🎤


stavares843 (approved):

WanderingHogan:
We are having some issues with our ipfs host - will the snapshot update itself when we move the glyphs internally? We are going to put them in the static folder until we get the marketplace actually set up

stavares843:
> will the snapshot update itself when we move the glyphs internally


yes 


we just need to use `yarn test -u` command

stavares843:
> (if that is what was meant by combining them into 1 ticket).


if its minor changes you can add everything into one PR, as long the PR doesn't get too big

PR #4382, opened 2022-08-19
fix(sounds): move sounds into static folder and remove ipfs load

Jekrimo:
**What this PR does** 📖
Move sound files into static

AP-2176


config.ts
```diff
@@ -28,15 +27,15 @@ export const Config = {
...
-    newMessage: `${soundsCID}/Notification.m4a`,
-    call: `${soundsCID}/Call.m4a`,
-    hangup: `${soundsCID}/Unused.m4a`,
-    mute: `${soundsCID}/Mute.m4a`,
-    unmute: `${soundsCID}/Unmute.m4a`,
-    deafen: `${soundsCID}/Deafen.m4a`,
-    undeafen: `${soundsCID}/Undeafen.m4a`,
-    upload: `${soundsCID}/Success.m4a`,
-    connected: `${soundsCID}/Success.m4a`,
+    newMessage: `static/Notification.m4a`,
```
WanderingHogan, line 30: here should be 'sounds' instead of 'static' -> look [here](https://nuxtjs.org/docs/directory-structure/static/) for more info:

libraries/SoundManager/SoundManager.ts
```diff
@@ -31,7 +31,7 @@ export default class SoundManager {
       string,
     ][]) {
       this.sounds[key] = new Howl({
-        src: [`${Config.ipfs.gateway}${value}`],
+        src: [`${Config.sounds}${value}`],
```
WanderingHogan, line 34: this should just be [value] - you are already looping over config.sounds, value is the individual item in the loop, eg 'sounds/Mute.mfa' or something

WanderingHogan (changes requested):
very close, some minor changes in the inline comments should help

WanderingHogan (approved):

PR #4394, opened 2022-08-22
chore(jest): add missing snapshot

stavares843:
**What this PR does** 📖

- Adds missing snapshot


WanderingHogan:
We don't host anything in pinata anymore

WanderingHogan (changes requested):
We don't host anything in Pinata and there wlll be an update to the glyphs to move them to static this afternoon

stavares843:
k will update the snap when that happens, thanks 😂 

closing this for now

PR #4397, opened 2022-08-22
fix(call-ui): fixed loader ui

ThomBos:
**What this PR does** 📖
fixes loader's ui 
AP-2173

**Additional comments** 🎤

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "@nuxtjs/toast": "^3.3.1",
     "@project-serum/anchor": "0.18.2",
     "@solana/wallet-adapter-wallets": "^0.16.8",
-    "@satellite-im/iridium": "git+https://git@github.com/Satellite-im/iridium.git#main",
     "@solana/web3.js": "1.43.1",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -34,7 +34,6 @@
     "@nuxtjs/toast": "^3.3.1",
     "@project-serum/anchor": "0.18.2",
     "@solana/wallet-adapter-wallets": "^0.16.8",
-    "@satellite-im/iridium": "git+https://git@github.com/Satellite-im/iridium.git#main",
```
josephmcg, line 37: Probably an accident :)

josephmcg (changes requested):

package.json
WanderingHogan: it doesn't really matter if it's in there or not if people are using the linkscript, @aewing mentioned making it an npm module that can be installed now that iridium is mostly stabilized soon, so it will matter then

josephmcg: oh cool, makes commits a lot easier

josephmcg (approved):

PR #4399, opened 2022-08-22
fix(comingsoon): move vif to wallet on chatbar

Jekrimo:
**What this PR does** 📖
stop v-if from blocking send of mobile, move v-if to wallet function

AP-2210


Jekrimo:
I think that was all the ticket was needing?

WanderingHogan (approved):

PR #4402, opened 2022-08-23
refactor(calls): improve mediaUser markup

josephmcg:
**What this PR does** 📖
- refactor tag component to use slots and new Text component
- remove extra container elements from mediaUser. display both video and screenshare at the same time if possible. Still needs some additional styling if 2 windows are shown for 1 user. I wanted to break this out separately since it may be fairly involved
- adjust call loader to not display user image, only loader
- switch position of incoming call buttons so accept is on the right. I believe this is the convention in most user interfaces
- add overflow scroll to media window

AP-2213

**Additional comments** 🎤

### DIFF
--- a/assets/styles/base.less
+++ b/assets/styles/base.less
@@ -1,5 +1,4 @@
 :root {
-  --app-height: 100vh;
   --flair-color: @satellite-color;
   --flair-color-secondary: @satellite-color-secondary;
   --flair-color-rgb: @satellite-color-rgb;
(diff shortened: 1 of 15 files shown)


molimauro:
I see the user background is always grey now, maybe we can discuss if make sense to select a color based on the user image color or the circle color

molimauro (approved):

josephmcg:
> I see the user background is always grey now, maybe we can discuss if make sense to select a color based on the user image color or the circle color

Yeah i took it from the figma. I like that idea though. I'll definitely check that out after getting the general layout in place

PR #4410, opened 2022-08-23
fix(reply): correct name of reply recipient

Jekrimo:
**What this PR does** 📖
Fix name display in reply bar

AP-2154

COMMENTS:
I have no idea why a _snapshots_/getters.test.ts.snap was included. Never touched it, let me know if we need to delete it or anything.

### DIFF
--- a/components/views/chat/chatbar/reply/Reply.vue
+++ b/components/views/chat/chatbar/reply/Reply.vue
@@ -3,7 +3,11 @@
 import Vue from 'vue'
 import { mapState } from 'vuex'
 import { XIcon } from 'satellite-lucide-icons'
-import { ConversationMessage } from '~/libraries/Iridium/chat/types'
+import {
+  Conversation,
+  ConversationMessage,
+} from '~/libraries/Iridium/chat/types'
+import iridium from '~/libraries/Iridium/IridiumManager'
 
 export default Vue.extend({
   components: {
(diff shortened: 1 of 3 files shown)


stavares843:
> I have no idea why a snapshots/getters.test.ts.snap was included. Never touched it, let me know if we need to delete it or anything.


no worries i deleted it 🔨 i also updated the test in order to avoid adding any more snapshots

components/views/chat/message/Message.vue
```diff
@@ -60,6 +60,16 @@ export default Vue.extend({
       return this.$route.params.id
     },
     author(): User {
+      // iridium.users.getUser can not return self, check if it's you first. On reload this would be missing,
```
josephmcg, line 63: the iridium users manager was recently updated to include your own user's details, so this workaround isn't necessary anymore. check `components/views/chat/message/Message.vue` for an example

components/views/chat/chatbar/reply/Reply.vue
```diff
@@ -4,13 +4,21 @@ import Vue from 'vue'
...
   },
   computed: {
     ...mapState(['chat']),
+    replyUserName(): string | undefined {
+      const conversationId = this.$route.params.id
+      if (!conversationId) {
+        return undefined
+      }
+      return iridium.chat.state.conversations[conversationId].name
+    },
```
josephmcg, line 21: we don't necessary want the conversation name there. We want the name of the message author. you should be able to pass message sender did into user manager.

josephmcg (changes requested):

josephmcg:
fixed here
https://github.com/Satellite-im/Core-PWA/pull/4452

PR #4421, opened 2022-08-23
feat(safer): validate files with Safer

maljuburi:
AP-2186


components/ui/Chat/Image/Image.html
```diff
@@ -13,13 +13,9 @@
       {{placeholderText}}
     </figcaption>
   </figure>
-  <ui-image-container
-    v-else
-    class="image-container"
-    :nsfw="image.nsfw && blockNsfw"
```
josephmcg, line 19: we need to check the blockNsfw user setting to determine whether they want to see nsfw content. This current change would always blur if its nsfw, regardless of user setting. Although, we could probably add it to the image-container component, rather than doing it at the parent level every time

aewing (approved):
If this doesn't break anything I would be in favor of merging it and following up with another UI ticket

josephmcg:
why switch thumbnail from cid to the blob itself?

maljuburi:
For the thumbnail, my thought was to _not_ store 2 files ( original image + thumbnail ) in Iridium . Instead I thought thumbnails should only be a manager/front-end utility that resizes the original image we store in Iridium imo. right?

maljuburi:
Maybe thumbnail shouldn't be in the attachment all together... I don't know! 
I think we should only send the original image to the users (after it gets Safer checked). 
And then, the users generate its thumbnail.

PR #4439, opened 2022-08-24
feat(calls): add fullscreen support

josephmcg:
**What this PR does** 📖
- uses fullscreen api
- if the user wants to enlarge call window but keep other parts of the app visible, they can use the dragbar

AP-2193

**Additional comments** 🎤


molimauro (approved):
Dope!

PR #4448, opened 2022-08-24
fix(call-ui): fixed call display on mobile devices

ThomBos:
**What this PR does** 📖
Fixes call ui for mobile devices

**Additional comments** 🎤

### DIFF
--- a/components/ui/Global/Global.vue
+++ b/components/ui/Global/Global.vue
@@ -33,7 +33,7 @@ export default Vue.extend({
       if (!this.$device.isMobile) {
         return this.isBackgroundCall
       }
-      return this.isBackgroundCall || (this.isActiveCall && this.ui.showSidebar)
+      return this.isBackgroundCall || (this.isActiveCall && this.ui.showSidebar) // this.ui.isMobileNavVisible
     },
   },
   watch: {
@@ -107,9 +107,6 @@ export default Vue.extend({
       const callId = this.webrtc.activeCall?.callId
       if (!callId) {
         return
-      }
-      if (this.$device.isMobile) {
-        this.$store.commit('ui/showSidebar', false)
       }
       const callingPath = `${
         this.$device.isMobile ? '/mobile/chat/' : '/chat/'
--- a/components/ui/Global/Global.vue
+++ b/components/ui/Global/Global.vue
@@ -108,8 +108,12 @@ export default Vue.extend({
       if (!callId) {
         return
       }
-
-      const callingPath = `/chat/${callId}`
+      if (this.$device.isMobile) {
+        this.$store.commit('ui/showSidebar', false)
+      }
+      const callingPath = `${
+        this.$device.isMobile ? '/mobile/chat/' : '/chat/'
+      }${callId}`
       if (this.$route.path !== callingPath) {
         this.$router.push(callingPath)
       }
--- a/components/ui/Pip/Pip.vue
+++ b/components/ui/Pip/Pip.vue
@@ -190,6 +190,7 @@ export default Vue.extend({
   cursor: -webkit-grab;
   overflow: hidden;
   &:extend(.no-select);
+  &:extend(.first-layer);
 }
 
 .dragging {
--- a/components/views/media/Media.less
+++ b/components/views/media/Media.less
@@ -5,7 +5,6 @@
   position: relative;
   border-radius: @corner-rounding;
   min-height: 100px;
-  &:extend(.first-layer);
 
   .media {
     position: relative;
--- a/components/views/media/Media.less
+++ b/components/views/media/Media.less
@@ -5,6 +5,7 @@
   position: relative;
   border-radius: @corner-rounding;
   min-height: 100px;
+  &:extend(.first-layer);
 
   .media {
     position: relative;
--- a/components/views/media/incomingCall/IncomingCall.less
+++ b/components/views/media/incomingCall/IncomingCall.less
@@ -25,19 +25,6 @@
     width: 100%;
     justify-content: space-evenly;
     margin-top: auto;
-
-    .button {
-      width: 3rem;
-      height: 2rem;
-    }
-  }
-}
-
-@media only screen and (max-width: @mobile-breakpoint) {
-  .incoming {
-    height: @full;
-    width: @full;
-    justify-content: center;
   }
 
   @media only screen and (max-width: @mobile-breakpoint) {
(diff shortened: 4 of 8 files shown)


components/ui/Global/Global.vue
```diff
@@ -96,8 +96,12 @@ export default Vue.extend({
...
         return
       }
-
-      const callingPath = `/chat/${callId}`
+      if (this.$device.isMobile) {
+        this.$store.commit('ui/showSidebar', false)
+      }
+      const callingPath = `${
+        this.$device.isMobile ? '/mobile/chat/' : '/chat/'
+      }${callId}`
```
josephmcg, line 104: there is no sidebar in mobile, no commit needed. You'd want to append call id to both mobile and desktop paths

pages/mobile/chat/_id.vue
```diff
@@ -63,11 +68,13 @@ export default Vue.extend({
               return
             }
             if (activeIndex === 0) {
+              this.$store.commit('ui/showSidebar', true)
```
josephmcg, line 71: dont need this

pages/mobile/chat/_id.vue
```diff
@@ -63,11 +68,13 @@ export default Vue.extend({
...
               this.swiper.allowSlidePrev = false
               this.swiper.allowSlideNext = true
               this.isMobileNavVisible = true
             }
             if (activeIndex === 1) {
+              this.$store.commit('ui/showSidebar', false)
```
josephmcg, line 77: dont need this

components/views/media/incomingCall/IncomingCall.less
```diff
@@ -24,6 +24,19 @@
     display: flex;
     width: 100%;
     justify-content: space-evenly;
-    margin-top: auto;
+    margin-top: @light-spacing;
+
+    .button {
+      width: 3rem;
```
josephmcg, line 30: I prefer the old margin. button size will be increased soon.

Also avoid using these variables

components/ui/Pip/Pip.vue
```diff
@@ -190,6 +190,7 @@ export default Vue.extend({
   cursor: -webkit-grab;
   overflow: hidden;
   &:extend(.no-select);
+  &:extend(.first-layer);
```
josephmcg, line 193: is this necessary? There's a general overuse of z-index in this project that I've been trying to cut back.

components/views/media/Media.less
```diff
@@ -5,6 +5,7 @@
   position: relative;
   border-radius: @corner-rounding;
   min-height: 100px;
+  &:extend(.first-layer);
```
josephmcg, line 8: same z index comment

josephmcg (changes requested):

components/ui/Global/Global.vue
ThomBos: how do we detect if the sidebar is open on mobile?

components/views/media/incomingCall/IncomingCall.less
ThomBos: with margin auto it breaks on mobile, do i have to set a fixed margin? (to avoid variables)

components/ui/Global/Global.vue
josephmcg: its all in the swiper config here
[pages/mobile/chat/_id.vue](https://github.com/Satellite-im/Core-PWA/blob/4da3ac92fe085108de156a252e8ac1bf5567b83b/pages/mobile/chat/_id.vue#L61)

components/views/media/incomingCall/IncomingCall.less
josephmcg: I merged this change. rebase and let me know if there's still issues

components/ui/Pip/Pip.vue
josephmcg: could you explain why it is necessary? I hid this from mobile in my recently merged PR as well, takes up too much room to be viable

components/views/media/incomingCall/IncomingCall.less
ThomBos: no no, this was the only issue, it just looked so bad 😂

josephmcg: thinking about it from the perspective of an actual mobile user, you want the buttons at the bottom.

actually increased button size here so I can stop parroting that line about 'it will look so much better when we...'
https://github.com/Satellite-im/Core-PWA/pull/4471

components/ui/Global/Global.vue
ThomBos: > its all in the swiper config here [pages/mobile/chat/_id.vue](https://github.com/Satellite-im/Core-PWA/blob/4da3ac92fe085108de156a252e8ac1bf5567b83b/pages/mobile/chat/_id.vue#L61)

can't find a way to use swiper's functions on external components, is there any way to do that?

josephmcg: You dont need to do anything. It will hide itself on slide 2 with this config. If you pass a route param, it will automatically go to slide 2

josephmcg:
dont resolve unless its actually fixed

josephmcg:
fixed here, closing
https://github.com/Satellite-im/Core-PWA/pull/4490

components/ui/Global/Global.vue
ThomBos: it doesn't if the route is the same and on mobile you don't change route while opening and closing the sidebar

pages/mobile/chat/_id.vue
ThomBos: need this as a workaround for the problem above

components/ui/Pip/Pip.vue
ThomBos: exactly, this is necessary on mobile since some ui goes upfront.
yeah it takes a quarter of a screen (don't think it might be helpful going with a smaller PiP), but it can be moved by touch (working on this rn)

josephmcg: @ThomBos 
You don't need z-index to make something appear on top of the other thing. HTML simply needs to be below to appear on top. Because the <global> component (therefore all the modals inside of it) are placed below the rest of the app content, they're automatically on top.

Read this article and let me know if you have any questions

https://www.codegram.com/blog/stop-ab-using-z-index/

josephmcg: regarding the mobile call indicator, I made a solution here that avoids a large draggable element https://github.com/Satellite-im/Core-PWA/pull/4492

PR #4461, opened 2022-08-25
fix(friends): send user to correct screen when "Add Friend" is pressed

nathan-power:
**What this PR does** 📖
- If user is on mobile and has no friends then default friends screen is the "Add Friend" screen.

AP-2237

**Additional comments** 🎤

### DIFF
--- a/components/interactables/KeybindInput/KeybindInput.vue
+++ b/components/interactables/KeybindInput/KeybindInput.vue
@@ -155,6 +155,7 @@ export default Vue.extend({
         return
       }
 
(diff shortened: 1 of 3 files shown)


josephmcg:
gonna switch to router instead

josephmcg:
if we use that same route names, then we can double it up for desktop too :)

nathan-power:
> if we use that same route names, then we can double it up for desktop too :) desktop currently using store, but should use router

Should we do that for the desktop route in this PR as well?

josephmcg:
nah separate is ok

PR #4474, opened 2022-08-25
fix(notifications): fix group chat notification messaging

Jekrimo:
**What this PR does** 📖
fix group chat notification messaging, set catch for group message type and add phrasing. 

AP-2218


molimauro (changes requested):
Can we just add a return after getUser just in case a user is not found? Other than that perfect!

molimauro (approved):

PR #4475, opened 2022-08-25
chore(unlock): notice on unlock page to recommend chrome

WanderingHogan:
**What this PR does** 📖
Adds a warning message on unlock page to recommend Chrome, but still allow user to log in/create account

Screen Shot 2022-08-25 at 1 41 45 PM

this is hidden on mobile, and removes the realms since we dont have realms rightnow. Firefox on desktop does not trigger the mobile flag when resized like chromium browsers do

**Additional comments** 🎤

### DIFF
--- a/locales/en-US.js
+++ b/locales/en-US.js
@@ -239,7 +239,7 @@ export default {
       delete_account_label: 'Not you? Create or import an account',
       generate_random_user: 'Generate a Random User (DEV)',
       browser_warning:
-        'Hey there, welcome to our early access! While one of our goals is to make Satellite work across all browsers, we have focused the early access on Chromium-based browsers. You are welcome to try Satellite on any browser you like, and we will focus on more broad support soon, but you may run into more UI and functionality issues.',
+        'Satellite is optimized for Chromium browsers. You may experience UI and functional issues. We will optimize for more browsers in the future.',
     },
     loading: {
       loading: 'Linking Satellites...',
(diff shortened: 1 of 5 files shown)


locales/en-US.js
```diff
@@ -238,6 +238,8 @@ export default {
       store_pin: 'Store Pin? (Less Secure)',
       delete_account_label: 'Not you? Create or import an account',
       generate_random_user: 'Generate a Random User (DEV)',
+      browser_warning:
+        'Hey there, welcome to our early access! While one of our goals is to make Satellite work across all browsers, we have focused the early access on Chromium based browsers. You are welcome to try Satellite on any browser you like, and we will focus on more broad support soon, but you may run in to more UI and functionality issues.',
```
stavares843, line 242: > Hey there, welcome to our early access! While one of our goals is to make Satellite work across all browsers, we have focused the early access on Chromium-based browsers. You are welcome to try Satellite on any browser you like, and we will focus on more broad support soon, but you may run into more UI and functionality issues.


suggestion

stavares843:
Captura de ecrã 2022-08-25, às 20 54 19


using safari, this gets cut off, maybe we can divide in multiple lines instead

locales/en-US.js
WanderingHogan: cool, can you make that change?

stavares843:
also, i only get the warning on safari

doesn't appear on opera or edge

WanderingHogan:
good find on safari

opera, brave, edge are all chromium based so we dont want the warning there

stavares843:
thanks for clarifying 🔨

locales/en-US.js
stavares843: will update 🔨

stavares843: done 

https://github.com/Satellite-im/Core-PWA/pull/4475/commits/558b121c880559a169a956412f67b5df3523d9df

stavares843 (approved):

PR #4495, opened 2022-08-26
feat(jest): 100% coverage for utilities/chat

drepram:
**What this PR does** 📖

Increase coverage for two untested cases.

AP-2247

**Additional comments** 🎤

### DIFF
--- a/utilities/chat.test.ts
+++ b/utilities/chat.test.ts
@@ -211,4 +211,38 @@ describe('chat.conversationMessageIsNotice', () => {
     })
     expect(result).toMatchSnapshot()
   })
+
+  it('returns true for member_join type', () => {
+    const result: any = chat.conversationMessageIsNotice({
+      id: '',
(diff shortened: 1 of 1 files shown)


stavares843 (approved):

josephmcg (approved):
going to approve because it doesn't hurt. but an object is always truthy, this doesn't really test anything

stavares843:
@josephmcg let us know what we should address there and we can address your comments on a new PR considering this one was already merged 🔨

josephmcg:
I'd say check object values specifically since objects are always truthy

image

josephmcg:
oh im sorry, I was totally reading it incorrectly. this works fine

josephmcg:
would avoid casting as 'any' as a rule of thumb, especially if we know the type (boolean in this case)

stavares843:
😂 

thanks @josephmcg 🔨

PR #4520, opened 2022-08-29
chore(actions): issue translator

stavares843:
**What this PR does** 📖

- If the community members add an issue that is not in English, this action will translate the issue into English

more info [here](https://github.com/marketplace/actions/issues-translator)

### DIFF
--- a/.github/workflows/issue-translator.yml
+++ b/.github/workflows/issue-translator.yml
@@ -1,0 +1,14 @@
+name: issue-translator
+on: 
+  issue_comment: 
+    types: [created]
+  issues: 
+    types: [opened]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tomsun28/issues-translate-action@v2.6
+        with:
+          IS_MODIFY_TITLE: false


tooshel:
How long do these checks usually take? Is it me or did github not run this action? If so that sucks . . . . you can't add new actions and see what they do before you merge them??

stavares843:
> How long do these checks usually take? Is it me or did github not run this action? If so that sucks . . . . you can't add new actions and see what they do before you merge them??


it didn't run on the PR, because this is for issues only

tooshel:
OHHHH, I didn't know you could add actions to issues. Interesting. So the idea is that we allow issues from all over and they get translated? Is that a problem we have now? 


> > How long do these checks usually take? Is it me or did github not run this action? If so that sucks . . . . you can't add new actions and see what they do before you merge them??
> 
> it didn't run on the PR, because this is for issues only

stavares843:
> OHHHH, I didn't know you could add actions to issues. Interesting. So the idea is that we allow issues from all over and they get translated? Is that a problem we have now?


is not a problem because no one from outside our team added any issues, but in case anyone from the community adds a issue on any other english besides english such as brazilian or indonesian or others, this would translate the issue

PR #4535, opened 2022-08-30
fix(add-friends): removed infinite loading

### LINKED ISSUE #4387: AP-2209: add friends on mobile - searches forever

ThomBos:
**What this PR does** 📖
Fixes infinite loading on mobile devices
Resolve #4387 

I think we might need to wait for a friends requests fix
**Additional comments** 🎤


stavares843:
updated snap to fix Jest tests 🔨

WanderingHogan:
@stavares843 please don't overwrite peoples code - i cant tell what was there before you overwrote this branch

stavares843:
@WanderingHogan is the same code, except the extra snap 🔨

stavares843:
you can see the original content here @WanderingHogan 

https://github.com/Satellite-im/Core-PWA/pull/4535/commits/78240fd04a20091e8981c57db9463330572e1cee

libraries/Iridium/users/UsersManager.ts
```diff
@@ -241,11 +241,14 @@ export default class UsersManager extends Emitter<IridiumUserPubsub> {
         resolve([])
       }, 10000)
 
-      this.once(`searchResults/${id}`, (results: User[]) => {
-        clearTimeout(timeout)
-        logger.info(this.loggerTag, 'peer search results', results)
-        resolve(results)
-      })
+      this.iridium.connector?.p2p.on(
```
aewing, line 244: This should be once.

aewing: This works for me, so what is it you're trying to solve? Once is the correct solution here, otherwise we will have an emitter handler registered for events that will never be emitted again every time we search. If `once` isn't working, we should look into why instead of changing it.

ThomBos: it doesn't work, like the function doesn't exists

ThomBos: Schermata 2022-08-31 alle 09 49 20

aewing: looks like you need to update iridium.

aewing: you may need to delete `sync-node/data/sync.sqlite` also

ThomBos: i might be missing something... i don't have sync.sqlite in the path you suggested and also after the update i have the same result

WanderingHogan:
this got fixed with iridium updates, thank you for working to address it!

PR #4548, opened 2022-08-31
fix(groups): group icon render issues

josephmcg:
**What this PR does** 📖
- fix group icon render
- improve readability & scalability for toolbar render logic
- create some conversation hooks, switch toolbar and mobileToolbar to script setup() syntax
- improve flexibility of existing call method, allowing params. this is helpful in cases of calling from the sidebar, since you don't always just want to scrape conversation id from the route. These hooks could probably use some additional refactoring in that regard, but that can be handled separately 

**Additional comments** 🎤

### DIFF
--- a/components/compositions/conversations.ts
+++ b/components/compositions/conversations.ts
@@ -1,6 +1,7 @@
 import { computed, ComputedRef } from 'vue'
 import { Conversation } from '~/libraries/Iridium/chat/types'
 import iridium from '~/libraries/Iridium/IridiumManager'
+import { User } from '~/libraries/Iridium/users/types'
 import { TrackKind } from '~/libraries/WebRTC/types'
 
 export function conversationHooks() {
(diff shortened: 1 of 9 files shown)


aewing (approved):

josephmcg:
working on composition api approach to reduce reused logic, moving to draft for now

components/compositions/conversations.ts
```diff
@@ -0,0 +1,37 @@
+import { computed, ComputedRef } from 'vue'
+import { Conversation } from '~/libraries/Iridium/chat/types'
+import iridium from '~/libraries/Iridium/IridiumManager'
+
+export function conversationHooks() {
```
aewing, line 5: Why

josephmcg: so any updates only have to happen in one file, rather than two

aewing: what do we gain vs putting the equivalent in chat & webrtc managers instead? (`setActiveConversation`, etc...)

josephmcg: that's a valid approach, I suppose I enjoy the separation of concerns.

The managers seem to be primarily concerned with iridium interaction and data manipulation. While these hooks are more about organizing data in a way that components can easily use it. I believe Vue has some special caching that goes on with computed values to improve performance

aewing: the managers are concerned with the data and methods of protocol functions. that includes participants, active conversation, whether it's a group, etc...

josephmcg: I'd like to keep the managers encapsulated if possible and avoid calling each other inside manager code. These hooks are resolving an issue of repeated computed properties across multiple components.

PR #4552, opened 2022-08-31
fix(webrtc): handle iridium publish error which was stopping settings…

nathan-power:
… manager init

**What this PR does** 📖
- Settings was not initiating as the WebRTC initiation before it was throwing an unhandled error. This PR handles the error so that the settings stored in iridium can be set on init

**Additional comments** 🎤

### DIFF
--- a/libraries/Iridium/webrtc/WebRTCManager.ts
+++ b/libraries/Iridium/webrtc/WebRTCManager.ts
@@ -108,7 +108,7 @@ export default class WebRTCManager extends Emitter {
       friends,
     })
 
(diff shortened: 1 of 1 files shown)


aewing (changes requested):
Can we instead check if profile.name && p2p.ready on attempted announce (and also catch & log)

nathan-power:
> Can we instead check if profile.name && p2p.ready on attempted announce (and also catch & log)
No worries, will make the changes now

aewing (approved):

PR #4560, opened 2022-08-31
feat(jest): 42% coverage for BlobManip

drepram:
**What this PR does** 📖

Self-explanatory. Commented out code at the last part of the code is due to possibly JSDOM being weird and not functioning well with FileStreams

AP-2257

**Additional comments** 🎤

### DIFF
--- a/utilities/BlobManip.test.ts
+++ b/utilities/BlobManip.test.ts
@@ -1,4 +1,4 @@
-import { blobToBase64 } from '~/utilities/BlobManip'
+import { blobToBase64, blobToStream } from '~/utilities/BlobManip'
 
 describe('convert blob to base 64 successfully', () => {
   test('file is jpeg', async () => {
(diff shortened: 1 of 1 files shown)


WanderingHogan (changes requested):
If blob doesn't exist in jsdom, is there a polyfill we can use, or can we mock it? As it stands this will always just fail out and never run this test

stavares843:
in case if we can't do what hogan mentioned above, please just ditch the commented out code

utilities/BlobManip.test.ts
```diff
@@ -49,3 +49,34 @@ describe('convert blob to base 64 failing', () => {
...
+    //   })
+    // })
+    // afterAll(() => {
+    //   Object.defineProperty(global, 'ReadableStream', {
+    //     writable: true,
+    //     value: original,
+    //   })
+    // })
+    try {
+      const jpegBlob = await new Blob(['testing'], { type: 'image/jpeg' })
```
josephmcg, line 76: this would be a text file right? not really a jpeg

drepram: ah yes, correct. please check again

stavares843:
@drepram 


in case if we can't do what hogan mentioned above, please just ditch the commented-out code

WanderingHogan:
There are a couple things here - 

We shouldn't worry about commented out code when it's stuff that's in flight and might help figure things out later. We should remove dead code, but I don't think this is the case here.

As-is, i don't think we should check this in, because it will never actually get to the test unless we can mock it (which i couldn't figure out how to do in the time I had to look at this)

Screen Shot 2022-09-01 at 12 14 31 PM
Screen Shot 2022-09-01 at 12 14 35 PM

stavares843:
will close this then, if you figure out a solution we can reopen @drepram

PR #4606, opened 2022-09-02
feat(sidebar): add loaders until conversations are loaded in iridium

josephmcg:
**What this PR does** 📖
- adds a couple loaders to the sidebar list until conversations are loaded in iridium.
  - This introduces some layout shift for users with a small number of conversations. However, users will ultimately fill up that list with regular use, so I feel like more is better than less. 
- I removed the pulsing effect because it looked kinda weird on this background.
  - maybe this is just me, would like to hear other opinions. But i might like a plain loader without a pulse animation better. It feels like the animation may bring attention to the loader. imo we want loaders to be low key and hopefully allow active pieces of the ui to draw the eye instead
- move single file loader components out of folders so theyre easier to get at

Resolve #

**Additional comments** 🎤


josephmcg:
seems like we still show loading screen until iridium is ready. This can be ignored for now. Gonna close, but keep branch

PR #4609, opened 2022-09-02
feat(permissions): permisions functionality added, improved its ui

### LINKED ISSUE #4556: Centralized place in app to manage permissions - Functionality Ticket
When the users toggle on and off the permissions, show the web API/built in 'enable permissions' alert to enable, and remove the browser permission when they remove access.

If removing permissions doesn't work like that, update chat. We might want to make these text buttons instead of a toggle, and give the user instructions on how to disable a permission if we cant just do it through the web api's

ThomBos:
**What this PR does** 📖
Improves permissions' tab ui and makes it actually work like expected
Resolve #4556 

**Additional comments** 🎤

### DIFF
--- a/components/interactables/Permissions/Permissions.vue
+++ b/components/interactables/Permissions/Permissions.vue
@@ -1,0 +1,34 @@
+<template>
+  <div class="permissions-container">
+    <InteractablesButton
+      v-if="state === 'prompt'"
+      color="primary"
+      :text="$t('pages.privacy.permissions.ask_permission')"
+      @click="$emit('click', $event)"
+    />
+    <TypographyText v-if="state === 'granted'" color="success">
+      {{ $t('pages.privacy.permissions.granted') }}
+    </TypographyText>
+    <TypographyText v-if="state === 'denied'" color="error">
+      {{ $t('pages.privacy.permissions.denied') }}
+    </TypographyText>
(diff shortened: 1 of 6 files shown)


josephmcg (approved):
thanks Thomas. Drew noticed some bugs in firefox, and I noticed some strange behavior in incognito chrome. We can investigate those in separate tickets

components/interactables/Permissions/Permissions.vue
```diff
@@ -0,0 +1,34 @@
...
+      :text="$t('pages.privacy.permissions.ask_permission')"
+      @click="$emit('click', $event)"
+    />
+    <TypographyText v-if="state === 'granted'" color="success">
+      {{ $t('pages.privacy.permissions.granted') }}
+    </TypographyText>
+    <TypographyText v-if="state === 'denied'" color="error">
+      {{ $t('pages.privacy.permissions.denied') }}
+    </TypographyText>
+    <UiLoadersLoadingBar v-if="state === 'loading'" />
```
josephmcg, line 15: just one note if you happen to check this. I think else-if would be better. Since these conditions are mutually exclusive, Vue would stop checking once it finds a condition that evaluates true. I believe this would check every condition regardless. This is minor, but impacts both performance and readability

PR #4629, opened 2022-09-07
fix(settings): remove loading message while updating remote settings

josephmcg:
**What this PR does** 📖
- this was more relevant while we used textile because it was super slow. The remote update is relatively quick now, so this keeps things more predictable. When a switch was changed to Locked, it would lose keyboard focus

Resolve #

**Additional comments** 🎤


nathan-power (approved):
LGTM!

PR #4640, opened 2022-09-07
feat(sidebar): add context call option

josephmcg:
**What this PR does** 📖
- refactor some of the context menu markup
- add sidebar calling if other user is online
- remove send message from context menu since that's what left click does. it was redundant
- hide quick profile until we can built it out more effectively

Resolve #

- I'd like to switch sidebarListItem to composition API, but it would be fairly involved
- This work makes me realize I should switch text component to use color type 'danger' instead of error to fit the rest of the app

**Additional comments** 🎤


WanderingHogan:
lit

WanderingHogan (approved):

PR #4667, opened 2022-09-08
feat(jest): 100% coverages

drepram:
**What this PR does** 📖

100% coverage for store/accounts/mutations.test.ts
100% coverage for store/sounds/actions.test.ts

**Additional comments** 🎤

### DIFF
--- a/store/accounts/mutations.test.ts
+++ b/store/accounts/mutations.test.ts
@@ -1,5 +1,7 @@
 import { AccountsState, RegistrationStatus } from './types'
-import mutations from '~/store/accounts/mutations'
+import InitialAccountsState from './state'
+import mutations from './mutations'
+import { User } from 'libraries/Iridium/users/types'
 
 describe('init', () => {
   let inst: any
@@ -40,7 +42,7 @@ describe('init', () => {
   })
 
   it('should setPin', () => {
-    const localStateForUnitTest = { ...state }
+    const localStateForUnitTest = InitialAccountsState()
     inst.setPin(localStateForUnitTest, '123456')
 
     expect(localStateForUnitTest).toMatchObject({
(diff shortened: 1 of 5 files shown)


store/accounts/mutations.test.ts
```diff
@@ -40,7 +42,7 @@ describe('init', () => {
   })
 
   it('should setPin', () => {
-    const localStateForUnitTest = { ...state }
+    const localStateForUnitTest = InitialAccountsState()
```
WanderingHogan, line 45: would this be a good change to use beforeEach?

```
  let localStateForUnitTest

  beforeEach(() => {
    localStateForUnitTest = InitialAccountsState()
  })
```

store/sounds/actions.test.ts
```diff
@@ -89,6 +89,27 @@ describe('Manage sounds', () => {
     )
   })
 
+  test('sound stops but argument is empty array', () => {
+    const result: any = inst.stopSounds(
+      {
+        state: {
+          newMessage: true,
```
WanderingHogan, line 96: you could clean this up a lot if you made an object at the top somewhere like this:

```
       const state = {
          newMessage: true,
          hangup: true,
          call: true,
          mute: true,
          unmute: true,
          deafen: true,
          undeafen: true,
          upload: true,
          connected: true,
        },
```

and then in the object you pass in, if you want undeafen to be false, for example

```
{
  state: {...state, undeafen: false}
}
        ```

WanderingHogan (changes requested):

store/sounds/actions.test.ts
drepram: resolved!

store/accounts/mutations.test.ts
drepram: resolved!

stavares843:
🎉 

feel free to update the label for missing review for hogan to re check 🔨 and thanks for updating 🎉

WanderingHogan (approved):

PR #4734, opened 2022-09-12
fix(typing-indicator): restored indicator

ThomBos:
### What this PR does 📖
- Restores typing indicator on the sidebar

- Resolve #4734

### DIFF
--- a/components/ui/User/State/State.vue
+++ b/components/ui/User/State/State.vue
@@ -10,7 +10,7 @@
         y="0"
         width="40"
         height="40"
-        :mask="`url(#circle-mask)`"
+        :mask="`url(#${finalMask}-mask)`"
       >
         <UiCircle
           :type="imageSource ? 'image' : 'random'"
@@ -27,13 +27,7 @@
           height="18"
           :mask="`url(#mask-state-${status})`"
         />
-        <foreignObject
-          v-if="status === 'typing'"
-          x="3"
-          y="9"
-          width="25"
-          height="6"
-        >
+        <foreignObject v-if="isTyping" x="3" y="9" width="25" height="6">
           <div id="typing-loader-container">
             <div id="typing-loader" />
           </div>
@@ -53,6 +47,10 @@ export default Vue.extend({
     user: {
       type: Object as PropType<User>,
       required: true,
+    },
+    isTyping: {
+      type: Boolean,
+      default: false,
     },
     size: {
       type: Number,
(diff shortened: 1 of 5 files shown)


phillsatellite:
Only 2 small things I found that I will create tickets for in the QA Jira board just wanted to share findings here as well, iOS typing indicator was appearing wrong and I noticed it didnt appear for Groupchats either but other than those 2 small things it worked extremely well! 🔥 
IMG_6572
https://user-images.githubusercontent.com/93608357/189741484-fbb447cc-c4d7-48b0-b965-0b5b3e3b1596.mov

phillsatellite (approved):

molimauro:
Thanks @phillsatellite! The group thing is not a bug, it’s intended :) we lo look into the ios bug though!! Thanks ;) all good on android?

components/ui/User/State/State.vue
```diff
@@ -75,16 +79,29 @@ export default Vue.extend({
...
+      return iridium.users.ephemeral.status[this.user.did]
     },
-    isTyping(): boolean {
-      return this.conversationId
-        ? iridium.chat.ephemeral.typing[this.conversationId]?.includes(
-            this.user?.did,
-          )
-        : false
+    finalMask(): string {
+      if (this.isTyping) {
```
molimauro, line 91: This should be redundant

stavares843:
thanks for testing @phillsatellite 💥 thanks for clarifying @molimauro 🔨

components/views/navigation/sidebar/list/item/Item.vue
```diff
@@ -67,10 +68,9 @@ export default Vue.extend({
       if (!this.user) return false
 
       return (
-        !!this.conversation &&
-        (iridium.chat.ephemeral.typing[this.conversation.id] || []).includes(
-          this.user.did,
-        )
+        (this.conversation.id &&
+          this.ephemeral.typing[this.conversation.id][this.user.did]) ||
+        false
```
nathan-power, line 73: Hey Thomas!
Tried adding a new friend but was met with this error. It might be good to include a guard clause here for if the conversation doesn't exist yet for the user.
image
image

nathan-power (changes requested):
Hey Thomas 👋
It's looking good, just one change I noticed that might be important

components/views/navigation/sidebar/list/item/Item.vue
stavares843: i saw this error before on dev as well

josephmcg:
https://github.com/Satellite-im/Core-PWA/pull/4745
Thanks for your work on this Thomas. I took some inspiration from your approach and resolved it here.

I opted for a conversationId prop to determine typing status render because the sidebarListItem component doesn't need to be aware of typing status. It was only calculating it to pass as a prop.

In that case, I think it's better to handle everything in the child component. This will also work well with a new group member sidebar I have in progress here
https://github.com/Satellite-im/Core-PWA/pull/4747

PR #4755, opened 2022-09-13
feat(chat): edit messages

### LINKED ISSUE #4723: chat - implement message edit
update your own state and announce to peers

molimauro:
### What this PR does 📖
Implement message edit functionality.

- Resolve #4723

If you send a new message and edit it while the other user is offline, the latter will receive the original copy. (same problem of the reactions while offline)

### Additional comments 🎤
Noticing some problems with the enhancers, disabled here, will fix on a separate PR.

### DIFF
--- a/libraries/Iridium/chat/ChatManager.ts
+++ b/libraries/Iridium/chat/ChatManager.ts
@@ -22,6 +22,8 @@ import {
   ConversationMessagePayload,
   IridiumConversationEvent,
   MessageAttachment,
+  MessageEdit,
+  MessageEditPayload,
   MessageReaction,
   MessageReactionPayload,
 } from '~/libraries/Iridium/chat/types'
@@ -39,7 +41,7 @@ export type ConversationPubsubEvent = IridiumMessage<
   IridiumDecodedPayload<{
     message?: ConversationMessage
     cid?: string
-    type: 'chat/message' | 'chat/reaction'
+    type: 'chat/message' | 'chat/reaction' | 'chat/edit'
     reaction?: MessageReaction
   }>
 >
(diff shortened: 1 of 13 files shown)


libraries/Iridium/chat/ChatManager.ts
```diff
@@ -296,13 +297,22 @@ export default class ChatManager extends Emitter<ConversationMessage> {
         [fromDID]: reaction.reactions,
       }
 
-      this.state.conversations[conversationId] = {
-        ...this.state.conversations[conversationId],
-        message: {
-          ...this.state.conversations[conversationId].message,
-          [message.id]: message,
-        },
-      }
```
josephmcg, line 305: does removing this cause reactivity issues? I'm not sure how Vue would detect the new object property

WanderingHogan:
this is so sweet, i think we should hold off until we figure out the offline piece of it though since we are so close to EA, what do you all think?

libraries/Iridium/chat/ChatManager.ts
molimauro: It should not since the message obj is already there so it's already being tracked

molimauro:
> this is so sweet, i think we should hold off until we figure out the offline piece of it though since we are so close to EA, what do you all think?

No problem of course ;)

libraries/Iridium/chat/ChatManager.ts
josephmcg: I thought this was an alternative way of writing
```
Vue.set(this.state.conversations[conversationId].message, [message.id], message)
```

josephmcg: oh yeah I see. I thought this was concerning replies, not reacts. Should be alright then

molimauro: Is replies reactivity working? Yesterday noticed some strange behaviour

josephmcg: You're right, there are some issues on the sender side. I'll make a ticket and check it out tomorrow (unless you want to get it today)

molimauro: Looks good, im trying to dig into this offline problem. Go with Vue.set :P

phillsatellite:
Tested and showed up edited on both sides. Had the refresh bug the first time but after re adding friends and sending edited messages I could refresh and messages would still show up edited 🔥

WanderingHogan (approved):

WanderingHogan:
@stavares843 @phillsatellite can you check this and merge if okay, otherwise comment here with screenshots

please check across all browsers, mac/win

stavares843 (approved):

PR #4758, opened 2022-09-13
chore(webrtc): Refactor webrtc

iltumio:
### What this PR does 📖
- Adds a mechanism to automatically connect with friends over direct webrtc. 
- Provides a single interface to send messages to peers that support both communication channels (direct webrtc as preferred and pubsub as fallback)

### Additional comments 🎤
Still working on retry mechanism in case the first connection attempt doesn't work

### DIFF
--- a/libraries/WebRTC/bus/DirectWebRTCBus.ts
+++ b/libraries/WebRTC/bus/DirectWebRTCBus.ts
@@ -1,0 +1,79 @@
+import { IridiumDocument, IridiumMessage } from '@satellite-im/iridium'
+import Peer from 'simple-peer'
+import type { SignalData, Instance } from 'simple-peer'
+import { Bus, BusListeners, BusOptions } from '~/libraries/WebRTC/bus/bus'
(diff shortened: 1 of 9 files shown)


libraries/WebRTC/bus/DirectWebRTCBus.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.bindPeerListeners()
+  }
+
+  onMessage(message: IridiumMessage<IridiumDocument>) {
+    this.emit('message', { type: this.type, message })
+  }
+
+  sendMessage(
+    message: IridiumDocument,
+    opts: BusOptions<{ recipients: string[] }>,
```
molimauro, line 30: is this used?

iltumio: the whole DirectWebRTCBus is not in use yet. It may be useful in the future, but for now we can think of removing it or just leave it there for further development

aewing (approved):
LGTM, with a followup to see if we can signal directly after connection

Issue #4736, opened 2022-09-12
Make first message a broadcast


WanderingHogan:
Currently for messaging we store the message and send the receiving user the CID once we're done.
We think we can improve performance in chat by calculating the CID (using new CID), sending the CID and the message body immediately, and _**then**_ storing and pinning the content, unblocking the UI on both sides while IPFS propagates.

The receiving user should still store the CID in place of the message for future loads, but can make the content available immediately on the UI while online.

PR #4761, opened 2022-09-13
fix(message): send message to user before store

Jekrimo:
### What this PR does 📖
- Send message before storing

- Resolve #
#4736 

I think this is right?

### DIFF
--- a/libraries/Iridium/chat/ChatManager.ts
+++ b/libraries/Iridium/chat/ChatManager.ts
@@ -748,7 +748,7 @@ export default class ChatManager extends Emitter<ConversationMessage> {
       },
     }
 
-    await iridium.connector.store(partial, {
+    iridium.connector.store(partial, {
       syncPin: true,
       encrypt: conversation?.participants
         ? { recipients: conversation?.participants }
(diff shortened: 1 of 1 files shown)


libraries/Iridium/chat/ChatManager.ts
```diff
@@ -744,24 +747,20 @@ export default class ChatManager extends Emitter<ConversationMessage> {
         lastReadAt: Date.now(),
       },
     }
+
+    await iridium.connector.store(partial, {
```
aewing, line 751: I think we can just call store and not await it for the UI's sake

aewing: Maybe we don't need to await the publish either... we might want to add a .catch that marks the message as unsent locally though so we can offer the user to re-send

aewing (approved):
Feel free to merge but it would be cool to test not blocking by removing the awaits

libraries/Iridium/chat/ChatManager.ts
Jekrimo: Good call, fixeded

PR #4770, opened 2022-09-14
fix(typing): typing bug

josephmcg:
### What this PR does 📖
- webrtc changes exposed holes in the old setTyping logic. This should be much better

- Resolve #


nathan-power (approved):
LGTM!

PR #4773, opened 2022-09-14
chore(cleanup): pre-launch cleanup

WanderingHogan:
### What this PR does 📖
super minor pre ea cleanup:
Elaborated in some of the settings descriptions
Hide register username publicly
Hide display current activity
Add ability to copy friendly name to Accounts and Devices
Hide Network (testnet select screen) tab in settings

- Resolve #

### DIFF
--- a/components/views/settings/modal/Modal.html
+++ b/components/views/settings/modal/Modal.html
@@ -30,6 +30,9 @@
       <SettingsPagesStorage
         v-show="ui.settingsRoute === SettingsRoutes.STORAGE"
       />
+      <!-- <SettingsPagesNetwork
+        v-show="ui.settingsRoute === SettingsRoutes.NETWORK"
+      /> -->
       <SettingsPagesKeybinds
         v-show="ui.settingsRoute === SettingsRoutes.KEY_BINDS"
       />
--- a/components/views/settings/modal/Modal.html
+++ b/components/views/settings/modal/Modal.html
@@ -30,9 +30,6 @@
       <SettingsPagesStorage
         v-show="ui.settingsRoute === SettingsRoutes.STORAGE"
       />
-      <SettingsPagesNetwork
-        v-show="ui.settingsRoute === SettingsRoutes.NETWORK"
-      />
       <SettingsPagesKeybinds
         v-show="ui.settingsRoute === SettingsRoutes.KEY_BINDS"
       />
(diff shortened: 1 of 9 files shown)


components/views/settings/pages/accounts/index.vue
```diff
@@ -17,12 +17,19 @@ export default Vue.extend({
...
     return {
       showPhrase: false,
+      profile: iridium.profile.state,
     }
   },
   computed: {
     ...mapState({
       accounts: (state) => (state as RootState).accounts,
     }),
+    getId(): string | unknown {
```
josephmcg, line 27: i typically see undefined rather than unknown

pages/mobile/settings/index.vue
```diff
@@ -32,7 +32,6 @@
         />
         <SettingsPagesProfile v-if="settingsRoute === SettingsRoutes.PROFILE" />
         <SettingsPagesStorage v-if="settingsRoute === SettingsRoutes.STORAGE" />
-        <SettingsPagesNetwork v-if="settingsRoute === SettingsRoutes.NETWORK" />
```
josephmcg, line 35: same question

josephmcg (changes requested):

components/views/settings/pages/accounts/Accounts.html
```diff
@@ -10,11 +10,20 @@
...
+      <clipboard-icon size="1x" />
+    </InteractablesInput>
+  </SettingsUnit>
+  <SettingsUnit>
+    <TypographyLabel :text="$t('pages.settings.accounts.active_did')" />
+    <InteractablesInput
+      readonly
+      :text="accounts.active"
+      @submit="copyAddress(accounts.active)"
+    >
```
josephmcg, line 23: maybe we want to move away from vuex for accounts information too. should fetch from iridium instead

components/views/settings/modal/Modal.html
```diff
@@ -30,9 +30,6 @@
       <SettingsPagesStorage
         v-show="ui.settingsRoute === SettingsRoutes.STORAGE"
       />
-      <SettingsPagesNetwork
-        v-show="ui.settingsRoute === SettingsRoutes.NETWORK"
-      />
```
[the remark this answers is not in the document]
WanderingHogan: we aren't going to use the page, it only has a dropdown from the solana networks, and we are never going back to the way we had that before, but good idea to put it back like you suggest, we can re-use it for other things more easily

components/views/settings/pages/accounts/Accounts.html
WanderingHogan: sweet, thank you

josephmcg (approved):

PR #4786, opened 2022-09-14
fix(sidebar): group icon flex bug

josephmcg:
### What this PR does 📖
- group icon was getting shrunk because of recent sidebarListItem changes. This prevents the following visual bug
- I applied at the component level since it should be a universal rule. I think we already apply this independently in the two Toolbar components
image

- Resolve #


nathan-power (approved):
LGTM!

PR #4794, opened 2022-09-14
fix(chat): offline msgs order

### LINKED ISSUE #4733: Message reactions and edits
If i send a message while the other user is offline and i attach a reaction to it (or edit a msg), sometimes the latter wont appear on the other user.

molimauro:
### What this PR does 📖
- Fix the offline message communication order so that reactions and edits wont be lost or overwritten

- Resolve #4733 

### Additional comments 🎤
 After this we can merge #4755 too


libraries/Iridium/chat/ChatManager.ts
```diff
@@ -146,28 +145,44 @@ export default class ChatManager extends Emitter<ConversationMessage> {
...
-            'iridium/chatmanager',
-            'sync/fetch/message - emitting synced message',
-            message,
+
+    // read msg infos in order
+    const msgs = await Promise.all(
+      message.payload.body.messages
+        .sort((a, b) => a.createdAt - b.createdAt)
+        .map(async (message) => {
+          const stored = await iridium.connector?.dag.get(
```
aewing, line 154: I think we can use iridium.load and skip the manual decryption step

aewing (approved):
🛳

libraries/Iridium/chat/ChatManager.ts
molimauro: for some reasons the `await iridium.connector?.load(msg.cid)` fails:

image

molimauro: thats because on iridium it loads this:

image

we should have object.body instead, where is the error? on core when storing the message?

WanderingHogan:
I am seeing weird issues where the offline messages don't come in consistently, or they will show up, then on reload they dissapear. Do you think this should be looked at in a follow up pr? Or maybe it works with your edit message pr, i will check that out now

molimauro:
Those should not be related to these PRs, are you experiencing the same on dev?

phillsatellite:
@molimauro I also tested this one as well and wanted to leave a little note of my findings. I was able to add friends and send messages from both sides and still not receive them on either side 

https://user-images.githubusercontent.com/93608357/190226569-1bb16f07-222a-4020-a99d-f6fd99f3cf80.mov

molimauro:
@phillsatellite i tried and they worked on my side, i dont think those inconsistencies are related to this PR honestly

aewing:
I think the inconsistency should be addressed in my branch. if not I will follow up on this.

libraries/Iridium/chat/ChatManager.ts
aewing: ok, maybe something about the structure of the payload. I'll take a look in a followup.

aewing (approved):

PR #4824, opened 2022-09-15
feat(chat): unblock sending while file upload in progress

### LINKED ISSUE #4823: unblock sending while files are uploading
show loader and allow other messages to be sent

josephmcg:
### What this PR does 📖
- move Chatbar uploading local state variable to vuex store.
  - I did this to put the message loader in a different component
  - it also ensures the state is maintained even if chatbar is unmounted
- display loader if store array contains this conversationId
- still allow new messages to be sent even if a previous upload is still in progress

- Resolve #4823

### DIFF
--- a/components/views/chat/chatbar/Chatbar.vue
+++ b/components/views/chat/chatbar/Chatbar.vue
@@ -243,14 +243,17 @@ const Chatbar = Vue.extend({
      * @example v-on:click="sendMessage"
      */
     async sendMessage() {
+      if (this.text.length > this.$Config.chat.maxChars) {
+        return
+      }
+
       if (
         !this.files.length &&
-        (this.text.length > this.$Config.chat.maxChars ||
-          !this.text.trim().length ||
-          !this.isSubscribed)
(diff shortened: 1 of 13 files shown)


components/views/chat/chatbar/Chatbar.vue
```diff
@@ -238,11 +244,10 @@ const Chatbar = Vue.extend({
      */
     async sendMessage() {
       if (
-        (!this.files.length &&
-          (this.text.length > this.$Config.chat.maxChars ||
-            !this.text.trim().length ||
-            !this.isSubscribed)) ||
-        this.activeUpload
+        !this.files.length &&
```
molimauro, line 247: if i have a file all the next conditions are skipped (i can send a message with more that maxChars)

molimauro (changes requested):
only one minor thing

components/views/chat/chatbar/Chatbar.vue
josephmcg: very true, fixed

aewing (approved):

PR #4830, opened 2022-09-15
fix(remove-friend): fixed sidebars' context menu

### LINKED ISSUE #4827: Right click on sidebar remove friend crash
While you have the chat opened, if you remove a friend from the sidebar the app throws an error

ThomBos:
### What this PR does 📖
- fixes crashes when removing a friend from the sidebar 
- Resolve #4827

### DIFF
--- a/components/views/navigation/sidebar/list/item/Item.vue
+++ b/components/views/navigation/sidebar/list/item/Item.vue
@@ -188,7 +188,9 @@ export default Vue.extend({
       }
       this.isLoading = true
 
-      this.$router.push(`/friends`)
+      if (this.conversation.id === this.$route.params.id) {
+        this.$router.push(this.$device.isMobile ? `/mobile/chat` : `/friends`)
+      }
 
       await iridium.friends
         .friendRemove(this.user.did)
--- a/components/views/navigation/sidebar/list/item/Item.vue
+++ b/components/views/navigation/sidebar/list/item/Item.vue
@@ -188,12 +188,7 @@ export default Vue.extend({
       }
       this.isLoading = true
 
-      if (this.$device.isDesktop) {
-        this.$router.push(`/friends`)
-      }
-      if (this.$device.isMobile) {
-        this.$router.push(`/mobile/chat`)
-      }
+      this.$router.push(`/friends`)
 
       await iridium.friends
         .friendRemove(this.user.did)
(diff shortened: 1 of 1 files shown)


components/views/navigation/sidebar/list/item/Item.vue
```diff
@@ -187,6 +187,14 @@ export default Vue.extend({
         return
       }
       this.isLoading = true
+
+      if (this.$device.isDesktop) {
+        this.$router.push(`/friends`)
+      }
+      if (this.$device.isMobile) {
+        this.$router.push(`/mobile/chat`)
+      }
```
josephmcg, line 196: I think you'd be a lot better off confirming that `this.conversationId === $route.params.id or whatever`, then pushing with a ternary.

```
this.$router.push(
        this.$device.isMobile
          ? `/mobile/chat}`
          : `/friends}`,
      )
```

josephmcg (changes requested):

components/views/navigation/sidebar/list/item/Item.vue
ThomBos: right!

components/views/navigation/sidebar/list/item/Item.vue
```diff
@@ -187,6 +187,9 @@ export default Vue.extend({
         return
       }
       this.isLoading = true
+
+      this.$router.push(`/friends`)
+
```
josephmcg, line 192: I tried explaining this in my previous comment, but maybe I wasn't clear.

A router push is a jarring event, especially if the user doesn't expect it. We only want to push if absolutely necessary. The crash only happens when you remove the friend for the active conversation you are looking at. So you would only want to push if that condition is true.

You also can't use 'friends' because mobile has a different friends page which contains all the swiper logic

This should be all you need

```
if (this.conversationId === this.$route.params.id) {
        this.$router.push(this.$device.isMobile ? '/mobile/chat' : '/friends')
      }
```

ThomBos: yeah, i didn't read the comment before!
once i answered i went updating it, but one thing... are we going to add an "on long press" event for mobile or we can clear that check for the route? (since mobile users can't open that context menu)

josephmcg: all good, thanks for updating. 

The same methods are passed to the action sheet/context mobile. It's a bit of a rabbit hole, but we're wrapping the whole component with this `UiContextMenu`. check it out if you're curious

josephmcg (approved):

PR #4832, opened 2022-09-15
fix(settings-safari): fixed crash

ThomBos:
### What this PR does 📖
- fixes the crash occured when the settings modal was opened on safari

- Resolve #

### DIFF
--- a/locales/en-US.js
+++ b/locales/en-US.js
@@ -288,6 +288,7 @@ export default {
       title: 'Privacy & Permissions',
       subtitle:
         'Choose which features to enable to best suit your privacy preferences.',
+      safari: 'We are currently not able to check permissions status on Safari',
       notifications: {
         title: 'Notifications',
         subtitle:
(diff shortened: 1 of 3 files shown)


locales/en-US.js
```diff
@@ -288,6 +288,8 @@ export default {
       title: 'Privacy & Permissions',
       subtitle:
         'Choose which features to enable to best suit your privacy preferences.',
+      update:
+        'We were not able to check permissions status, update your browser',
```
aewing, line 292: What permissions? this seems a bit cryptic

aewing (approved):

locales/en-US.js
stavares843: do you have a suggestion for this? 😂

aewing: well we specifically look for notifications, microphone, and camera (we might also want clipboard-write) -- so something specific that tells the lay person that we're looking for permissions to access their devices for the call and not their bank account or OS.

aewing: also is an outdated browser the thing that would stop them from checking permissions? or would it be operating system settings?

molimauro: yes on Safari works from version 16

PR #4835, opened 2022-09-15
fix(chat): typing indicator

### LINKED ISSUE #4834: typing indicator
I think there are some problems with the typing indicator:

image

Dont know how but this happened on 1to1 chat

molimauro:
### What this PR does 📖
- Fix typing indicator problem

- Resolve #4834

### DIFF
--- a/libraries/Iridium/chat/ChatManager.ts
+++ b/libraries/Iridium/chat/ChatManager.ts
@@ -770,13 +770,17 @@ export default class ChatManager extends Emitter<ConversationMessage> {
     const typingList = this.ephemeral.typing[conversationId]
 
     if (typing) {
-      // if reactive array exists, push. otherwise create
-      typingList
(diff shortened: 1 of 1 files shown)


molimauro:
> Do we need an index check on the splice below? Only splice if >= 0. I think splice w/ -1 as index removes elements from the end

done ✅

stavares843 (commented):
image



phil and I were testing and still happens seems more reproducible with pipes

stavares843:
not against merging 😂

molimauro:
Where is the comma between the names 😅

stavares843:
there is none 😂 

i think that one is just pipe specific, we also had this bug even before iridium so lol

stavares843:
not saying is from the PR, just saying we also have that one 😂

aewing (approved):

PR #4863, opened 2022-09-16
feat(personalFiles): upload files using iridium

### LINKED ISSUE #4854: switch personal files to use iridium encryption rather than plugging directly into ipfs
copy chatManager code

josephmcg:
### What this PR does 📖
- title

- Resolve #4854


phillsatellite:
Only thing I noticed was it took a little longer (Im assuming cause they're coming through encrypted now) but also noticed the percentage is gone when uploading. Other than that worked great!

josephmcg:
yeah there's no progress callback built in (yet), so ui suffers a bit. definitely will be slower too. Also limited in size until we resolve other issues

josephmcg:
disabling uploads until we can properly hook it up to safer.

WanderingHogan (approved):

Issue #4852, opened 2022-09-16
Group chat - invisible member on group list + undefined push + internal missing text on chat message


stavares843:
STR:
- If Phil is added as my friend and he adds a group with me Phil and Hogan and Hogan is not added as my friend will appear like this

<img width="1090" alt="Captura de ecrã 2022-09-16, às 01 55 43" src="https://user-images.githubusercontent.com/29093946/190535410-b79e6c4a-df01-4107-ab84-4ac51a63b4c9.png">

- the member that is not added on my friends list appears as invisible

then

the push’s will appear as undefined

![image](https://user-images.githubusercontent.com/29093946/190535521-95316c90-c79e-4b01-810a-a45be855313b.png)

and the messages will appear as internal missing text

![image](https://user-images.githubusercontent.com/29093946/190535541-5aeee4b1-05e6-4614-a054-25433bb9e2c7.png)

josephmcg:
i can't reproduce this locally, but we can at least make it fail gracefully. like add an 'unknown' instead of no user information

molimauro:
This is definitely because the searchPeer fails or times out when creating a group, not a frontend bug i guess unfortunately

PR #4901, opened 2022-09-16
fix(chat): send message after store&publish

molimauro:
### What this PR does 📖
- Make sure not to append the message in chat until it has been stored and published in iridium. The loader in chat is missing though

- Resolve #

### DIFF
--- a/libraries/Iridium/chat/ChatManager.ts
+++ b/libraries/Iridium/chat/ChatManager.ts
@@ -730,19 +730,29 @@ export default class ChatManager extends Emitter<ConversationMessage> {
       ...partial,
       id: tempCid.toString() as string,
     }
-    await iridium.connector.publish(
-      `/chat/conversations/${conversationId}`,
-      {
-        type: 'chat/message',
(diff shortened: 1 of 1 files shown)


WanderingHogan (approved):

WanderingHogan:
good idea

stavares843:
not saying this is from the PR just to be informative (even though prod has/had a similar bug)

this PR

- phil added me to a group with 4 users

image

- on my side was only 2 users and no messages

image


we did the same on core and i was able to get the messages from the other users + they were on the group list as well


note
- before on prod there was this [bug](https://github.com/Satellite-im/Core-PWA/issues/4852) similar, not the same even though i can no longer reproduce that atm

WanderingHogan:
i've seen that other bug on prod before and think it's unrelated, this should take care of the issue where you can send things and see it on one side, reload, and then not see it on the other because iridium didn't receive it yet

stavares843:
fine by me i also saw a similar one on prod that i added above but adding this just for info 😂

PR #4906, opened 2022-09-16
fix(thisisfine): add message to user that service is down

Jekrimo:
### What this PR does 📖
- Add message to loading screen


WanderingHogan:
oh thank god we don't need this, thank you for prepping it in case we did, though

PR #4920, opened 2022-09-19
feat(settings): settings header component

### LINKED ISSUE #4919: add nav back to mobile settings pages

jasonwoodland:
### What this PR does 📖
- Add SettingsHeader component
- Fix mobile styling on settings page
- Add settings page header

- Resolve #4919

### DIFF
--- a/components/views/settings/Sidebar.vue
+++ b/components/views/settings/Sidebar.vue
@@ -107,7 +107,6 @@ export default Vue.extend({
 .menu {
   overflow-y: auto;
   &:extend(.no-select);
-  padding: 8px;
   background: @semitransparent-dark-gradient;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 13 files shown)


molimauro:
Looks good, only one minor thing: If i go to Settings > Personalize and then click the back arrow on top i get this:

image

jasonwoodland:
@molimauro Thanks! did any of the other pages do it? I'll add a null check now

jasonwoodland:
I would have liked to use the `useSwiper` hook, but we're using the old version of swiper, so I had to use the `$el.closest()` method

molimauro:
> @molimauro Thanks! did any of the other pages do it? I'll add a null check now

Yes pretty much every page seems to be affected

jasonwoodland:
Chrome?

molimauro:
Yes!

molimauro:
> I would have liked to use the `useSwiper` hook, but we're using the old version of swiper, so I had to use the `$el.closest()` method

cant we update that? :)

jasonwoodland:
we definitely should, just didn't want to exceed the scope of this pr

jasonwoodland:
I can't reproduce the problem you're having, I'm getting @nathan-power to test also

molimauro:
Yes i totally agree! Let's open a ticket then :)

jasonwoodland:
Opened https://github.com/Satellite-im/Core-PWA/issues/4921

molimauro:
> I can't reproduce the problem you're having, I'm getting @nathan-power to test also

Maybe this can help!

https://user-images.githubusercontent.com/50131258/190969321-85f4b0db-6180-4ac4-a902-ea9ae0ec597b.mov

jasonwoodland:
@molimauro Could you please try that I just pushed, I'm not sure why else it won't be able to find the swiper ancestor in the dom tree.

nathan-power:
Just an update, it seems it's working on my end

molimauro:
Yes it's working now! I noticed that before was working locally but not on fleek preview 🤔 Seems fixed now though, thanks!

jasonwoodland:
Nice, thanks for checking it out :)

molimauro (approved):

PR #4924, opened 2022-09-19
fix(chat): scroll behaviour

### LINKED ISSUE #4917: Need to fix scroll to bottom on new message
If a user is scrolled all the way down, incoming messages need to scroll the chat window.

If the user is scrolled up, then we should not auto scroll on incoming messages

molimauro:
### What this PR does 📖
- Switch MutationObserver in favour of ResizeObserver, simpler in this case;
- Watch new messages and if im the sender, bring me to chat bottom when sending a message. 

- Resolve #4917 

Test by adding emojis etc

### Additional comments 🎤
Still a problem when the chatbar height increase, should we add another resizeObserver for the conversation container?


phillsatellite (approved):

WanderingHogan:
Screen Shot 2022-09-19 at 4 14 24 PM

🚀 🚀 🚀

stavares843 (commented):
https://user-images.githubusercontent.com/29093946/191130106-57ec29d3-c82d-4942-9f83-57534cb3a820.mov

sometimes when you are scrolled up and you send something doesn't scroll the whole chat but not against merging


Chrome, macOS

WanderingHogan:
@stavares843 it looks like it will scroll but bounce?

I am not sure we want to force autoscroll if the user has manually scrolled up, just if there is a message and they are scrolled down or send a message already?

stavares843:
>  it looks like it will scroll but bounce?


yes

sometimes bounces, sometimes doesn't bounce

stavares843:
the idea we want here is

if you scroll up and i send you a message
it shouldn't scroll down
if you scroll up and you send me a message
it should scroll down

same logic we had before

WanderingHogan:
oh sweet, sounds good then as long as @jasonwoodland is okay with it

jasonwoodland (approved):

PR #4926, opened 2022-09-19
fix(announce): announce to all users

### LINKED ISSUE #4790: webrtc - group members that are not friends don't receive typing indicators
it seems like only friends send typing indicators to eachother

### LINKED ISSUE #4953: Friending requests problem
When adding/updating/removing a request or removing a friend we send an extra network request back to the sender

molimauro:
### What this PR does 📖
- Fix `Wire` logic for sending events (like the typing indicator) to all related-users instead to only friends
- Keep only users needed in the userState manager instead of keeping them all
- Fix extra network request to the sender in case of accept/reject/create request and remove friend (we keep users that leave a group for old messages, we remove them when the user itself leaves the group)
- Fix some problems related to the search in "Add friend" section
- Hide the user status in the "Add friend" section
- Add error message in case of duplicate request
- Remove ContextMenu in case of a request or a preview in the "Add friend" section and in the "Request" section

- Resolve #4790 
- Resolve #4953  

Please review that friending is working and there are no problems also in special case like: remove a friend while he is also on another group etc

Keep in mind that sometimes request may not go through not because of the frontend, you may understand that by sending a message and see if the other user is receiving it

### DIFF
--- a/components/mixins/UI/ContextMenu.ts
+++ b/components/mixins/UI/ContextMenu.ts
@@ -19,6 +19,7 @@ export default Vue.extend({
      */
     contextMenu(e: Event) {
       e.preventDefault()
+      if (!this.contextMenuValues.length) return
       const contextMenuStatus = this.ui.contextMenuStatus
       if (!contextMenuStatus) {
         this.$store.commit('ui/toggleContextMenu', true)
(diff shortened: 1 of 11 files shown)


molimauro:
By further thinking, we're storing users even when searching for them (like in Add Friend) or we keep them even if they leave a group. We should find a nice way not to spam users that are not related anymore with the account

iltumio (approved):

stavares843 (approved):

phillsatellite (approved):

PR #3062, opened 2022-05-05
feat(solana): error popup shows whenever an error occurs to prevent infinite loading

vimercati-samir:
What this PR does 📖
Error popup shows whenever an error occurs to prevent infinite loading.

Which issue(s) this PR fixes 🔨

AP-1494

Special notes for reviewers 🗒️

Additional comments 🎤
The error popup looks like this and it should appear when solana is down (instead of "Aligning satellites..." infinitely): https://cdn.discordapp.com/attachments/909842607159537725/970763603324043324/Screen_Shot_2022-05-02_at_2.07.40_PM.png

PR #4936, opened 2022-09-19
fix(signin-loading): removed auth middleware

### LINKED ISSUE #4933: Add timeout to 'linking satellites'
Add a timeout to linking satellites that kicks the user back to the sign in page after 30 seconds with an error message on the sign in page that says something like 'We were unable to log you in, please try again soon or create a new account.

Also investigate root cause of issue, Sheldon and Drew are debugging

ThomBos:
### What this PR does 📖
- fixes "forever linking" issue

- Resolve #4933


aewing (approved):

stavares843:
since the last deploy i haven't had that linking issue

stavares843 (changes requested):
Captura de ecrã 2022-09-21, às 21 37 47



when using this PR, it goes right away to this page, instead of the create account view

stavares843:
can we reuse the same logic we did [here](https://github.com/Satellite-im/Core-PWA/pull/3062) instead? when we were using solana, we added a modal that would appear after 30s, can we reuse that but for the nodes timeout?

ThomBos:
> when using this PR, it goes right away to this page, instead of the create account view
> 
> and then if you try to create an account, it returns an error because the user didn't have the previous view to setup a PIN
oh yeah, didn't noticed when refreshing, this happens the first time you open the app.
i'm working on it

stavares843:
is this ready to review again? @ThomBos 🔨

ThomBos:
> is this ready to review again? @ThomBos 🔨

yes, check if you see any issue 👍

components/views/friends/friend/Friend.vue
```diff
@@ -81,7 +81,12 @@ export default Vue.extend({
   methods: {
     async createFriendRequest() {
       this.loading = true
-      await iridium.friends?.requestCreate(this.user, false)
+      try {
+        await iridium.friends?.requestCreate(this.user, false)
+        this.$toast.show(this.$t('friends.request_sent') as string)
+      } catch (error: any) {
+        this.$toast.error(error)
+      }
```
molimauro, line 89: revert this, already fixed in dev

components/views/friends/search/Search.vue
```diff
@@ -120,7 +120,6 @@ export default Vue.extend({
     onFriendRequestSent() {
       this.user = null
       this.query = ''
-      this.$toast.show(this.$t('friends.request_sent') as string)
```
molimauro, line 123: revert this

molimauro (changes requested):

aewing (approved):

stavares843 (changes requested):
Captura de ecrã 2022-09-23, às 21 46 08


using fleek on this PR, returns error when doing a new account

aewing:
> Captura de ecrã 2022-09-23, às 21 46 08
> 
> using fleek on this PR, returns error when doing a new account

Looking at the logs it looks like the dev environment is misconfigured (probably CORS headers).
I don't think it's related to this PR. Can we test it locally while the dev & prod deployments are being worked on?

stavares843:
> Looking at the logs it looks like the dev environment is misconfigured (probably CORS headers).

fleek URL's were fine yesterday so

stavares843:
i can check locally though, but we shouldn't merge this before the deployment, also that this seems a overkill PR, we could have just reused the same logic we did [here](https://github.com/Satellite-im/Core-PWA/pull/3062) instead

stavares843:
Captura de ecrã 2022-09-23, às 23 08 04


locally is the same though

aewing:
> > Looking at the logs it looks like the dev environment is misconfigured (probably CORS headers).
> 
> fleek URL's were fine yesterday so

fleek and sync nodes weren't auto

ThomBos:
> using fleek on this PR, returns error when doing a new account

added catch for this, it should redirect to a network error modal now 👍

ThomBos:
> locally is the same though

It's not the same from the pictures you posted 😅 
Can't see any CORS errors in console in the last one... The only issue i see there, is the timeout error we throw.
It should be good with the catch just added

josephmcg:
Is there a way we could approach more simply that doesn't spread across so many different components?

Perhaps adding a check on the Global.vue component (which is rendered in desktop and mobile layouts. we can add it to default as well). It could determine if iridium is ready on mount, then kick the user back to the main login page if they aren't on another auth page

ThomBos:
> Is there a way we could approach more simply that doesn't spread copy pasted code across so many different components?
> 
> Perhaps adding a check on the Global.vue component (which is rendered in desktop and mobile layouts. we can add it to default as well). It could determine if iridium is ready on mount, then kick the user back to the main login page if they aren't on another auth page


Everytime we call "loadaccount" the catch is necessary, that's why that pasted code can't be removed. Maybe making changes on Global.vue can be removed stuff from layouts components 👍

molimauro:
Why do we want to ditch the middleware? Any particular reason?

ThomBos:
> Why do we want to ditch the middleware? Any particular reason?

@aewing was it to make it more simple, or you had something to integrate?

ThomBos:
> Is there a way we could approach more simply that doesn't spread copy pasted code across so many different components?
> 
> Perhaps adding a check on the Global.vue component (which is rendered in desktop and mobile layouts. we can add it to default as well). It could determine if iridium is ready on mount, then kick the user back to the main login page if they aren't on another auth page

Can tell you it didn't change when putting this code in Global.vue

josephmcg:
That's only half the process. Global isn't rendered on the default layout

aewing:
> Why do we want to ditch the middleware? Any particular reason?

We see some strange redirect loops and errors with the middleware approach. I think throwing a catch-all solution at this isn't necessarily the answer. We should be able to implement a view on the homepage (`/`) that checks for stored pin & attempts to restore, failing that it could redirect to the import/create account screen, which can redirect to profile entry if one isn't present. We shouldn't need to globally check state every time it changes if we think about when to redirect the user in the UI.

molimauro:
I think we should consider closing this one and:
- introducing a timeout for the signin starting from what we have in dev
- keep the middleware (its purpose is to run before every page so more reliable then a custom one) but refactor it since it is based on a old logic

Let me know your thoughts!

josephmcg:
yeah it seems like we could write some middleware that checks your current route + iridium status and knocks back appropriately. 

According to the doc it runs before render
https://nuxtjs.org/docs/directory-structure/middleware/

molimauro:
Opened the tickets for those, closing this one

PR #4943, opened 2022-09-20
fix(window): detect multiple instances of core in same browser

### LINKED ISSUE #4706: Detect if core is open on multiple tabs of the same browser
We need a way to detect if the user is trying to load the same site in multiple tabs. Not sure how to do it, or what the best feeling would be for users - do you want to block it from loading and have a warning, do you want to refresh the other tab they have open so it disconnects, etc? Experiment and find a good way to prevent the connection to the sync nodes from getting all wonky.

josephmcg:
### What this PR does 📖
- display window alerts in both instances if a second one is opened

- Resolve #4706

### DIFF
--- a/locales/en-US.js
+++ b/locales/en-US.js
@@ -719,6 +719,10 @@ export default {
       permission_denied:
         'Permission denied, please check your browser settings.',
     },
+    opened:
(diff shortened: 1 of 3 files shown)


nathan-power (approved):
LGTM!

PR #4981, opened 2022-09-22
fix(inputs): remove webkit autofill color

### LINKED ISSUE #4980: remove chrome password autofill white background
looks weird with the new eyeball to show/hide characters

josephmcg:
### What this PR does 📖
- title

previously looked like this 😵‍💫 
image

- Resolve #4980

### DIFF
--- a/components/interactables/Input/Input.less
+++ b/components/interactables/Input/Input.less
@@ -64,6 +64,14 @@
         &[type='search']::-webkit-search-cancel-button {
           display: none;
         }
+
+        &:-webkit-autofill,
+        &:-webkit-autofill:focus {
+          transition: background-color 600000s 0s, color 600000s 0s;
+        }
+        &[data-autocompleted] {
+          background-color: transparent !important;
+        }
       }
 
       .append {


components/interactables/Input/Input.less
```diff
@@ -64,6 +64,14 @@
         &[type='search']::-webkit-search-cancel-button {
           display: none;
         }
+
+        &:-webkit-autofill,
+        &:-webkit-autofill:focus {
+          transition: background-color 600000s 0s, color 600000s 0s;
```
molimauro, line 70: why 600000s?

josephmcg: To be completely honest, i copy pasted this lol

molimauro (approved):

phillsatellite (approved):

stavares843:
neat 🎉 


chrome is fixed

before on safari was yellow

image


vs this PR

image


better, still needs some fixing that we can do later, not important 


and i also noticed that in firefox, even though not from the PR but just adding that info in case you wanna fix it later, when you first generate a pw to auto fill, it shows the pw instead of being masked

https://user-images.githubusercontent.com/29093946/191808860-bdcf2a31-93ee-421b-9ca0-75c407de7a26.mov

image


how it should be


Captura de ecrã 2022-09-22, às 18 06 06

stavares843 (approved):

PR #5000, opened 2022-09-23
chore(fix): fixing recover account and minor changes

### LINKED ISSUE #4759: Recover Account needs to be restored
We need to allow users to restore their accounts with the seed. Right now this option is disabled. Improvements can come later

JustZacca:
### What this PR does 📖
Self-explanatory: This PR fixes the recover account function and does some other minor changes.

- Resolve #4759


libraries/Phantom/PhantomManager.ts
```diff
@@ -26,6 +26,7 @@ export default class PhantomManager {
    **/
   async initWallet() {
     if (this.provider.isPhantom) {
+      await this.$PhantomWalletAdapter.connect()
```
stavares843, line 29: do we need this duplicated line of code?

store/accounts/actions.ts
```diff
@@ -360,10 +364,10 @@ export default {
...
 
     const { pinHash } = state
-    const entropyMessage = IdentityManager.generateEntropyMessage(
-      $BlockchainClient.account.publicKey.toBase58(),
-      pinHash,
-    )
+    // const entropyMessage = IdentityManager.generateEntropyMessage(
+    //   $BlockchainClient.account.publicKey.toBase58(),
+    //   pinHash,
+    // )
```
stavares843, line 370: do we need this commented out code?

stavares843 (changes requested):
image


when you import a account asks you the username again


also, not as relevant but its cut off in the top when importing the account

image

store/accounts/actions.ts
JustZacca: Same as before

JustZacca:
Maybe is a configuration of your iridium node, because for me it works without re asking for the username.
It's a strange topic, maybe if @WanderingHogan can test it to verify

stavares843:
my iridium node? im using the fleek URL

hogan has better things to do than going to test something that i already did @JustZacca

libraries/Phantom/PhantomManager.ts
JustZacca: removed
I'll wait for the proper fix on Multi Wallet

JustZacca:
Okay, talking with @molimauro I've discovered an issue with the remote nodes of Iridium. We thought that it would be better to wait for the merge till it gets fixed.

aewing (approved):

josephmcg:
I could not successfully recover my account. it pushed me to new registration then was spinning forever

JustZacca:
> I could not successfully recover my account. it pushed me to new registration then was spinning forever

Okay, then is not an Iridium problem, the account recovery is inconsistent (sometimes it happens, sometimes it does not) I will investigate that

It could be still an Iridium Problem (not a version problem but a settings problem) so idk
I'll investigate

PR #5019, opened 2022-09-26
fix(friends): replace add friend button with message button if alread…

### LINKED ISSUE #5018: replace add friend button with message button if already friend

nathan-power:
…y friend

### What this PR does 📖
- When searching users in add friend search, if user is already added as a friend it currently allows you to add them again, which results in an error. This PR replaces the make friend request button with an open chat with user button if user is already friend.

- Resolve #5018

### DIFF
--- a/components/views/confirmation/Confirmation.vue
+++ b/components/views/confirmation/Confirmation.vue
@@ -73,7 +73,7 @@ export default Vue.extend({
   flex-direction: column;
   align-items: stretch;
   justify-content: center;
-  gap: 32px;
+  gap: 16px;
   height: 100%;
   max-width: 500px;
   padding: 24px;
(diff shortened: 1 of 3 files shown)


molimauro:
Dont you have:

image

?

molimauro:
This may happen when you have an outgoing/incoming request as well. I think we should handle that similarly, i'll create another ticket if you agree ;)

nathan-power:
> Or this is when you type partially the name?

Sorry yeah should have been more specific in the description. When there are multiple results in the search it's currently impossible to tell which accounts have already been added. This PR addresses that and changes the button to send a friend request into a button to open a chat.

Before:

nathan-power:
> This may happen when you have an outgoing/incoming request as well. I think we should handle that similarly, i'll create another ticket if you agree ;)

Yeah I agree, those are also things we need to address 👍

josephmcg (approved):

Issue #5031, opened 2022-09-27
MediaTrackContraints.volume is deprecated


josephmcg:
we need to replace this with something else eventually

components/views/settings/pages/audio/index.vue
https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/volume

PR #5032, opened 2022-09-27
feat(settings): toggle mic sensitivity test

### LINKED ISSUE #5021: Settings modal
I think we should access the microphone only when into the "Audio & Video" and not immediately.

josephmcg:
### What this PR does 📖
- will no longer automatically open audio track on settings page
- add button to create/stop audio track to test mic sensitivity
- improve type sense in audio settings index

- Resolve #5021 

### Additional comments 🎤
- Due to better type sense, I learned that volume is deprecated. I made a ticket to replace that https://github.com/Satellite-im/Core-PWA/issues/5031

### DIFF
--- a/components/mixins/UserPermissions/index.ts
+++ b/components/mixins/UserPermissions/index.ts
@@ -3,13 +3,13 @@ import DetectRTC from 'detectrtc'
 export type PermissionRequestOptions = MediaStreamConstraints | undefined
 
 interface permissionObject {
-  hasWebcam: Boolean
-  hasMicrophone: Boolean
-  hasSpeakers: Boolean
-  browser: String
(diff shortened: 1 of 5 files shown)


molimauro:
A minor ui problem:

image

Can we also disable the audio once we move away from that Audio&Video section?

josephmcg:
good idea, added a watcher for that.

I was having trouble getting the select component to render the ellipsis overflow properly. Since that has a `<ul>` outside the bounds of the parent element, I couldn't apply overflow hidden to trigger it. I opted to move the testing button up a line instead

molimauro (approved):
Nice!! Agree ;)

stavares843 (approved):

PR #5057, opened 2022-09-28
feat(chat): add message loader in chat

### LINKED ISSUE #4999: Add message loader in chat
add message loader in chat while message is sending. Just add the message in chat right after the send with grey color, and change the color to white once the send has been confirmed

molimauro:
### What this PR does 📖
- Add pending and failed status for messages in chat
- Message actions disabled while message are not confirmed
- Red color in case of text message failed, empty `src` for glyphs

- Resolve #4999 

- Pending/failed msgs are kept in an ephemeral state, wont be there on refresh

### Additional comments 🎤
In the future we should add the `retry` action for failed msgs

### DIFF
--- a/libraries/Iridium/chat/ChatManager.ts
+++ b/libraries/Iridium/chat/ChatManager.ts
@@ -18,8 +18,8 @@ import { v4 } from 'uuid'
 import {
   ChatError,
   Conversation,
-  ConversationMessage,
   ConversationMessagePayload,
+  ConversationMessage,
   IridiumConversationEvent,
   MessageAttachment,
   MessageEdit,
@@ -38,6 +38,7 @@ import {
   NotificationType,
 } from '~/libraries/Iridium/notifications/types'
 import { uploadFile } from '~/libraries/Iridium/utils'
+
 export type ConversationPubsubEvent = IridiumMessage<
   IridiumDecodedPayload<{
     message?: ConversationMessage
@@ -65,9 +66,11 @@ export default class ChatManager extends Emitter<ConversationMessage> {
   public ephemeral: {
     typing: { [key: Conversation['id']]: string[] | undefined }
     subscriptions: Conversation['id'][]
+    conversations: { [key: string]: ConversationMessage[] }
   } = {
     typing: {},
     subscriptions: [],
+    conversations: {},
   }
 
   async start() {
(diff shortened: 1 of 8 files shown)


stavares843:
how can we force the pending/failed messages? 💥 removed the internet and sent a message, but didn't show the failed/pending 🔨

libraries/Iridium/chat/ChatManager.ts
```diff
@@ -64,9 +65,11 @@ export default class ChatManager extends Emitter<ConversationMessage> {
   public ephemeral: {
     typing: { [key: Conversation['id']]: string[] | undefined }
     subscriptions: Conversation['id'][]
+    conversations: { [key: string]: ConversationMessage[] }
```
josephmcg, line 68: Maybe add | undefined since its key value pairs

josephmcg (commented):
I like it, just 1 thing i noticed

libraries/Iridium/chat/ChatManager.ts
molimauro: done!

molimauro:
> how can we force the pending/failed messages? 💥 removed the internet and sent a message, but didn't show the failed/pending 🔨

Sorry didnt see this. You can see the pending state when sending a message, it will appear with a lower opacity in the chat and then becomes white after the send is confirmed.
The failed one is not easy to trigger since it seems that we dont throw (iridium-side) in case of offline connection, must fix separately. But i can provide you the code to make it failing locally if needed

stavares843:
thanks for explaining 🔨

phillsatellite (approved):

aewing:
> > how can we force the pending/failed messages? 💥 removed the internet and sent a message, but didn't show the failed/pending 🔨
> 
> Sorry didnt see this. You can see the pending state when sending a message, it will appear with a lower opacity in the chat and then becomes white after the send is confirmed. The failed one is not easy to trigger since it seems that we dont throw (iridium-side) in case of offline connection, must fix separately. But i can provide you the code to make it failing locally if needed

Opened a PR to have the `.send` function return a boolean indicating perceived success.
https://github.com/Satellite-im/iridium/pull/112

I plan to improve this when I return from vacation to be based on a read receipt from the receiver.

molimauro:
Great thanks a lot @aewing!!

josephmcg (approved):

PR #5059, opened 2022-09-28
refactor(friends): consolidate mobile and desktop friend list items

### LINKED ISSUE #5058: consolidate best parts of desktop and mobile friends item

josephmcg:
### What this PR does 📖
- title
- simplify the template by using a Map inside of computed to fetch function, tooltip, and icon
- adjust buttons to be less reliant on colors for accessibility and ui reasons. Feel free to debate
- prevent friends and pending request users from showing up in the user search results
- breakout background to `#app` instead of body for better css modal transitions. all layouts use `#app` so its ok
- standardize 12px gap between user picture and name everywhere

- Resolve #5058 

- plan to add quickProfile on any user click, but that's still a bit buggy. it still relies on timeouts, so I want to clean that up before adding it everywhere


components/views/friends/search/Search.vue
```diff
@@ -67,58 +65,49 @@ export default Vue.extend({
...
-      }
       if (matches.length === 0) {
-        this.error = this.$t('friends.not_found') as string
+        this.error = this.$t('friends.not_found')
+      }
+
+      if (matches.length === 1) {
+        const did = matches[0].did
+        if (did === iridium.id) {
+          this.error = this.$t('friends.self_add')
```
molimauro, line 95: this is never reached i think

molimauro:
Dont know if its related to this but if you right click on a friend in the friendsList and then click to the "goto convo" icon, the context menu is still there after the redirect

josephmcg:
Thanks, I was experimenting with the quickprofile stuff and forgot to remove the stopPropagation after we decided to skip for now

components/views/friends/search/Search.vue
josephmcg: true, will revise

nathan-power (approved):
LGTM

PR #5064, opened 2022-09-28
feat(editable): removed old component, added status and about me to quick profile

### LINKED ISSUE #4989: InteractablesClickToEdit
I think we can ditch this component, just use a button and the Editable component we already have

ThomBos:
### What this PR does 📖
- removes old component

- Resolve #4989


josephmcg:
I dont think we want to delete the profile stuff. I thought the intended flow was quick profile would offer a path to the full profile.

The scope of the ticket only mentions the clickToEdit

molimauro:
Yes i already talked to Thomas. We're keeping the profile stuff

ThomBos:
> I dont think we want to delete the profile stuff. I thought the intended flow was quick profile would offer a path to the full profile.
> 
> The scope of the ticket only mentions the clickToEdit

yeah, that's why it was a draft PR, it should be good to go now

josephmcg:
Ah i was reviewing from the mobile app, the labels weren't updated in real time

josephmcg (commented):
a couple things I've updated

- avoid using TypographySubtitle and Title. These are old components which are being phased out. TypographyText should be used with props
- no need to wrap things in divs just for v-if. You can just attach it to the element itself, or use a template if you need to wrap multiple elements. div creates an extra element in the dom which may disrupt the flow
- avoid absolute unless you need it. the name could potentially flow into the button area. this should be handled with flex. let me know if you're unsure how to do that and I can provide an example
  - if the edit button does the same thing as clicking the user profile, then we don't need it
- we have a method on the user manager (getUser) that will fetch your own info from profile state or necessary info from user manager state. The quickProfile state is set using this logic, so you can just access user prop for everything rather than fetching status and about separately

I can't really provide much feedback on the profile changes because we don't have a flow to render it at the moment

josephmcg (approved):

PR #5084, opened 2022-09-29
fix(slimbar): vertical padding

josephmcg:
### What this PR does 📖
- switch from 8 on top to 16 so it's inline with the rest of the app
- also simplify props for sidebarListItem name, I've been using h4 for username everywhere and overriding color as needed

- Resolve #


nathan-power (approved):
LGTM!

PR #5091, opened 2022-09-29
chore(search): search improved

### LINKED ISSUE #5080: friends search - improve error message around existing friend or request
only show those errors if the input format is a regex match for did or short id. otherwise, just say no matches found

molimauro:
### What this PR does 📖
- Return "user not found" msg in case the only match is a friend (or is in the requests) and the query string is not an exact shortDID or DID
- Add utils for checking the shortDID and DID format

- Resolve #5080

### DIFF
--- a/components/views/friends/search/Search.vue
+++ b/components/views/friends/search/Search.vue
@@ -40,6 +40,7 @@ import { mapState } from 'vuex'
 import iridium from '~/libraries/Iridium/IridiumManager'
 import { User } from '~/libraries/Iridium/users/types'
 import { RootState } from '~/types/store/store'
+import { isDid, isShortDid } from '~/libraries/Iridium/utils'
 
 export default Vue.extend({
   data() {
@@ -64,7 +65,7 @@ export default Vue.extend({
         : `${iridium.id}`
     },
   },
-  async mounted() {
+  mounted() {
     if (this.$route.params.id) {
       this.query = this.$route.params.id
       this._searchFriend()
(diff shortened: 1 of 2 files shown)


phillsatellite (approved):

stavares843 (commented):
Captura de ecrã 2022-09-29, às 17 12 42


I just wonder if it would be more user-friendly if the user copy pasted a user that is already added as a friend maybe we could return the user and the icon to message directly 


Captura de ecrã 2022-09-29, às 17 13 33


just a suggestion though 😂

molimauro:
It was like that before, yesterday a PR went in that reverted the feature

stavares843:
and was that in purpose? 😂

josephmcg:
There are a couple factors behind that decision

- The bright buttons on the friend component were pretty jarring and attention grabbing.
- We shouldn't rely on color to convey meaning from a ux perspective.
  - https://uxmag.com/articles/usability-tip-dont-rely-on-color-to-convey-your-message
- due to the removal of color for the reasons above, it would be difficult to tell the buttons apart if they were all in one large list together
- if the tab is `add friend`, the user wants to add friends. There's not much value added from showing things other than potential users to add as a friend

if they do a search for 1 specific user (whether its did or short id), the error message should indicate that the user is already in the friends/requests list.

stavares843:
thanks for explaining 🔨

josephmcg:
maybe we could make the error message a link that leads to the request/friends tab depending on the case?

josephmcg (approved):
sometimes the short id doesn't return any matches if you're already friends, but we can investigate that later.

Would also like to move magic numbers to project config, and maybe add some regex instead. But this is better than it was before

molimauro:
> sometimes the short id doesn't return any matches if you're already friends, but we can investigate that later.
> 
> Would also like to move magic numbers to project config, and maybe add some regex instead. But this is better than it was before

Ideally those utils should come from Iridium, maybe we can ask Drew to include them should not be difficult

PR #5097, opened 2022-09-30
refactor(sidebar): simplify sidebar button styles

### LINKED ISSUE #5083: change sidebar buttons active display
switch files and friends buttons to a more low-key active state. the flair color is a little intense and draws too much focus

image

josephmcg:
### What this PR does 📖
- switch sidebar top buttons to be more inline with figma
- switch tabs indicator to use Tag component for the friend request count indicator so everything is standardized

- Resolve #5083 

- if everybody likes this, we could make a sidebarButton component to cut back copy/paste here

### DIFF
--- a/components/interactables/Tabs/Tabs.html
+++ b/components/interactables/Tabs/Tabs.html
@@ -5,9 +5,9 @@
     v-for="tab in tabs"
     @click="setRoute(tab.route)"
     :class="{active: route === tab.route}"
-    :data-badge="tab.badge || null"
   >
     {{tab.text}}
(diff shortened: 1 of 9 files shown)


molimauro:
I like this, but i noticed that on figma the are blue when selected.. Honestly dont know, should we ask Liz?

josephmcg:
my primary goal here was getting away from the blue. It's a pretty intense color that draws a lot of focus. I went with this style because it matches the hover/active state of the chat items in the sidebar

molimauro:
I prefer this too. Ok to me then ;) Let's see if anyone else has something to say!

jasonwoodland (approved):

PR #5101, opened 2022-09-30
refactor(users): implement better type safety and type guard

### LINKED ISSUE #4809: make better type guard utility
move the changes here out to a utility function that other computeds can use
https://github.com/Satellite-im/Core-PWA/pull/4808/files

maybe combine with the existing notNull guard

https://stackoverflow.com/questions/47632622/typescript-and-filter-boolean

josephmcg:
### What this PR does 📖
- add truthy type guard for array filtering
- add more accurate type definition for friends and users manager. key value pairs should always have an undefined option
  - this should be added to other places as well separately
- replace request `user` object with did. We should use the user manager as a single source of truth
- slight style adjustments to desktop friend request tab. leading with an `<hr>` looked kinda weird.
- rearrange the order of friend acceptance and deletion
  - was getting crashes with new data structure
  - friend request sender was seeing outgoing request and conversation in sidebar at the same time if the other user accepted while both are online

- Resolve #4809


josephmcg:
I'd like QA to review this thoroughly

stavares843:
we will do 🔨

stavares843 (commented):
phil and i were testing, so far the only issue within this PR that doesn't happen on dev is the following

image

image


if you remove a friend from the sidebar returns the above error

and then if you try to remove again returns this

image


never saw that error in dev

STR:
- add friend
- sidebar
- right click
- remove friend
- error

josephmcg:
thanks, that was a separate bug we introduced yesterday.
It has been fixed here and rebased into this branch
https://github.com/Satellite-im/Core-PWA/pull/5131

molimauro (approved):

PR #5128, opened 2022-10-03
refactor(user-profile): card style rework

### LINKED ISSUE #5127: Restore/rework user profile card's style

ThomBos:
### What this PR does 📖
- 

- Resolve #5127

### DIFF
--- a/components/interactables/Tabs/Tabs.less
+++ b/components/interactables/Tabs/Tabs.less
@@ -1,10 +1,9 @@
 .tabs {
   display: flex;
   gap: 32px;
-  margin-right: auto;
-  justify-content: flex-start;
-  flex-grow: 1;
   user-select: none;
+  padding: 0 24px;
+  border-bottom: 1px solid rgba(@gray, 0.5);
 
   .tab {
     display: flex;
@@ -22,6 +21,12 @@
 @media only screen and (max-width: @mobile-breakpoint) {
   .tabs {
     gap: @normal-spacing;
+    padding: 0;
+
+    .tab {
+      flex: 1;
+      justify-content: center;
+    }
   }
 }
 
--- a/components/ui/Global/Global.html
+++ b/components/ui/Global/Global.html
@@ -64,10 +64,10 @@
       />
     </UiModal>
     <UiModal
-      v-if="ui.modals.userProfile"
-      @close="toggleModal(ModalWindows.USER_PROFILE)"
+      v-if="ui.fullProfile"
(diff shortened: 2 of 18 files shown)


ThomBos:
Schermata 2022-10-03 alle 15 30 22

this is before

Schermata 2022-10-03 alle 15 21 25

Schermata 2022-10-03 alle 15 23 19

Schermata 2022-10-03 alle 15 24 06

Schermata 2022-10-03 alle 15 24 14

this is after

Posted images, because this can't be tested since these elements are still unused on dev (might be added soon again)

Although wiring with iridium is almost done, it might need some more style work (opened to suggestions) that's the reason i opened this pr

WanderingHogan:
@stavares843 can you check this out?

stavares843:
🔨

josephmcg:
I think allowing a user to view their own profile would be good so the user knows exactly what it looks like. The user can't get an exact idea of that based on the settings modal alone.

The only editable thing on the profile should be the note, imo

ThomBos:
> I think allowing a user to view their own profile would be good so the user knows exactly what it looks like. The user can't get an exact idea of that based on the settings modal alone.
> 
> The only editable thing on the profile should be the note, imo

Ah good point, let's do it!
The only thing that i don't fully understand, why do we keep the note? I mean, i understand using it on other users, but is it needed if the profile is mine? i think i can only read my own notes, no?

josephmcg (changes requested):
This isn't a full review, just a couple things i noticed while riding the train

josephmcg:
i'm going to make a couple adjustments

josephmcg:
- removed absolute positioning in favor of static
  - absolute is typically harder to manage across devices
- needed to apply overflow: hidden to a container for proper scroll behavior of the tab content. This overflow hidden would cause the negative margin on the image to get cut off. Went with a side by side display instead.
- removed image on mobile view. It looks fine on desktop, but it was difficult to fit it into mobile. Not sure if we even need this many pictures, realistically speaking. I feel like profile picture (inside the user circle) plus a profile banner would be plenty.

stavares843 (commented):
image


i see the same profile

other user

image



any thoughts?

josephmcg:
those are the quick profile and the settings profile, respectively. This profile is a separate modal. There's no render method implemented yet because it's not totally polished. You could add this line to QuickProfile if you want to see

```
        this.$store.commit('ui/setFullProfile', this.user)
```
something like this would be good

image

PR #5150, opened 2022-10-05
refactor(files): use UiModal with files fullscreen preview

### LINKED ISSUE #5149: switch FilesView fullscreen preview to use UiModal
take advantage of modal transition, better background styling, and esc event listener

josephmcg:
### What this PR does 📖
- wrap FilesView with UiModal to use better background styles, animation, modal focus trap, and esc event listener
- remove duplicate markup and download button for files with no preview
- unrelated - move incoming call to the first rendered modal. It would get hidden by settings/others. Maybe these should all be v-else-if instead

- Resolve #5149

### DIFF
--- a/components/ui/Global/Global.html
+++ b/components/ui/Global/Global.html
@@ -2,6 +2,9 @@
   <UiSvgDefs class="hidden" />
   <InteractablesContextMenu v-if="ui.contextMenuStatus" />
   <transition name="modal">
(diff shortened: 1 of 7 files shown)


WanderingHogan:
@josephmcg wouldn't we still want the file download button if we can't preview the file?

josephmcg:
Its still in the corner, just removing a duplicate button. This way every type of file has a consistent button location

molimauro (approved):

PR #5156, opened 2022-10-05
feat(notifications): add unread messages and pending request counts t…

### LINKED ISSUE #5140: add notification count to tabname
include

- incoming friend requests
- unread messages

something like this
image

you can add to this computed
https://github.com/Satellite-im/Core-PWA/blob/716b35cac1a6ac0b35ac2e2a4aa16aab9926f1fb/components/compositions/useMeta.ts#L1-L11

nathan-power:
…o head title

### What this PR does 📖
- Display the amount of unread messages and pending incoming friend requests in the browser tab

- Resolve #5140


libraries/Iridium/chat/ChatManager.ts
```diff
@@ -52,10 +47,12 @@ export type ConversationPubsubEvent = IridiumMessage<
 
 export type State = {
   conversations: { [key: string]: Conversation }
+  unreadCounts: { [key: string]: number }
```
josephmcg, line 50: Nathan and I discussed this in depth and are looking for other opinions

My thoughts

- This is extra state that can be derived from messages. I agree with Nathan that looping through every message of every conversation to check unread values would be computationally heavy.
- I propose we ditch an unread 'count' per conversation, just show an indicator if they have new messages at all. Logic below
- determine this boolean of 'new messages' by comparing `conversation.lastReadAt` to `messages.at(-1).at`. If there are messages later than your last viewed timestamp, show an indicator

components/views/navigation/sidebar/list/item/Item.vue
```diff
@@ -52,6 +52,12 @@ export default Vue.extend({
       const count = Object.values(
         iridium.chat.state.conversations[this.conversationId].message,
       ).filter((a) => a.at > this.conversation.lastReadAt).length
+
+      // Update the total unread count in iridium if it's different
+      if (iridium.chat.state.unreadCounts[this.conversationId] !== count) {
+        iridium.chat.updateUnreadCount(this.conversationId, count)
+      }
+
```
josephmcg, line 60: could we move this outside computed?

libraries/Iridium/chat/ChatManager.ts
molimauro: Just some random thoughts at the end of the day, tomorrow i'll think on that more carefully:

* if we have `messages.at(-1)` that means that msgs are ordered which also means that we dont have to loop through the full array but till the first read msg to have the unreads num
* when the slimbar will be ready we'll need the single convo unread number and show that for the user so probably we should think towards that
* cant we store that number in the `conversations` state and sum that up when we need the total?
* keep in mind that we will not load all messages and convos in the future, probably just a small portion of them. will load the rest of them based on user interactions (go to a specific convo, channel etc)

josephmcg: thanks for your input, Mauro. For now, we'll go for a computed based approach with Vue hooks. We plan on starting at the end of the message array and looping backwards until we reach a message that has been read to find the total.

Depending on how we implement message lazy loading/infinite scroll, we can adjust this later

josephmcg:
blocked until this goes through, will take advantage of new hooks
https://github.com/Satellite-im/Core-PWA/pull/5166

molimauro:
unlocked

josephmcg:
gonna close this and reopen a new branch

PR #5166, opened 2022-10-06
refactor(sidebar): switch SidebarListItem to composition api, add new hooks

### LINKED ISSUE #5165: refactor SidebarListItem to use composition api and hooks

josephmcg:
### What this PR does 📖
- update this component to use pre-existing hooks
- create new hooks such as `numUnreadMessages` and `sortedMessages`
- move dayjs timestamp utilities outside of vuex. we should purely rely on OS timezone rather than a user controlled setting. User can update OS timezone if they choose.

- Resolve #5165 

- left behind some options API because I haven't found a clean way to interact with vuex store inside script setup(). can revise later

### DIFF
--- a/components/compositions/conversations.ts
+++ b/components/compositions/conversations.ts
@@ -60,11 +60,16 @@ export function conversationHooks(conversationId?: Conversation['id']) {
   })
 
   const sortedMessages: ComputedRef<ConversationMessage[]> = computed(() => {
-    if (!conversationId) {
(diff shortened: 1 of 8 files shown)


molimauro:
you mean:

```
const { $store } = useNuxtApp()
```

?

molimauro:
Agreed with the timezone, we should remove that option from the settings and remove the package `vvo/tzdb` accordingly, writing a ticket for that

molimauro (approved):

PR #5178, opened 2022-10-07
chore(scroll): better loader ux and logic

### LINKED ISSUE #5157: Chat loader
Improve infinite scroll loader message

molimauro:
### What this PR does 📖
- Improve the chat loader
- Add tolerance to infinite scroll for loading new messaging before reaching the top
- Add random interval loading for simulating network conditions

- Resolve #5157 

### Additional comments 🎤
I think we can ditch the InfiniteScroll component and just use the onScroll method on Conversation.vue.
We need to make the renders (when adding new messages) *fast* or the scroll will glitch sometimes


components/views/chat/conversation/Conversation.vue
```diff
@@ -100,7 +100,7 @@ export default Vue.extend({
         const timeDiff = prevMessage ? message.at - prevMessage.at : 0
         const isNextDay = prevMessage
           ? !this.$dayjs(prevMessage.at).isSame(message.at, 'day')
-          : false
+          : true
```
molimauro, line 103: Show this when there are no previous messages (right under the loaders for example)

molimauro:
Need this in for the next ticket ;)

components/views/chat/conversation/Conversation.vue
```diff
@@ -253,21 +269,34 @@ export default Vue.extend({
...
+            // in case the user has scrolled beyond the loaders, lets reset the scroll position above the lastMessage
+            Math.min(
+              container.scrollTop +
+                container.clientHeight -
+                lastMessageOffsetTop,
+              0,
+            )
         })
-      }, 1000)
+      }, Math.random() * 1500)
```
WanderingHogan, line 299: i dont know why you are using Math.random() here, but i dont think it will cause anything negative

WanderingHogan (approved):

stavares843:
i tested but i wanted to verify hippo before adding feedback but hippo is currently unavailable

components/views/chat/conversation/Conversation.vue
molimauro: Just for simulating network conditions

stavares843 (commented):
Captura de ecrã 2022-10-12, às 19 14 16

only issue i had was this since date empty when scrolling and i couldn't reproduce in dev

besides that phil and i tested and was neat

PR #5179, opened 2022-10-07
feat(change-status): added change status modal

### LINKED ISSUE #5158: Popup for selecting the user state
when clicking here:

image

show a popup above it with the following options: online/away/busy/offline

ThomBos:
### What this PR does 📖
- Creates the Ui for the status change functionality

- Resolve #5158


components/ui/User/State/State.vue
```diff
@@ -3,6 +3,8 @@
     class="user-state"
     data-cy="user-state"
     :style="`width:${size}px; height:${size}px`"
+    :class="{ clickable: clickable }"
+    @click="clickable && $emit('openStatus')"
```
josephmcg, line 7: We need this to be keyboard accessible, could we go for a dynamic component instead? Button if clickable prop is true, otherwise div

components/views/user/statusChange/StatusChange.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <div class="modal">
```
josephmcg, line 2: Could we avoid repeating so much code? Probably better off with a for loop for every status option with various conditional render checks based off computed values. 

I handled a similar issue with a Map in the new FriendsItem component

components/views/user/statusChange/StatusChange.vue
```diff
@@ -0,0 +1,171 @@
...
+})
+</script>
+
+<style scoped lang="less">
+.modal {
+  &:extend(.modal-gradient);
+  display: flex;
+  flex-direction: column;
+  position: absolute;
+  bottom: calc(@sidebar-controls-height + @light-spacing);
```
josephmcg, line 116: Try adding position relative to whatever container this is in. Then values will be a lot easier and less bug prone

josephmcg (changes requested):
A couple things. Im wondering if we could break out some of those icons to be more reusable too

components/views/user/statusChange/StatusChange.vue
ThomBos: i think this is the best i can get, its container has to be just .5 rems below this div, even moving this div outside its container wouldn't help 😵‍💫

josephmcg: How about top with a negative value instead?

ThomBos: it works, but a thing like -17.5rem (not calculed) is better than that calculed value?

josephmcg: is the container holding this relative?

josephmcg: I like it because it's relying on the height of itself, rather than the height of something else. Keeps it simpler

ThomBos: ok gonna add a fixed height for it, in order to prevent any height change

ThomBos: I've been testing different solutions for this, but still think this is the best possible way and looks better overall.
The only thing, is that it depends on the bottom bar height for the positioning, but i think it's not a big deal since that's a fixed height and it can't stretch.

Gonna push soon, let me know if there's something to change 😄

components/views/user/statusChange/StatusChange.vue
ThomBos: It's similiar, but i had to go on another solution due to these icons, they are not components and can't be rendered in the same way. 
Might add them to the other repo so we can make a Map like that 👍

josephmcg:
I'll look at this again tomorrow

josephmcg:
Making some additional adjustments here. Ended up combining this into quickProfile and render based on a prop

josephmcg:
### New additions
- added this to quick profile options like i mentioned above. shows on click of your own user icon in the bottom left
- removed flavor text associated with statuses. This real estate is somewhat valuable, so we only want to show text that conveys clear meaning
- remove unused options from DynamicIcon component

### Additional things to do later
- onclick of quickprofile user circle, open full profile. maybe wait until we can flesh out the full profile a bit more
- move the short ID copy method into the quick profile and expand the quick profile open click area in the bottom of the sidebar to include your own username/status
- show icon next to current status indicator in quick profile. commented out for now because there were so many online indicators, it looked bad
- add focus trap to quick profile. hide status option list on keyboard blur

stavares843 (commented):
Captura de ecrã 2022-10-12, às 14 24 31

the status gets cut off if is too long

stavares843:
minor comments above, merging 🔨

PR #5190, opened 2022-10-10
chore(android): android build

WanderingHogan:
### What this PR does 📖
This build gets the basics back in to build the android version of Core, so we can start adding the capacitor modules we need to generate an APK we can submit to the app store.

N/A

This should not affect the web version at all. In the future, updates to include capacitor modules will need to test to verify they dont break the web version, as well as fix the native versions.


android/app/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
stavares843, line 1: we don't need ProGuard

android/app/src/main/res/drawable/ic_launcher_background.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0" encoding="utf-8"?>
```
stavares843, line 1: we don't need these assets

android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
```diff
@@ -0,0 +1,18 @@
+package com.getcapacitor.myapp;
```
stavares843, line 1: we can ditch this test

android/app/proguard-rules.pro
WanderingHogan: we probably will one day. https://developer.android.com/studio/build/shrink-code 
`Shrink, obfuscate, and optimize your app`

We will want to shrink and optimize, but not obfuscate. The source will be open anyways so it doesn't matter for that part.

android/app/src/main/res/drawable/ic_launcher_background.xml
WanderingHogan: This is the default capacitor android build just to get the work started, and there is a ticket to fix the icons, which can be updated to fix the loading background.

PR #5233, opened 2022-10-12
feat(native-android): clipboard improvement

drepram:
### What this PR does 📖
- Uses capacitor plugin for copying text on Accounts page. Haven't been able to test the APK, so reviews are most certainly awaited :D

- Tries to solve #5199

### DIFF
--- a/components/views/settings/pages/accounts/index.vue
+++ b/components/views/settings/pages/accounts/index.vue
@@ -53,44 +53,15 @@ export default Vue.extend({
     togglePhrase() {
       this.showPhrase = !this.showPhrase
     },
-    copyAddress(copyText: string) {
-      // navigator.clipboard.writeText(copyText)
-
-      // Uses @capacitor plugin now
-      const copyTextPromise = new Promise((resolve, reject) => {
-        resolve(
-          Clipboard.write({
-            string: copyText,
-          }),
-        )
-      })
-      copyTextPromise
-        .then(() => {
-          this.$toast.show(this.$t('ui.copied') as string)
+    async copyText(text: string) {
+      try {
+        await Clipboard.write({
+          string: text,
         })
-        .catch(() => {
-          this.$toast.error('Error when copying') // TODO: Put this in en-US later
-        })
-      this.$toast.show(this.$t('ui.copied') as string)
-    },
-    copyPhrase() {
-      // Uses @capacitor plugin now
-      const copyTextPromise = new Promise((resolve, reject) => {
-        resolve(
-          Clipboard.write({
-            string: this.accountPhrase,
-          }),
-        )
-      })
-      copyTextPromise
-        .then(() => {
-          this.$toast.show(this.$t('ui.copied') as string)
-        })
-        .catch(() => {
-          this.$toast.error('Error when copying') // TODO: Put this in en-US later
-        })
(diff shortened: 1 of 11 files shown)


components/views/settings/pages/accounts/index.vue
```diff
@@ -53,11 +54,42 @@ export default Vue.extend({
...
+        )
+      })
+      copyTextPromise
+        .then(() => {
+          this.$toast.show(this.$t('ui.copied') as string)
+        })
+        .catch(() => {
+          this.$toast.error('Error when copying') // TODO: Put this in en-US later
+        })
+      this.$toast.show(this.$t('ui.copied') as string)
```
stavares843, line 92: is

`      this.$toast.show(this.$t('ui.copied') as string)`



supposed to appear 2x?

components/views/settings/pages/accounts/index.vue
```diff
@@ -53,11 +54,42 @@ export default Vue.extend({
       this.showPhrase = !this.showPhrase
     },
     copyAddress(copyText: string) {
-      navigator.clipboard.writeText(copyText)
+      // navigator.clipboard.writeText(copyText)
+
+      // Uses @capacitor plugin now
+      const copyTextPromise = new Promise((resolve, reject) => {
```
WanderingHogan, line 60: you can instead make the whole function async, like this:
```
    async copyAddress(copyText: string) {
      await Clipboard.write({
        string: copyText,
      })
        .then(() => {
          this.$toast.show(this.$t('ui.copied') as string)
        })
        .catch(() => {
          this.$toast.error('Error when copying') // TODO: Put this in en-US later
        })
    },
    ```

File: components/views/settings/pages/accounts/index.vue [hunk 3]
```diff
@@ -53,11 +54,42 @@ export default Vue.extend({
...
+        })
+        .catch(() => {
+          this.$toast.error('Error when copying') // TODO: Put this in en-US later
+        })
       this.$toast.show(this.$t('ui.copied') as string)
     },
     copyPhrase() {
-      navigator.clipboard.writeText(this.accountPhrase)
+      // Uses @capacitor plugin now
+      const copyTextPromise = new Promise((resolve, reject) => {
```
WanderingHogan, line 78: you can get rid of this whole copyPhrase method if you update the template. Instead of where you see @click="copyPhrase" you could do @click="copyAddress(accountPhrase)"

I would suggest renaming copyAddress to just something like copyMethod, or something generic

WanderingHogan:
Hey @drepram, great first pass. In addition to Sara T's comments above and my comments inline, can you look at making this changes globally? Anywhere you might see navigator.clipboard?

WanderingHogan (changes requested):

components/views/settings/pages/accounts/index.vue
josephmcg: imo, one of the benefits of using async await would be avoiding `then`. I'll push something up

josephmcg:
consulted with Andre, he's going to break this out into a reusable composition and update other components accordingly

josephmcg:
Looks good, I fixed a missing import and made some changes to calm the linter down.

everything works well on desktop, if someone else could confirm the apk works then we're good to go

PR #5255, opened 2022-10-13
fix(status-change-ui): fixed status overflow and busy icon size

ThomBos:
### What this PR does 📖
- 

- Resolve #

### DIFF
--- a/components/ui/User/State/State.vue
+++ b/components/ui/User/State/State.vue
@@ -75,7 +75,9 @@ export default Vue.extend({
       if (this.isTyping) {
         return 'typing'
       }
-      return 'busy'
+      return this.user.did === iridium.id
+        ? 'online'
+        : this.users.ephemeral.status?.[this.user.did] ?? 'offline'
     },
     isTyping(): boolean {
       if (!this.conversationId.length) {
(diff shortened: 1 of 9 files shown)


stavares843 (approved):

molimauro:
If we change that here we should change also on the UiUserState accordingly

molimauro (dismissed):
fix also on the main user state component

josephmcg:
those css classes didnt exist. plus we don't need to define these classes because there is a utility class.

name was properly overflowing with the current styles, just copy that. Please ask questions if anything is unclear, don't just switch labels

components/ui/User/State/State.vue
```diff
@@ -75,9 +75,7 @@ export default Vue.extend({
       if (this.isTyping) {
         return 'typing'
       }
-      return this.user.did === iridium.id
-        ? 'online'
-        : this.users.ephemeral.status?.[this.user.did] ?? 'offline'
+      return 'busy'
```
molimauro: revert i guess

ThomBos: yep, done! that was just for testing it

PR #5309, opened 2022-10-18
refactor(enhancers): cleanup enhancers js logic, resolve css bug

### LINKED ISSUE #5308: enhancers z index issue
image

josephmcg:
### What this PR does 📖
- title

- Resolve #5308 

- enhancers still needs a rework, some of this is kinda weird

### Additional comments 🎤
-

### DIFF
--- a/components/interactables/ButtonGroup/ButtonGroup.vue
+++ b/components/interactables/ButtonGroup/ButtonGroup.vue
@@ -1,83 +1,9 @@
-<template src="./ButtonGroup.html"></template>
+<template>
+  <div class="button-group">
+    <slot />
+  </div>
+</template>
 
-<script lang="ts">
-import Vue from 'vue'
-export default Vue.extend({
-  props: {
-    value: {
-      type: [String, Boolean, Number],
-      required: false,
-      default: null,
-    },
-    values: {
-      type: Array,
-      required: false,
-      default: null,
-    },
-    /**
-     * Callable function when button is clicked
-     *
-     * @remarks
-     * This should usually be a store mutation or UI change.
-     */
-    action: {
-      type: Function,
-      required: false,
-      default: null,
-    },
-    /**
-     * Set the active class of selected button
-     *
-     * @remarks
-     * This class has already defined
-     */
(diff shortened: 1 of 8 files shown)


molimauro:
Yes also the enhancers on the edit chatbar are not working.
It's weird to have the <Enhancers> in the Chatbar.vue and trigger it with vuex also from outside (like from the Editable in the Edit.html). Furthermore, by using `setChatbarContent` we broke the emoji insertion in the middle of a string, we always insert the new emojy at the end even if the cursor is not at the end. We should use `handleChatTextFromOutside` instead

josephmcg:
Agreed, i can take a look at it tomorrow

josephmcg:
got sidetracked working on other bugs. maybe merging this first would be good so the PR isnt too large in scale

components/interactables/ButtonGroup/ButtonGroup.vue
```diff
@@ -1,83 +1,9 @@
...
-          if (this.action != null) {
-            this.action(text)
-          }
-        })
-      })
-    },
-  },
-})
-</script>
+<script setup lang="ts"></script>
```
molimauro, line 7: Why is this empty?

josephmcg: i assumed it was needed, but I'm not sure. can try removing

josephmcg: they were very heavy-handed. I opted to placed click listeners on the buttons to achieve the same functionality

molimauro: Ahaha yeah i noticed. So we're good. Can we remove the props from the component in `components/views/chat/enhancers/Enhancers.html` then right?

josephmcg: ah yes, thanks

molimauro (approved):

phillsatellite (approved):

stavares843 (approved):

PR #5323, opened 2022-10-19
fix(context-menu): fix action sheet styles

### LINKED ISSUE #5322: Native Mobile - Fix action sheet styles

nathan-power:
### What this PR does 📖
- Fixes the styling for action sheets in mobile

- Resolve #5322 

- To bring up an action sheet long press a list item in the chat list on mobile.
- Styling bug for action sheets opened inside the friends screen will be fixed in a separate PR.

### DIFF
--- a/components/ui/ActionSheet/ActionSheet.vue
+++ b/components/ui/ActionSheet/ActionSheet.vue
@@ -1,7 +1,6 @@
 // ActionSheet
 <template>
   <div
-    ref="contextMenu"
     class="context-menu-container"
     :class="{ 'is-visible': isVisible }"
     data-cy="context-menu"
@@ -56,7 +55,7 @@
 </template>
 
 <script setup lang="ts">
-import { watch, computed, ComputedRef, ref } from 'vue'
+import { computed, ComputedRef } from 'vue'
 import { ContextMenuItem, EmojiUsage } from '~/store/ui/types'
 
 interface Props {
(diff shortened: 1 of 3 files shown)


components/ui/ActionSheet/ActionSheet.vue
```diff
@@ -51,56 +55,53 @@
...
+    return
+  }
+  emit('action', func)
+}
+
+const isVisible = computed(() => props.isVisible)
+
+watch(isVisible, (isVisible) => {
+  $store.commit('ui/setIsMobileNavVisible', !isVisible)
+})
```
josephmcg, line 100: we saw copilot doing this yesterday, but maybe a function watcher is better if we use props
https://stackoverflow.com/questions/59125857/how-to-watch-props-change-with-vue-composition-api-vue-3

josephmcg (changes requested):
nice, just 1 minor change. we can take a look at the mixin later and remove it

components/views/navigation/mobile/nav/Nav.less
```diff
@@ -1,27 +1,13 @@
 .nav {
+  height: @mobile-nav-height;
   width: 100%;
```
josephmcg, line 3: maybe we dont need this height value since jason adjusted the transitions in a separate PR

nathan-power: You're right it's not required anymore
I'll remove it now

josephmcg (changes requested):
upon closing context menu on mobile, it feels like there's a second where you can't click anything

josephmcg (changes requested):
this currently renders bottom nav on slide 2 of a chat page once you close the action sheet.

Do we need to hide/show the bottom nav on action sheet render?

josephmcg (approved):

PR #5357, opened 2022-10-20
chore(cypress): One CI file for all cypress tasks and fixes for videocalls

luisecm:
### What this PR does 📖
- Just a small draft test for creating a CI with cypress running on chrome and firefox. Ideally, if it works, we could add the ability to execute the tests in chrome, firefox and safari at the same time.
- Using only one CI file for all cypress tasks
- Updates to videocall tests to avoid CI issues
- Removed firefox for now since it was causing issues but we can add it later once that we figured it out why its failing only with firefox

- Resolve #None

- Sending as draft since I am pretty sure it will need several commits

### Additional comments 🎤
-


stavares843 (approved):

stavares843:
🎉

PR #5366, opened 2022-10-21
feat(tag): highlighted user tag

### LINKED ISSUE #5188: autocomplete - clickable profile
probably when we send a @ in the chat, we need to make it bold, clickable and to show the quick profile

https://github.com/Satellite-im/Core-PWA/pull/5163#pullrequestreview-1134714858

thoughts? @molimauro

ThomBos:
### What this PR does 📖
- adds user tags on chat

- Resolve #5188 

- 

### Additional comments 🎤
-

### DIFF
--- a/libraries/ui/Markdown.ts
+++ b/libraries/ui/Markdown.ts
@@ -1,9 +1,11 @@
 import * as markdown from 'simple-markdown'
+import { User } from '~/libraries/Iridium/users/types'
 
 interface Options {
   liveTyping?: boolean
   escapeHTML?: boolean
   cssModuleNames?: Record<string, string>
+  conversationParticipants?: User[] | null
 }
 
 type MarkdownRules = Record<
@@ -281,7 +283,9 @@ const rules: MarkdownRules = {
     match: (source, state) => {
       if (
         (state.prevCapture && !/^$|\s+$/.test(state.prevCapture[0])) ||
-        state.liveTyping
+        !state.conversationParticipants?.find(
+          (el: User) => el.name === /^@([^\s]+)/.exec(source)?.[1],
+        )
       ) {
         return null
       }
@@ -298,16 +302,27 @@ const rules: MarkdownRules = {
       }
     },
     html: (node, output, state) => {
-      return htmlTag(
-        'span',
-        output(node.content, state),
-        {
-          class: 'user-tag',
-          tabindex: '0',
-          role: 'button',
-        },
-        state,
-      )
+      return state.liveTyping
+        ? htmlTag(
+            'span',
+            output(node.content, state),
+            {
+              class: 'user-tag dark',
+              tabindex: '0',
+              role: 'button',
+            },
+            state,
+          )
+        : htmlTag(
+            'span',
+            output(node.content, state),
+            {
+              class: 'user-tag light',
+              tabindex: '0',
+              role: 'button',
+            },
+            state,
+          )
     },
   },
 }
@@ -325,6 +340,7 @@ export function toHTML(source: string, options: Options = {}): string {
     liveTyping: options.liveTyping ?? true,
     escapeHTML: options.escapeHTML ?? true,
     cssModuleNames: options.cssModuleNames || null,
+    conversationParticipants: options.conversationParticipants ?? null,
   }
 
   return _htmlOutput(_parser(source, state), state)
--- a/libraries/ui/Markdown.ts
+++ b/libraries/ui/Markdown.ts
@@ -285,7 +285,6 @@ const rules: MarkdownRules = {
       ) {
         return null
       }
-      console.log('#####', state)
       return /^@([^\s]+)/.exec(source)
     },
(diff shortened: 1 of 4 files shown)


libraries/ui/Markdown.ts
```diff
@@ -276,6 +276,37 @@ const rules: MarkdownRules = {
...
+        ],
+      }
+    },
+    html: (node, output, state) => {
+      return htmlTag(
+        'span',
+        output(node.content, state),
+        { class: 'user-tag' },
+        state,
+      )
```
josephmcg, line 307: could you add some accessibility attributes to make it keyboard accessible?
tabindex=0, role="button"

ThomBos:
tried many attempts, but i'm still struggling to find a way to add a function to the returned component, lmk if you know any way to accomplish such a thing 😬

molimauro:
I was thinking the same about the payload, when i wrote the md file i had in mind a way to pass callbacks, i just need to recover that idea. I'll try to have a look at this when i have time ;)

Agree with the use of typescript generics, there are several types like the messages one that can benefit a lot! Thanks @josephmcg for the input, we can create a ticket and focus on that when we have time!

ThomBos:
For my side i think this is ready and can be tested, let me know if there's anything else to change 💥

josephmcg:
Users will eventually be able to change their name, so we cant rely on the name in the message to identify someone. Plus multiple people in a group could have the same name.

Can you check the solution I suggested in a previous comment?

josephmcg:
I'm working on this now, it's a bit tricky

josephmcg:
I got sucked into a different rabbit hole today, didn't have time to look into this

josephmcg:
I couldn't find a simple way to integrate this with our current setup.

https://github.com/ariabuckles/simple-markdown
Our markdown library is deprecated. They claim development has moved to a new repo but there haven't been any updates to src in 6 months. We should consider something else

Slate looks super nice, but only for react.
https://github.com/ianstormtaylor/slate

markdown-it seems robust enough
https://github.com/markdown-it/markdown-it

molimauro:
> I couldn't find a simple way to integrate this with our current setup.
> 
> [ariabuckles/simple-markdown](https://github.com/ariabuckles/simple-markdown) Our markdown library is deprecated. They claim development has moved to a new repo but there haven't been any updates to src in 6 months. We should consider something else
> 
> Slate looks super nice, but only for react. [ianstormtaylor/slate](https://github.com/ianstormtaylor/slate)
> 
> markdown-it seems robust enough [markdown-it/markdown-it](https://github.com/markdown-it/markdown-it)

Slate would have been my primary choice for sure, i was considering porting it to Vue when have time (there is already a lib that ports Slate to Vue but does not seem updated). Is the lib Discord is using (kind of, a fork of Slate).

Markdown-it was our old one, Matt wanted to replace that (cant remember why), but we can discuss and see if that makes sense to integrate back.

I would maybe switch to the new simple-markdown lib, even though im not sure how to solve this problem. I think here the big problem is that we cannot render and id under the hood right?

josephmcg:
I think our source and result being the same string is somewhat problematic. We can't programmatically replace content because that content will no longer continue to match the initial condition.

I suggested a new library because the current one seems dead, or at least heading in that direction. Markdown-it is more popular and also supports plugins, so there may be an out-of-the-box solution to mentions.

The discord guide was pretty enlightening
https://v12.discordjs.guide/miscellaneous/parsing-mention-arguments.html#how-discord-mentions-work

molimauro:
Yes exactly, our only problem is only while we type, once the message is sent we can replace easily the id with a name. We can try keeping track of when mentions are inserted/removed in the chatbar like in an array. Or another temp solution can be replace the name in the chatbar with `@mauro#Lwkgh1` so that we have a ref of the did (or even visually hide the # part but this is not a lot elegant)

ThomBos:
> Yes exactly, our only problem is only while we type, once the message is sent we can replace easily the id with a name. We can try keeping track of when mentions are inserted/removed in the chatbar like in an array. Or another temp solution can be replace the name in the chatbar with `@mauro#Lwkgh1` so that we have a ref of the did (or even visually hide the # part but this is not a lot elegant)

A thing that might work, could be keeping that # in the tag, in order to filter users using that.... I don't see any other ways even adding a "tags" field on messages, because it doesn't have any did or id to detect which user is. (so in case one user changed name we'd have the same issue)

molimauro:
We can also add the # part to a data attribute

https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

ThomBos:
> We can also add the # part to a data attribute
> 
> https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

it looks interesting, need to check if the filter we have detects the html element after the refresh

ThomBos:
it now checks for did or name after an @, but it cannot check for the key we are now passing to the html

PR #5368, opened 2022-10-24
refactor(tabs): comp api upgrade

### LINKED ISSUE #5367: tab component comp api upgrade

josephmcg:
### What this PR does 📖
- title
- reorder interface declarations in Button. That was an early refactor, I now have a more defined template for script setup

- Resolve #5367 

- 

### Additional comments 🎤
-


components/interactables/Tabs/Tabs.vue
```diff
@@ -1,24 +1,39 @@
...
+interface Props {
+  route: string
+  tabs: Tab[]
+}
+interface Emits {
+  (e: 'setRoute', route: string): void
+}
+
+const props = withDefaults(defineProps<Props>(), {
+  route: '',
```
nathan-power, line 31: As we're referring to the prop variables directly without using this props variable we can just remove the instantiation of "const props" to avoid an unused variable warning and just have:
`withDefaults(defineProps<Props>(), {`

nathan-power (changes requested):
Looks good! Just one small change and it's good to merge 👍

components/interactables/Tabs/Tabs.vue
josephmcg: truuu

PR #5379, opened 2022-10-24
refactor(user-picker): upgrade to composition api

### LINKED ISSUE #5370: userpicker comp api upgrade
components/interactables/UserPicker/UserPicker.vue

nathan-power:
### What this PR does 📖
- Upgrades UserPicker component to use composition API

- Resolve #5370


josephmcg (changes requested):
needs to exclude users based on exclude prop. you should be able to filter the friends list you get from the hook

components/interactables/UserPicker/UserPicker.vue
```diff
@@ -1,64 +1,89 @@
...
+const selected: Ref<User[]> = ref([])
+const filter: Ref<string> = ref('')
+
+const friendsWithoutExcluded: ComputedRef<User[]> = computed(() => {
+  return friends.value.filter((friend) => !props.exclude.includes(friend.did))
+})
+
+const filteredFriends: ComputedRef<User[]> = computed(() => {
+  if (!filter.value) return friends.value
+  return fuzzysort
```
josephmcg, line 68: double check guard clause. need to return friendsWithoutExcluded

components/interactables/UserPicker/UserPicker.vue
```diff
@@ -1,64 +1,89 @@
...
+const toggle = (friend: User) => {
+  if (isSelected(friend)) {
+    selected.value = selected.value.filter((f) => f.did !== friend.did)
+  } else {
+    selected.value = [...selected.value, friend]
+  }
+  emit('input', selected.value)
+}
+
+watch(selected, (value: User[]) => emit('input', value), { deep: true })
```
molimauro, line 86: is the `deep` necessary?

components/interactables/UserPicker/UserPicker.vue
nathan-power: Good catch, will fix that now

components/interactables/UserPicker/UserPicker.vue
nathan-power: True, for this use case it's pretty redundant
Thanks for the catch 👍

josephmcg (approved):

PR #5383, opened 2022-10-24
feat(releaseData): add new hook for release data

josephmcg:
### What this PR does 📖
- create new compositions folder outside of /components. All compositions should be moved here eventually. we get lots of console warning about the current structure missing default exports
- settings modal is fetching data every single mount. This will fetch one time on first render, then never again.
- not yet implemented, waiting for other refactor to go through first

- Resolve #

- 

### Additional comments 🎤
-

### DIFF
--- a/compositions/release.ts
+++ b/compositions/release.ts
@@ -1,0 +1,13 @@
+import { ref, Ref } from 'vue'
+
+export async function fetchReleaseData(): Promise<any> {
+  const data: Ref<any> = ref(null)
+
+  if (!data.value) {
+    const res = await fetch(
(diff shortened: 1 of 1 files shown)


molimauro:
There is also the built-in `https://nuxtjs.org/docs/features/data-fetching`

josephmcg:
oh cool, are there any benefits to using nuxt to fetch data?

molimauro:
oh it provides you all the fetching utilities, options, as well as the various pending/error states, the automatic caching on page revisits, browser refresh etc. Maybe we can even use this directly: https://v3.nuxtjs.org/getting-started/data-fetching#usefetch

josephmcg:
im ok merging for now, down to take another look later

PR #5417, opened 2022-10-27
feat(call): resolve duration timestamp issue

### LINKED ISSUE #5353: Calls - broken timer
STR:
- go to hippo
- call someone
- answer call
- timer call is always zero

Captura de ecrã 2022-10-20, às 12 45 01

Chrome, macOS

drepram:
### What this PR does 📖
- Resolve duration timestamp issue
- Add a description on how to resolve Yarn cache error

- Resolve #5353 

- Please let me know if my way of patching the error needs to be polished (in the case that it needs to be rewritten to fit in with Satellite's code standard)

### Additional comments 🎤
Proof that it works:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Run the linkscript - this will build iridium and let you install it:
 - Install dependencies `yarn`
 - Run hot reload/dev version `yarn dev`
 
+Or for short: $ chmod +x linkscript.sh && ./linkscript.sh && yarn && yarn dev $. If any error comes up during the process, be sure to remove your `node_modules`, clear your yarn cache, and run it again.
+
 ## Running Cypress tests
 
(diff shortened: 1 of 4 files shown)


josephmcg (approved):
Thanks, Andre. this was a great first pass that achieved the desired functionality.

I made a couple small improvements:

- anytime you set an interval, it should also be cleared when it is no longer needed to avoid memory leaks
- the previous implementation would only set the timer after the 1 second interval triggers. After component creation, the timestamp would be blank for 1 second. I added a method call inside created to set the value immediately.
- I changed the method and variable names to be more descriptive, and removed some extra variables (while still maintaining readability)

since we're copy pasting logic, we could break this out to a single composable after upgrading these components to use script setup.

let me know if you have any questions

PR #5434, opened 2022-10-28
refactor(callTimer): add duration function to webrtc hooks

drepram:
### What this PR does 📖
- Extract duration timestamp functionality into composable on webrtcHooks
- Implement duration timestamp function composable on Call
- Implement duration timestamp function composable on Heading

### Additional comments 🎤

### DIFF
--- a/components/compositions/webrtc.ts
+++ b/components/compositions/webrtc.ts
@@ -5,6 +5,7 @@ import iridium from '~/libraries/Iridium/IridiumManager'
 import { User } from '~/libraries/Iridium/users/types'
 import { TrackKind } from '~/libraries/WebRTC/types'
 import { conversationHooks } from '~/components/compositions/conversations'
+import { formatDuration } from '~/utilities/duration'
 
 export function webrtcHooks(conversationId?: Conversation['id']) {
   const managers = reactive({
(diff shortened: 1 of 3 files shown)


components/compositions/webrtc.ts
```diff
@@ -63,5 +64,11 @@ export function webrtcHooks(conversationId?: Conversation['id']) {
       $nuxt.$toast.error($nuxt.$i18n.t(e.message))
     }
   }
-  return { enableRTC, isActiveCall, call }
+
+  function updateDuration(callStartedAt: number) {
+    const durationOfCall = Date.now() - callStartedAt
+    return formatDuration(durationOfCall / 1000)
+  }
+
```
josephmcg, line 72: rather than passing this callStartedAt value in, you can access it directly inside of the composable.

components/views/chat/message/call/Call.vue
```diff
@@ -65,9 +72,9 @@ export default Vue.extend({
     },
   },
   created() {
-    this.updateDuration()
+    this.updateLocalDuration()
     this.timestampInterval = setInterval(() => {
-      this.updateDuration()
+      this.updateLocalDuration()
     }, 1000)
   },
```
josephmcg, line 79: Rather than keeping this logic copy/pasted and tracking two different values with two intervals, I imagined all of it being centralized

josephmcg (changes requested):

components/compositions/webrtc.ts
drepram: good point! thank you

josephmcg:
solution here
https://github.com/Satellite-im/Core-PWA/pull/5452

PR #5444, opened 2022-11-02
refactor(glyphs): move information out of mocks, improve type sense, …

### LINKED ISSUE #5443: refactor glyphs to be simpler. remove from vuex state

josephmcg:
…comp api upgrade

### What this PR does 📖
- title
- fix bug where ephemeral message wasn't getting removed after send
- only partially done. wanted to break it up across PRs
- temporarily remove 'recently used' glyphs. This was being stored in vuex store and persisted in local storage. that approach will not support multiple devices or platforms.

- Resolve #5443 

- 

### Additional comments 🎤
-


components/views/chat/enhancers/glyphs/Glyphs.vue
```diff
@@ -1,31 +1,40 @@
-<template src="./Glyphs.html"></template>
-<script lang="ts">
-import Vue from 'vue'
-import VirtualList from 'vue-virtual-scroll-list'
```
molimauro, line 4: Can we remove this lib if not needed anymore?

molimauro (approved):
Approved but we have problems with the preview, for some reasons we cannot create accounts, not related to this though.

phillsatellite (approved):

luisecm:
PR should be good to go on the cypress side, since the cypress GH actions failing now are related to iridium things that we are fixing now

components/views/chat/enhancers/glyphs/Glyphs.vue
josephmcg: yes! thank you

